PR #15, opened 2014-11-24
Prepare refactoring: Add two value objects OSTarget and OSArchTarget (requires PR #14)

cr7pt0gr4ph7:
Based on ros-infrastructure/ros_buildfarm#14.

### DIFF
--- a/ros_buildfarm/common.py
+++ b/ros_buildfarm/common.py
@@ -1,3 +1,17 @@
+from collections import namedtuple
+
+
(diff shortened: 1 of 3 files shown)


dirk-thomas:
We haven't applied the remaining changes from this PR - mostly passing things like `os_name` and `os_code_name` as a single argument. I don't think the slight shorter signature is enough benefit over the need to wrap / unwrap the data. Therefore I will close this - I hope you understand.

PR #33, opened 2015-01-06
Turn this into a setuptools Python package

wjwwood:
This pull request does a few things in order to set this up as a `setuptools` package:

- Moves the `templates` folder into the source tree at `ros_buildfarm/templates`
- Moves the implementation of `templates.py` into `templates/__init__.py` to avoid confusion
- Updates the global module variable which points to the root of the templates directory
- Adds `setup.py` and `MANIFEST.in` to allow installation using `setuptools`

I'm not sure if we should leave the implementation of `templates.py` in `templates/__init__.py` or revert to the old layout, or move the templates to a different folder name. I can change that if desired.


esteve:
+1

wjwwood:
Ok, I added an extension in 052be71 which stores the VCS information in a generated Python module at install time, if it is available. This should cover:

- Running from source (just adding to `PYTHONPATH`)
- Running after installing from source (extracts git url at install time)
- Running after installing from tarball (uses fallback url of ros-infrastructure/ros_buildfarm)
 - This will be the default when installing from pip, which makes sense because the pip one will be release from here

I have some concerns about the fragility of this though. For example, these scenarios will likely not work as expected and might be pretty hard to debug for the user:

- Running after installing from a local git repo which was cloned from a private or non-github repo
- Running after installing from a local git repo which was cloned from another local folder or from another machine over ssh
 - This may fail as the credentials will not exist in the docker container
- Running after installing from a git repo where the desired remote is named something other than origin
 - origin may also exist, but it will not pull the one they are expecting

Basically this strategy assumes that the url for "origin" is the correct url and that it is accessible from within the docker container.

I'm not sure what else, if anything, we should do, but I just wanted to voice my concerns about the approach.

Next I have to tackle the locating of scripts at run time and handling both the structure in the source repo and their flattened structure after being installed.

wjwwood:
Ok, I rebased this and addressed the issue of locating the scripts at runtime when installed or not installed.

To accomplish this I copied the `scripts` folder into the `ros_buildfarm` folder and added `__init__.py` files so that they were addressable modules. Then I modified the scripts in the `scripts` folder to simply import the corresponding main functions and invoke them. Now `@ros_buildfarm_path` will be a proper prefix regardless of whether or not the package is from source or it is installed.

One other thing I noticed about the use of the current git repo for determining the source that the docker container uses for `ros_buildfarm` is that it does not handle branches, which you have to be careful to ensure is correct if you want to test this locally.

I have run the prerelease on my Trusty VM, and it works. I had trouble with permissions of shared volumes on my mac, otherwise I am confident it would work there too with boot2docker.

dirk-thomas:
#34 addresses:
* first two commits in: 4ed7912dd7f0e9ab61c5115eb0d68565f15ad468
* commits 3, 4 and 7 have been cherry-picked and squashed in: ad71332d6695e1c0f47600495699249781619733
* commit 5 has been revamped: b27f92fedc7a96d4ebd016764a98fbf12bb53016
* commit 6 has been simplified: 331a99a72628e999b93d161b9c8137dc74b06e17
* commit 8 has been replaced with: 3480d76f072a69c8bf0a7f66d602a251cf5a47be
* commit 9 has been applied: 0bab4bc51c76dd3346af4eb2953730ff6d9f75db

PR #36, opened 2015-01-29
Style fixup

wjwwood:
I think these changes (isolated to `ros_buildfarm/status_page.py`) improve the readability.

### DIFF
--- a/ros_buildfarm/status_page.py
+++ b/ros_buildfarm/status_page.py
@@ -23,8 +23,13 @@ from .templates import template_basepath
 
 
 def build_release_status_page(
-        config_url, rosdistro_name, release_build_name,
-        cache_dir, output_dir, copy_resources=False):
+    config_url,
+    rosdistro_name,
+    release_build_name,
+    cache_dir,
+    output_dir,
+    copy_resources=False,
+):
     start_time = time.localtime()
 
     config = get_config_index(config_url)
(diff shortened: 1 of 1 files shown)


wjwwood:
My feeling is that vertical space is cheap (free), but you can choose to flatten the them out. The most important visual change for those commits is this difference:

```python
# Ok
def foo_func(
        arg1, arg2):
    pass
# Better, IMO
def bar_func(
    arg1, arg2
):
    pass
```

I think the second way is better because it visually preserves the end of the function definition and makes the start of the function content clearer, but I guess that's pretty subjective.

dirk-thomas:
I cherry-picked have of the commits for now. Regarding the two remaining one mostly related to formatting we should talk about these example in the next meeting and make a decision which style we want to apply.

Example 1: the function definition from the previous comment.

Example 2: variable assignment with long values:

    foo = (
        long_value
    )

vs.

    foo = \
        long_value

Example 3: wrapped strings

    'foo' + \
    'bar'

vs.

    (
        'foo'
        'bar'
    )

wjwwood:
In assignment, the use of parenthesis is sort of ugly, but much of the time I'm assembling strings as a parameter to a function and in that case it feels more natural.

The biggest advantage to not using `+` to concatenate strings in Python, imo, is this:

```
# This will produce an error as the first `%s` is not subject to the `%` operator
a = "This is a %s, " + \
      "as you can %s." % ('string', 'see')
# This will not have a problem
b = ("This is a %s, "
       "as you can %s." % ('string', 'see'))
```

I'll close the pr now.

PR #37, opened 2015-02-10
simple docker image generator proof of concept, work in progress

tfoote:
This PR is to provide a place for a discussion of how to generate Docker base images. 

Follow up from ros-users discussion: http://lists.ros.org/pipermail/ros-users/2015-February/069172.html

### DIFF
--- a/ros_buildfarm/templates/docker_images/create_base_image.Dockerfile.em
+++ b/ros_buildfarm/templates/docker_images/create_base_image.Dockerfile.em
@@ -1,0 +1,43 @@
+# generated from @template_name
+
+@(TEMPLATE(
+    'snippet/from_base_image.Dockerfile.em',
+    os_name=os_name,
+    os_code_name=os_code_name,
+    arch=arch,
+))@
+MAINTAINER Dirk Thomas dthomas+buildfarm@@osrfoundation.org
+
+RUN locale-gen en_US.UTF-8
+ENV LANG en_US.UTF-8
+ENV TZ @timezone
+
+@(TEMPLATE(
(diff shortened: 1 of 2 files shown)


tfoote:
And I've created a repository for the Dockerfiles at: https://github.com/ros-infrastructure/docker_images which we can setup for Docker Hub to monitor. My vision is to have the set of directories with the Dockerfiles and a script to generate all of them. Some of the things we may want to consider doing is things like embedding version numbers in the images so they change when packages get updated the images get rebuilt, like we do on the buildfarm. And my current template has a lot of room to be improved as well.

dirk-thomas:
-1 since this is not related to the build farm at all. Therefore it should live in a separate repository.

Also any Docker image we provide for public use should be based on an `Automated Build` (https://docs.docker.com/docker-hub/builds/).

tfoote:
We can find a new location. There are several things which we leverage from this repo, and I would like to make sure that we don't diverge our testing and our image generation which is why I put it here for now.  

This will use the automated builds, this is for generating the Dockerfiles for the automated builds.

dirk-thomas:
Any other location is fine - it can use anything from this repository since it is released as a Python package.

ruffsl:
@tfoote, 
Have you moved this to a different repo?
Where are you working on this now, [docker_images](https://github.com/ros-infrastructure/docker_images)?

tfoote:
@ruffsl I've copied this code into [docker_images](https://github.com/ros-infrastructure/docker_images). It's not currently running as the template logic is searching in the wrong directories at the moment. And I haven't had a chance to clean up the paths for the new locations. 

I'll close this for now and we can discusson more in that repository.

PR #42, opened 2015-02-25
Update git scm checkout to support branches with slashes

tfoote:
From: https://issues.jenkins-ci.org/browse/JENKINS-27115 and https://github.com/ros/rosdistro/issues/7346

It looks like using the wildcard syntax */branch/name should work. 

By default the git plugin trims after a slash in the branch name. 

@dirk-thomas I think this is the right place.

### DIFF
--- a/ros_buildfarm/templates/snippet/scm_git.xml.em
+++ b/ros_buildfarm/templates/snippet/scm_git.xml.em
@@ -10,7 +10,7 @@
     </userRemoteConfigs>
     <branches>
       <hudson.plugins.git.BranchSpec>
-        <name>@ESCAPE(branch_name)</name>
+        <name>refs/heads/@ESCAPE(branch_name)</name>
       </hudson.plugins.git.BranchSpec>
     </branches>
     <doGenerateSubmoduleConfigurations>false</doGenerateSubmoduleConfigurations>


dirk-thomas:
I am not sure how this reg exp will match other things. Are you sure that it matches exactly what is desired and never overmatches / undermatches?

tfoote:
From https://issues.jenkins-ci.org/browse/JENKINS-20767

```
... "refs/heads<branchSpec>" if wildcards are used (e.g. "*/master" -> "refs/heads*/master". "refs/heads/*/master" would not match the "master" branch)
```

My understanding is that * matches anything except `/` ** matches anything including `/` As long as we don't have multiple remotes this should not be ambiguous.  

The default for adding a new branch is also */master

dirk-thomas:
From reading the documentation (https://github.com/jenkinsci/git-plugin/blob/master/src/main/resources/hudson/plugins/git/BranchSpec/help-name.html) I am worried about the following sentence:

    In addition, BRANCH is recognized as a shorthand of */BRANCH

That sounds to me that specifying `*/master` instead of `master` would not be sufficient to match a branch called `feature/master`. Therefore I don't feel comfortable merging this without having checked those cases with an example repository / job.

tfoote:
That did overmatch. It turns out that they don't do the truncation on `refs/heads/<branch/name>` so I've updated the PR to follow that convention.

dirk-thomas:
I think the suffix `refs/heads/` does only work for branch names. But in our case the version value might be a branch name, a tag name or even a hash. And since we don't know the type I don't think using that prefix would work.

tfoote:
This is a problem upstream. I think we cannot solve this by working around it. Closing as wontfix for now.

PR #50, opened 2015-03-29
add-on forks require REP143 index files

mikeferguson:


mikeferguson:
I realized after making this pull request that the README mentions REP-143, but not the custom_rosdistro.md

PR #53, opened 2015-04-05
Add ccache inside debbuild and devel jobs

tfoote:
This will save a lot of processing time. 

On amd64 it takes a 2.5minute build of roscpp to a 12 second rebuild. 
On armhf it takes a 29 minute build of roscpp to 2.1 minute rebuild.

This will only share the cache on a single host. But that will be shared across the 4 executors at the moment. And jenkins does seem to have an affinity for reusing the same executor, but if we're queued up I don't think it takes that into account. 

This requires: https://github.com/ros-infrastructure/buildfarm_deployment/pull/60 for clean startup.

### DIFF
--- a/ros_buildfarm/templates/devel/devel_task.Dockerfile.em
+++ b/ros_buildfarm/templates/devel/devel_task.Dockerfile.em
@@ -52,6 +52,8 @@ RUN echo "@today_str"
     os_code_name=os_code_name,
 ))@
 
+RUN python3 -u /tmp/wrapper_scripts/apt-get.py update-and-install -q -y ccache
+
 @(TEMPLATE(
     'snippet/install_dependencies.Dockerfile.em',
     dependencies=dependencies,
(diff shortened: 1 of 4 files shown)


dirk-thomas:
This will not have any effect as far as I know. Ccache only installs a ccache binary but does not "override" your default compilers. Have you tried this setup?

tfoote:
It injects /usr/lib/ccache onto the path [(such as here)](https://github.com/ros-infrastructure/ros_buildfarm/pull/53/files#diff-d0cd65cef4f57542d3cccb7c4e1d7d59R60) which has the necessary symlinks. 

I ran the jobs locally for testing to get the above timings.

dirk-thomas:
Oh, that is a very different setup to what I am usually having setup on my machine. Sounds good!

ros_buildfarm/templates/devel/devel_task.Dockerfile.em
```diff
@@ -55,14 +57,14 @@ RUN echo "@today_str"
 USER buildfarm
 ENTRYPOINT ["sh", "-c"]
 @{
+cmd = 'PATH=/usr/lib/ccache:$PATH' + \
```
dirk-thomas: This lacks a trailing whitespace within the string.

ros_buildfarm/templates/devel/devel_task.Dockerfile.em
```diff
@@ -55,14 +57,14 @@ RUN echo "@today_str"
...
+      'PYTHONPATH=/tmp/ros_buildfarm:$PYTHONPATH python3 -u '
```
dirk-thomas: The indentation seems to be off.

ros_buildfarm/templates/devel/devel_task.Dockerfile.em
tfoote: Good catch. I made all the command extensions consistently prefixed with whitespace to be consistent with the debbuild job, and internally consistent.

ros_buildfarm/templates/devel/devel_task.Dockerfile.em
```diff
@@ -61,18 +63,18 @@ RUN echo "@today_str"
 USER buildfarm
 ENTRYPOINT ["sh", "-c"]
 @{
+cmd = 'PATH=/usr/lib/ccache:$PATH' + \
+      ' PYTHONPATH=/tmp/ros_buildfarm:$PYTHONPATH python3 -u'
```
dirk-thomas: The indentation is still off. The quotes should not align but the line should simply be indented one level aka four spaces. Same below with the cmd concatenation.

tfoote:
Updated indention to 4 spaces on wrap.

dirk-thomas:
lgtm

tfoote:
This saved a lot of time on roscpp: http://54.183.26.131:8080/view/Jbin_uT64/job/Jbin_uT64__roscpp__ubuntu_trusty_amd64__binary/buildTimeTrend

roscpp_ccache_speedup

Issue #58, opened 2015-04-24
Saucy was just pulled from archive.ubuntu.com


tfoote:
We need to either switch to http://old-releases.ubuntu.com/releases/ or turn them off.

@dirk-thomas @wjwwood FYI

tfoote:
The saucy tag is not on the supported list any more: https://registry.hub.docker.com/_/ubuntu/

We may need to roll our own now. Otherwise there's only a few places we will need to update from a quick search: https://github.com/ros-infrastructure/ros_buildfarm/search?utf8=%E2%9C%93&q=archive&type=Code

dirk-thomas:
Since we have not communicated this topic with the community we have to continue building Saucy packages for now. Please go ahead and make a PR for this. I have already patched the old farm last night: https://github.com/ros-infrastructure/buildfarm/commit/1c88f7640eba35675e2fa2011f51a0c61fb6008a

tfoote:
I've tested this sed script on a local docker container and it gets apt running
`find /etc/apt/ -name *.list -exec sed -i -e 's/archive.ubuntu.com\|security.ubuntu.com/old-releases.ubuntu.com/g'  {} \;`

Modified [from here](http://unix.stackexchange.com/questions/146551/apt-get-cant-find-the-package-on-a-old-version-of-my-distro)

We will just need to embed that into the Dockerfiles for old-releases.

PR #60, opened 2015-04-24
host ourselves the required version of doxygen

tfoote:
instead of relying on archives.ubuntu.com (preventative re #58)


tfoote:
@dirk-thomas for review

dirk-thomas:
We are currently using the Precise version of `doxygen` which will be around for many more years. Therefore I would rather not start hosting it on download.ros.org.

PR #62, opened 2015-04-27
retry docker pull when it fails with known errors

dirk-thomas:
I created this PR based on @wjwwood report: http://54.183.26.131:8080/job/Jrel_arm_release-status-page/7266/

    Get base image name from Dockerfile 'Dockerfile': ubuntu:trusty
    Check docker base image for updates: docker pull ubuntu:trusty
    Pulling repository ubuntu
    ...
    14f14d3d9: Error pulling dependent layers
    b7cf8f0d9e82: Error pulling image (trusty) from ubuntu, endpoint: https://registry-1.docker.io/v1/, Server error: Status 502 while fetching image layer (2c014f14d3d95811df672ddae2af376f9557f6b8f5623e3e3f8c5ca3f6ff42e6)
    b7cf8f0d9e82: Error pulling image (trusty) from ubuntu, Server error: Status 502 while fetching image layer (2c014f14d3d95811df672ddae2af376f9557f6b8f5623e3e3f8c5ca3f6ff42e6)

But I am not sold on the fact that we should try to catch and retry those.

@tfoote @wjwwood RFC.

### DIFF
--- a/scripts/misc/docker_pull_baseimage.py
+++ b/scripts/misc/docker_pull_baseimage.py
@@ -2,6 +2,7 @@
 
 import subprocess
 import sys
+from time import sleep
 
 
 def main(argv=sys.argv[1:]):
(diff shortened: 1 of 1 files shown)


tfoote:
It seems like it shouldn't be necessary, but as we see it regularly and can allow us to avoid bothering user's about these intermittent failures we should make it happen. 

lgtm

PR #68, opened 2015-05-04
adding additional labels to provide better introspection

tfoote:
This adds a snippet which adds executor labels which will not be actually used, but can provide charts like I use here: https://rawgit.com/tfoote/b7e87790e41d6f383755/raw/ros_multi_load.html

Then use the snippet in each of the job build types. 

@dirk-thomas Please review. Is there a way to test this without deploying?

### DIFF
--- a/ros_buildfarm/templates/doc/doc_independent_job.xml.em
+++ b/ros_buildfarm/templates/doc/doc_independent_job.xml.em
@@ -21,7 +21,13 @@
 @(SNIPPET(
     'scm_null',
 ))@
-  <assignedNode>@(node_label if node_label else 'buildslave')</assignedNode>
+@(SNIPPET(
+    'assigned_node',
+    jobtype='doc_independent',
+    node_label=node_label,
+    rosdistro_name=rosdistro_name,
+    arch=arch,
+))@
   <canRoam>false</canRoam>
   <disabled>false</disabled>
   <blockBuildWhenDownstreamBuilding>false</blockBuildWhenDownstreamBuilding>
--- a/ros_buildfarm/templates/doc/doc_metadata_job.xml.em
+++ b/ros_buildfarm/templates/doc/doc_metadata_job.xml.em
@@ -21,7 +21,13 @@
 @(SNIPPET(
     'scm_null',
 ))@
-  <assignedNode>@(node_label if node_label else 'buildslave')</assignedNode>
+@(SNIPPET(
+    'assigned_node',
+    jobtype='doc_metadata',
+    node_label=node_label,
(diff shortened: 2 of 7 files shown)


ros_buildfarm/templates/snippet/assigned_node.xml.em
```diff
@@ -0,0 +1 @@
+    <assignedNode>@(node_label if node_label else 'buildslave') || @('_'.join(['', job_type, rosdistro_name, arch])) || @('_'.join(['', job_type, rosdistro_name])) || @('_'.join(['', job_type, arch]))</assignedNode>
```
dirk-thomas: Please wrap this logic into separate lines with individual conditions. In the current form it is extremely difficult to read.

ros_buildfarm/templates/doc/doc_independent_job.xml.em
```diff
@@ -21,7 +21,13 @@
 @(SNIPPET(
     'scm_null',
 ))@
-  <assignedNode>@(node_label if node_label else 'buildslave')</assignedNode>
+@(SNIPPET(
+    'assigned_node',
+    jobtype='docind',
```
dirk-thomas: Please avoid unnecessary abbreviations. Just use `doc_independent`.

ros_buildfarm/templates/doc/doc_metadata_job.xml.em
```diff
@@ -21,7 +21,13 @@
 @(SNIPPET(
     'scm_null',
 ))@
-  <assignedNode>@(node_label if node_label else 'buildslave')</assignedNode>
+@(SNIPPET(
+    'assigned_node',
```
dirk-thomas: Please avoid unnecessary abbreviations. Just use `doc_metadata`.

ros_buildfarm/templates/snippet/assigned_node.xml.em
dirk-thomas: Also the indentation looks wrong.

dirk-thomas:
Can you please share the URL how to access these stats on the current test farm.

tfoote:
Updated PR to resolve comments.

dirk-thomas:
You should try to generate job xml for each modified job type to make sure it works. You can either use a local test farm or modify the code to not push changes to the Jenkins master but perform everything else.

From reading the diff I would expect it to fail for some job types since not all of them have an `arch`.

dirk-thomas:
If this would be helpful to visualize the current load more granular it should be updated and deployed.

dirk-thomas:
Superseded by #317.

Issue #64, opened 2015-04-29
reenable slaves after having enough disk space again


dirk-thomas:
The slaves are being set to "offline temporarily due to the lack of disk space" successfully. But they don't reenable themselves when disk space has been recovered later.

While visiting the "Manage Nodes" page of Jenkins is enough to reenable them (you don't even have to click the "refresh" button) it should happen without user intervention.

I have created a test job running a groovy script to check if that reenables the slaves without user interaction: http://54.183.26.131:8080/view/Manage/job/check_disk_space_of_disabled_slaves/

dirk-thomas:
It looks like that this job has proven to be valuable. Recently a slave was disabled due to low free disk space but was reenabled by the the job shortly after it had recovered disk space again:

    May 01, 2015 5:30:50 PM WARNING java_util_logging_Logger$warning call
    Making ip-172-31-18-229.us-west-1.compute.internal offline temporarily due to the lack of disk space
    May 01, 2015 5:36:00 PM WARNING hudson.node_monitors.AbstractDiskSpaceMonitor data
    Putting ip-172-31-18-229.us-west-1.compute.internal back online as there is enough disk space again

PR #69, opened 2015-05-04
add job to check if disabled slaves have recovered from not having enough free disk space

dirk-thomas:
Addresses #64.

The job has been created using the script in this PR: http://54.183.26.131:8080/view/Manage/job/check_slaves/

@tfoote Please review.


tfoote:
+1

PR #73, opened 2015-05-14
Fix generate release cmdline

tfoote:
I was using this for testing and needed the following tweaks to make it work. 

@dirk-thomas for review

### DIFF
--- a/ros_buildfarm/argument.py
+++ b/ros_buildfarm/argument.py
@@ -32,7 +32,7 @@ def add_argument_rosdistro_name(parser):
 def add_argument_build_name(parser, build_file_type):
     parser.add_argument(
         '%s_build_name' % build_file_type,
-        help="The name of the '%s-build' file from the index" %
+        help="The name / key of the '%s-build' file from the index" %
              build_file_type)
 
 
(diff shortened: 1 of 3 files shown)


ros_buildfarm/argument.py
```diff
@@ -32,7 +32,7 @@ def add_argument_rosdistro_name(parser):
 def add_argument_build_name(parser, build_file_type):
     parser.add_argument(
         '%s_build_name' % build_file_type,
-        help="The name of the '%s-build' file from the index" %
+        help="The name(key) of the '%s-build' file from the index" %
```
dirk-thomas: `name(key)` sounds a bit weird to me. Would `name / key` be better?

scripts/release/generate_release_job.py
```diff
@@ -8,7 +8,7 @@
 from ros_buildfarm.argument import add_argument_os_code_name
 from ros_buildfarm.argument import add_argument_os_name
 from ros_buildfarm.argument import add_argument_package_name
-from ros_buildfarm.argument import add_argument_rosdistro_index_url
+from ros_buildfarm.argument import add_argument_config_url
```
dirk-thomas: Please keep the modified import in alphabetical order to follow PEP8.

dirk-thomas:
I applied the suggested changes and squashed the commit. Merging...

PR #81, opened 2015-06-17
Prevent release script from failing on non-existent *.debian.tar.xz

ipa-mdl:
https://github.com/ros-infrastructure/ros_buildfarm/blob/master/doc/jobs/release_jobs.rst#example-invocation fails with
```
cp: cannot stat '/home/mdl/buildfarm/source/sourcedeb/*.debian.tar.xz': No such file or directory
```
This patched version matches *.debian.tar.xz and/or *.debian.tar.gz.


tfoote:
Thanks I just ran into this too.

PR #82, opened 2015-06-19
Docker api

ruffsl:
Adding --rm to docker run command in templates,
should help ease troubles with the docker reaper.

### DIFF
--- a/ros_buildfarm/templates/devel/devel_job.xml.em
+++ b/ros_buildfarm/templates/devel/devel_job.xml.em
@@ -152,6 +152,7 @@ if pull_request:
         'mkdir -p $WORKSPACE/docker_build_and_install',
         'mkdir -p $WORKSPACE/docker_build_and_test',
         'docker run' +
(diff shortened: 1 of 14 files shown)


tfoote:
LGTM 

We never reattach to the containers so we might as well clean them up immediately instead of leaving them on the disk. This hopefully will help with the problem of memory growth when the cleanup script is invoked by having less in the docker databases. 

@dirk-thomas can you review as well. Then we can turn a large batch over and see if it helps.

dirk-thomas:
Does this also imply removing the images used by the container?

tfoote:
No this only clears the containers.
​

dirk-thomas:
How does removing the container impact the storage size then?

tfoote:
It doesn't significantly effect the storage size. Containers are typically under 100K in size.

dirk-thomas:
Sounds like a minor change then. Let's go ahead and see if it brings any advantage.

PR #88, opened 2015-07-02
[doc] sudo was needed to run ./prerelease.sh

130s:
Besides this change, I don't know if there's usage document for the new docker-based prerelease ([this wiki](http://wiki.ros.org/regression_tests#How_do_I_setup_my_system_to_run_a_prerelease.3F) doesn't tell what to do after some point). I'm not sure if this repository needs to be pointed from there either.


dirk-thomas:
I am pretty sure that `sudo` is not required for the command. You might want to check your Docker setup that your user is in the right group to be able to run docker container: https://docs.docker.com/installation/ubuntulinux/#create-a-docker-group

tfoote:
You should set it up that you're in the docker group instead of using sudo. 

I added a note to follow those instructions on the linked wiki page.

130s:
I confirmed that the updated wiki works.

PR #93, opened 2015-08-16
Use Release targets for pre-releases (fixes #90)

### LINKED ISSUE #90: prerelease_script: Problem with looking up valid OS codes
Apparantly there is a problem in looking up valid OS codes when creating pre-release jobs for distros other than local build machine's distro. My local build machine runs 64 bit _Ubuntu Trusty_ and has been setup to run prerelease jobs based on [the documentation on ROS wiki](http://wiki.ros.org/regression_tests#How_do_I_setup_my_system_to_run_a_prerelease.3F).

The prerelease command works fine when target distro is set to _Trusty_. However, when the target distro is set to _Vivid_ or _Utopic_ the generated command fails with the following error:

```
$ generate_prerelease_script.py \
>   https://raw.githubusercontent.com/ros-infrastructure/ros_buildfarm_config/master/index.yaml \
>   jade default ubuntu utopic amd64 \
>   ardrone_autonomy \
>   --level 1 \
>   --output-dir .
Fetching buildfarm configuration...
Fetching rosdistro cache...
Evaluating job templates...
  File "/usr/bin/generate_prerelease_script.py", line 263, in <module>
    sys.exit(main())
  File "/usr/bin/generate_prerelease_script.py", line 157, in main
    source_repository=source_repository)
  File "/usr/lib/python2.7/dist-packages/ros_buildfarm/devel_job.py", line 237, in configure_devel_job
    ', '.join(sorted(build_file.targets[os_name].keys())))
ros_buildfarm.common.JobValidationError
```

Is there any way to setup a local and [minimal development environment to debug this issue? If so, I can try to track down the bug and fix it.

mani-monaj:
For pre-release jobs, overwrite `build.targets` feteched from Source build config file with targets fetched from corresponding Release config file. This fixes invalid OS code lookups for pre-release jobs (reported in #90).

### DIFF
--- a/ros_buildfarm/devel_job.py
+++ b/ros_buildfarm/devel_job.py
@@ -183,7 +183,8 @@ def configure_devel_job(
         jenkins=None, views=None,
         is_disabled=False,
         groovy_script=None,
-        source_repository=None):
+        source_repository=None,
+        build_targets=None):
     """
     Configure a single Jenkins devel job.
 
@@ -198,6 +199,9 @@ def configure_devel_job(
     if build_file is None:
         build_files = get_source_build_files(config, rosdistro_name)
         build_file = build_files[source_build_name]
+    # Overwrite build_file.targets if build_targets is specified
+    if build_targets is not None:
+        build_file.targets = build_targets
 
     if index is None:
         index = get_index(config.rosdistro_index_url)
--- a/scripts/prerelease/generate_prerelease_script.py
+++ b/scripts/prerelease/generate_prerelease_script.py
@@ -23,6 +23,7 @@ from ros_buildfarm.argument import add_argument_os_name
 from ros_buildfarm.argument import add_argument_output_dir
 from ros_buildfarm.argument import add_argument_rosdistro_name
 from ros_buildfarm.config import get_index as get_config_index
+from ros_buildfarm.config import get_release_build_files
 from ros_buildfarm.config import get_source_build_files
 from ros_buildfarm.devel_job import configure_devel_job
 from ros_buildfarm.prerelease import add_overlay_arguments
@@ -78,6 +79,22 @@ def main(argv=sys.argv[1:]):
     build_file = build_files[args.source_build_name]
 
     print('Fetching rosdistro cache...')
+    # Targets defined by source build file are subset of targets
+    # defined by release build files. To increase the number of supported
+    # pre-release targets, we combine all targets defined by all release
+    # build files and use that when configuring the devel job.
+    release_build_files = get_release_build_files(config, args.rosdistro_name)
+    release_targets_combined = {}
+    if release_build_files:
+        release_targets_combined[args.os_name] = {}
+        for build_name, rel_obj in release_build_files.items():
+            if args.os_name not in rel_obj.targets:
+                continue
+            for dist_name, targets in rel_obj.targets[args.os_name].items():
+                if dist_name not in release_targets_combined[args.os_name]:
+                    release_targets_combined[args.os_name][dist_name] = {}
+                release_targets_combined[args.os_name][dist_name].update(targets)
+
     index = get_index(config.rosdistro_index_url)
     dist_cache = get_distribution_cache(index, args.rosdistro_name)
     dist_file = dist_cache.distribution_file
@@ -154,7 +171,8 @@ def main(argv=sys.argv[1:]):
         config=config, build_file=build_file,
         index=index, dist_file=dist_file, dist_cache=dist_cache,
         jenkins=False, views=False,
-        source_repository=source_repository)
+        source_repository=source_repository,
+        build_targets=release_targets_combined)
 
     templates.template_hooks = None
 


tfoote:
This is a neat way to mutate the build_file. However I'd prefer it to be explicitly passed instead of mutating the build_file object, there may be other uses of the build_file targets in the future which would want the original data from that field and might incorrectly assume the data is the original. I don't think passing the additional data will be too hard, and it will make the maintainability much better.

mani-monaj:
Sounds fair. I'll work on it again later this week and will update this PR.

mani-monaj:
@tfoote I implemented your suggestion and updated this PR (it's a single squashed commit).

ros_buildfarm/devel_job.py
```diff
@@ -198,6 +200,9 @@ def configure_devel_job(
     if build_file is None:
         build_files = get_source_build_files(config, rosdistro_name)
         build_file = build_files[source_build_name]
+    # Overwrite build_file.targets if build_targets is specified
+    if not build_targets is None:
```
tfoote: A slight style note: PEP8 calls for 'is not None' it's functionally the same but cleaner to read.

tfoote:
This looks good to me. One stylistic comment above.

mani-monaj:
Thanks @tfoote. I fixed the stylistic issue and updated the PR.

tfoote:
lgtm, can someone else review?

scripts/prerelease/generate_prerelease_script.py
```diff
@@ -77,6 +78,12 @@ def main(argv=sys.argv[1:]):
     build_files = get_source_build_files(config, args.rosdistro_name)
     build_file = build_files[args.source_build_name]
 
+    # Use targets defined by corresponding Release config instead of
+    # targets defined by Source config. This is to increase the number of
+    # supported targets. New targets are explicitly passed to configure_devel_job().
+    release_build_files = get_release_build_files(config, args.rosdistro_name)
+    release_build_file = release_build_files[args.source_build_name]
```
dirk-thomas: This won't work well since the `source_build_name` might not be the same. E.g. their is an `arm` release build file with no equivalent source build file.

ros_buildfarm/devel_job.py
```diff
@@ -1,6 +1,7 @@
 from __future__ import print_function
 
 import sys
+from copy import deepcopy
```
dirk-thomas: Please keep imports in alphabetical order following PEP8.

scripts/prerelease/generate_prerelease_script.py
```diff
@@ -24,6 +24,7 @@
 from ros_buildfarm.argument import add_argument_rosdistro_name
 from ros_buildfarm.config import get_index as get_config_index
 from ros_buildfarm.config import get_source_build_files
+from ros_buildfarm.config import get_release_build_files
```
dirk-thomas: Please keep imports in alphabetical order following PEP8.

scripts/prerelease/generate_prerelease_script.py
mani-monaj: You mean `arg.source_build_name` might not be a key in `release_build_files`? If so, is the following a safer alternative:

```python
if release_build_files:
    release_build_file = release_build_files.get(args.source_build_name, None)
```

ros_buildfarm/devel_job.py
mani-monaj: Sure. I'll fix it.

scripts/prerelease/generate_prerelease_script.py
mani-monaj: Sure. I'll fix it.

scripts/prerelease/generate_prerelease_script.py
dirk-thomas: This won;t be sufficient. The index file might define a `default` key for source builds but only a `linux` and `fedora` key for release. There is simply no correlation between those.

mani-monaj: > This won't be sufficient. The index file might define a default key for source builds but only a linux and fedora key for release. There is simply no correlation between those.

In this case and considering the code I recommended `release_build_file` will be set to `None` which is a fallback to the default behavior. Pre-release script will use `build_file = build_files[args.source_build_name]` as the build target (line 79).

dirk-thomas: This won't fix the problem you are aiming to address.

As mentioned above collecting the targets from all release build files seems necessary.

mani-monaj: @dirk-thomas 

`generate_prerelease_script.py` is called with the following parameterts:

```
generate_prerelease_script.py \
  <config_url:https://raw.githubusercontent.com/ros-infrastructure/ros_buildfarm_config/master/index.yaml> \
  <ros_distro:jade> <build_name:default> <os_name:ubuntu> <os_code_name:vivid> <arch:amd64> \
  <pkg_name:ardrone_autonomy> \
  --level 1 \
  --output-dir ./
```

As far as I understood, the script first parses the index file at `config_url` and looks up the `ros_distro` key. Next it looks up `ros_distro/source_builds` key and fetches the YAML config file specified by `ros_disto/source_builds/build_name`. Finally in that config file, it check if `targets/os_name/os_code_name/arch` exists. The problem I am trying to solve here is when the fetched YAML file from `ros_distro/source_builds/build_name` does not define all the targets that exist in the YAML file specified by `ros_distro/release_builds/build_name` key. 

Could you please give some examples of other failure cases? One other failure case that comes to my mind is when the user sets `build_name` to something that is only defined in the release build config file. In case of default index files, `build_name:armhf, arch:armhf` will break the pre-release script.

dirk-thomas: The problem is that there is no correlation between the `build_name` you are using for `source_builds` and `release_builds`. The following `index.yaml` is perfectly valid:

    distributions:
      jade:
        release_builds:
          some_release_build: jade/some-release-build.yaml
          other_release_build: jade/other-release-build.yaml
        source_builds:
          default_source_builds: jade/default-source-build.yaml

Without considering the targets of all release build files you won't be able to use more then the small subset of targets found in the source build file.

mani-monaj: @dirk-thomas Thanks for clarifying the situation. How about something along this line to merge all the targets:

```python
    release_build_files = get_release_build_files(config, args.rosdistro_name)
    release_targets_combined = {}
    release_targets_combined[args.os_name] = {}
    for arch, rel_obj in release_build_files.items():
        for os, t in rel_obj.targets[args.os_name].items():
          if os not in release_targets_combined[args.os_name]:
            release_targets_combined[args.os_name][os] = {}
          release_targets_combined[args.os_name][os].update(t)

  # ...
  # pass release_targets_combined to configure_devel_job()
```

dirk-thomas: Yes, that looks like the best approach. And since the targets defined in the source build file must always be a subset we don't have to consider them.

mani-monaj: Great. I'll refactor/test this, fix minor styling issues and then update this PR.

mani-monaj:
@dirk-thomas I cleaned up the code to merge all build targets, squashed the changes and updated this PR. It's working fine on my machine, but is it possible for you to review it again and/or test it? Is there any test cases that I can run the script against?

ros_buildfarm/devel_job.py
```diff
@@ -198,6 +200,9 @@ def configure_devel_job(
     if build_file is None:
         build_files = get_source_build_files(config, rosdistro_name)
         build_file = build_files[source_build_name]
+    # Overwrite build_file.targets if build_targets is specified
+    if build_targets is not None:
+        build_file.targets = deepcopy(build_targets)
```
dirk-thomas: I don't think the `deepcopy` is necessary here. The caller should not modify the argument after passing it anyway.

scripts/prerelease/generate_prerelease_script.py
```diff
@@ -78,6 +79,22 @@ def main(argv=sys.argv[1:]):
     build_file = build_files[args.source_build_name]
 
     print('Fetching rosdistro cache...')
+    # Targets defined by source config file is a subset of targets
+    # defined by release config files. To increase the number of supported
+    # pre-release targets, we combine all targets defined by all release
+    # config files and use that when configuring the devel job.
```
dirk-thomas: Can you please update the comment to use the term `build file` instead of `config file` since that is the more precise type of these yaml files.

dirk-thomas:
Thank you for updating the PR. I just tried the patch and it works well. I just made two minor comments which should be addressed before the merge.

ros_buildfarm/devel_job.py
mani-monaj: You are right. I fixed it.

scripts/prerelease/generate_prerelease_script.py
mani-monaj: Sure. fxed.

mani-monaj:
@dirk-thomas You are welcome. I applied your fixes and updated the PR.

PR #72, opened 2015-05-13
Docker template

ruffsl:
Adding support for Dockerfile generation:
- Support for prioritizing template packages to be use in finding template resources
- And unit tests for newly added sub functions and total Dockerfile generation

PR #96, opened 2015-08-25
search templates in a list of prefix paths

dirk-thomas:


tfoote:
This works, but using the global is a shortcut. There is an extended implementation in #72 which includes this functionality. The template_prefix_path can be passed into the expand_template method via the `data` argument dict. And the search done inside with a default value as currently defined for fallback.

dirk-thomas:
The referenced PR is more then 3 month old and is not being actively proceeded. I would also consider mixing the template location with the actual data a very bad practice.

If there are no problems with the patch proposed in this PR I will go ahead and merge it soon.

tfoote:
Yes, I'm not saying use the PR, but that using the global is a shortcut which doesn't require too significant a workaround. Passing the search path could also be added as an optional argument instead of in the data dict. Looking at the use case in #72 being able to give specific search paths is valuable, if we do this now as a global it will mean that changing it to be explicitly evaluated will break the code paths relying on the global which are very hard to find. It only requires passing the search path one layer deep (into `expand_template`). 

We should also be looking at using pkg_resources instead of [hard coding paths](https://github.com/ros2/ros2/pull/97/files#diff-42a7cd652d15099c3f6b6c27e05dce31R39) which are not as portable (such as cannot be used installed as an egg). This is what pkg_resources is designed for and is core python functionality.

dirk-thomas:
The change would not be limited to a single layer but it must be passed along to recursive templates and snippets. That's why it is a global variable in the first place.

While `pkg_resources` is certainly of value it needs to be possible to use the template lookup without. E.g. the ci_scripts branch can't use pkg_resources since it is not even a complete Python package.

tfoote:
This broke the status jobs: http://54.183.26.131:8080/job/Irel_repos-status-page/17770/

PR #100, opened 2015-09-18
Add support for custom rosdep URLs in devel builds (fixes #87)

### LINKED ISSUE #87: devel jobs are not using custom rosdep repository
This is a report from Bosch and ticketed as suggested by Tully.

We have a custom rosdep yaml (containing the output from the Mike Purvis’ rosdep-generator). This is properly used by the buildfarm’s binary-deb jobs, however, it is not used by the devel-builds in favor of the public rosdep repo.

This is the output of the “Idev_deepfield_ubuntu_trusty_amd64” job. As you can see, it does not pull the rosdep .yaml files from our server (“bfrepo”), and then fails because it cannot resolve the actionlib package:

 
```
23:36:45 reading in sources list data from /etc/ros/rosdep/sources.list.d

23:36:45 Hit https://raw.githubusercontent.com/ros/rosdistro/master/rosdep/osx-homebrew.yaml

23:36:45 Hit https://raw.githubusercontent.com/ros/rosdistro/master/rosdep/base.yaml

23:36:45 Hit https://raw.githubusercontent.com/ros/rosdistro/master/rosdep/python.yaml

23:36:45 Hit https://raw.githubusercontent.com/ros/rosdistro/master/rosdep/ruby.yaml

23:36:45 Hit https://raw.githubusercontent.com/ros/rosdistro/master/releases/fuerte.yaml

23:36:45 Query rosdistro index http://bfrepo/rosdistro/index.yaml

23:36:45 Add distro "indigo"

23:36:45 updated cache in /home/buildfarm/.ros/rosdep/sources.cache

23:36:45 Crawling for packages in workspace '/tmp/catkin_workspace/src'

23:36:46 Found the following packages:

23:36:46   - df-package1

23:36:46   - …

23:36:46 Package maintainer emails: …

23:36:48 Always install the following generic dependencies:

23:36:48   - build-essential

23:36:48   - python3

23:36:48   - ros-indigo-catkin

23:36:48 Identified the following build dependencies (ignoring packages available from source):

23:36:48   - actionlib

23:36:48   - …                         

23:36:48 Traceback (most recent call last):

23:36:48   File "/tmp/ros_buildfarm/scripts/devel/create_devel_task_generator.py", line 196, in resolve_names

23:36:48     rosdep_key, view, installer, os_name, os_code_name)

23:36:48   File "/usr/lib/python3/dist-packages/rosdep2/catkin_support.py", line 81, in resolve_for_os

23:36:48     d = view.lookup(rosdep_key)

23:36:48   File "/usr/lib/python3/dist-packages/rosdep2/lookup.py", line 187, in lookup

23:36:48     return self.rosdep_defs[rosdep_name]

23:36:48 KeyError: 'actionlib'

23:36:48

23:36:48 During handling of the above exception, another exception occurred:

23:36:48

23:36:48 Traceback (most recent call last):

23:36:48   File "/tmp/ros_buildfarm/scripts/devel/create_devel_task_generator.py", line 212, in <module>

23:36:48     main()

23:36:48   File "/tmp/ros_buildfarm/scripts/devel/create_devel_task_generator.py", line 98, in main

23:36:48     debian_pkg_names_building = resolve_names(build_depends, **context)

23:36:48   File "/tmp/ros_buildfarm/scripts/devel/create_devel_task_generator.py", line 199, in resolve_names

23:36:48     "Could not resolve the rosdep key '%s'" % rosdep_key)

23:36:48 RuntimeError: Could not resolve the rosdep key 'actionlib'
```
 

Why do the devel builds not use our custom rosdep repository like the release-builds do? Is this missing somewhere in our configuration files?

jjekircp:
I went by [tfoote's comment on issue #87](https://github.com/ros-infrastructure/ros_buildfarm/issues/87#issuecomment-117556312) suggesting to use the _config section of the targets in the release build file to get a list of custom rosdep URLs.
I then pass that variable through to the devel job, where I wget those URLs into /etc/ros/rosdep/sources.list.d. If no custom rosdep URLs are provided the job runs "rosdep init" as usual.

Here is the targets: entry I use in ros_buildfarm_config/indigo/source_build.yml that makes this work:
```
targets:
  _config:
    custom_rosdep_urls:
      - http://my-git-server/rosdistro-mine/raw/master/rosdep/sources.list.d/20-default.list
```


gregwjacobs:
+1

tfoote:
Thanks this looks reasonable. I want to get a moment to test it before I merge it. 

@ddimarco @dejanpan FYI.

dirk-thomas:
The PR should also update the documentation and describe the new configuration options: https://github.com/ros-infrastructure/ros_buildfarm/blob/master/doc/configuration_options.rst

ros_buildfarm/common.py
```diff
@@ -51,6 +53,8 @@ def get_repositories_and_script_generating_key_files(
             len(build_file.repository_keys)
         repository_urls += build_file.repository_urls
         repository_keys += build_file.repository_keys
+        if hasattr(build_file, "custom_rosdep_urls"):
```
dirk-thomas: Please use single quotes wherever possible. Same multiple times below.

Regarding the `hasattr` I will comment below.

ros_buildfarm/config/source_build_file.py
```diff
@@ -92,6 +92,10 @@ def __init__(self, name, data):
             self.skip_ignored_repositories = \
                 bool(data['skip_ignored_repositories'])
 
+        if '_config' in data['targets']:
+            if "custom_rosdep_urls" in data['targets']['_config']:
+                self.custom_rosdep_urls = data['targets']['_config']['custom_rosdep_urls']
```
dirk-thomas: Please add an assertion to ensure the expected type.

ros_buildfarm/templates/devel/devel_create_tasks.Dockerfile.em
```diff
@@ -20,6 +20,7 @@ RUN useradd -u @uid -m buildfarm
     add_source=False,
 ))@
 
+
```
dirk-thomas: Please avoid white space only changes.

ros_buildfarm/templates/devel/devel_create_tasks.Dockerfile.em
```diff
@@ -28,12 +29,16 @@ RUN useradd -u @uid -m buildfarm
...
-RUN python3 -u /tmp/wrapper_scripts/apt-get.py update-and-install -q -y git python3-apt python3-catkin-pkg python3-empy python3-rosdep python3-rosdistro
+RUN python3 -u /tmp/wrapper_scripts/apt-get.py update-and-install -q -y git python3-apt python3-catkin-pkg python3-empy python3-rosdep python3-rosdistro wget
 
 # always invalidate to actually have the latest rosdep state
 RUN echo "@now_str"
 ENV ROSDISTRO_INDEX_URL @rosdistro_index_url
-RUN rosdep init
+
+@(TEMPLATE(
+    'snippet/add_custom_rosdeps.Dockerfile.em',
```
dirk-thomas: Since the task of the snippet is more general then adding custom rosdeps I would propose to call it "rosdep_init.Dockerfile.em".

scripts/devel/run_devel_job.py
```diff
@@ -9,6 +9,7 @@
 from ros_buildfarm.argument import \
     add_argument_distribution_repository_key_files
 from ros_buildfarm.argument import add_argument_distribution_repository_urls
+from ros_buildfarm.argument import add_argument_custom_rosdep_urls
```
dirk-thomas: Please insert the import in alphabetical order following PEP 8.

jjekircp:
Thanks for the prompt feedback. I've pushed an update addressing all of @dirk-thomas's comments.

ros_buildfarm/config/source_build_file.py
```diff
@@ -92,6 +92,11 @@ def __init__(self, name, data):
             self.skip_ignored_repositories = \
                 bool(data['skip_ignored_repositories'])
 
+        if '_config' in data['targets']:
+            if 'custom_rosdep_urls' in data['targets']['_config']:
+                self.custom_rosdep_urls = data['targets']['_config']['custom_rosdep_urls']
```
dirk-thomas: Please always add the member variable to this class. E.g. with a default value of an empty list.

Then you also don't have to use `hasattr` above but can unconditionally access the member.

doc/configuration_options.rst
```diff
@@ -81,7 +81,8 @@ A set of options which can be used in any build file.
   The first level key contains the OS name.
   The second level key contains the OS code name.
   The third level key contains the CPU architecture.
-  The third level value is always empty.
+  The third level value is always empty. The first level key "_config" is
```
dirk-thomas: Please start each sentence on a new line (see https://github.com/ros2/ros2/wiki/Developer-Guide#markdown for the rational).

Same below.

jjekircp:
Updated again.
This wasn't called out specifically, but after taking a look at the Markdown Developer Guide linked above, I marked the yaml code block in the docs as yaml.
I added the custom_rosdep_urls variable to the BuildFile class, not the SourceBuildFile class. I hope this was the right choice; it ensures that the ``get_repositories_and_script_generating_key_files()`` function in common.py can access the member in all cases, including when processing a doc or release build file.

doc/configuration_options.rst
```diff
@@ -235,6 +237,13 @@ The following options are valid in version ``2`` (beside the generic options):
   * ``default``: a boolean flag as described for *test_commits*.
   * ``force``: a boolean flag as described for *test_commits*.
 
+The following options are valid as keys in the ``_config`` dict under
+``targets``:
+
+* ``custom_rosdep_urls``: a list of URLs containing rosdep sources.list.d entry
+ files that are downloaded into /etc/ros/rosdep/sources.list.d at the beginning
+  of the devel job.
+  If unset, the job simply runs _rosdep init_ instead.
```
dirk-thomas: The indentation is inconsistent in this bullet which breaks the rendering.

doc/custom_rosdistro.rst
```diff
@@ -124,6 +124,17 @@ rosdep database:
...
+To make devel jobs use your custom rosdep database, update the ``targets`` dict
+in the  \<distribution\>/source_build.yaml file in your forked
+ros_buildfarm_config repository with the ``custom_rosdep_urls`` option:
+
+```yaml
+targets:
+  _config:
+    custom_rosdep_urls:
+      - https://raw.githubusercontent.com/YOUR_FORK/rosdistro/master/rosdep/sources.list.d/20-default.list
+```
```
dirk-thomas: This block seems to not render as you might expect it to.

ros_buildfarm/config/build_file.py
```diff
@@ -37,6 +37,8 @@ class BuildFile(object):
     def __init__(self, name, data):
         self.name = name
 
+        self.custom_rosdep_urls = []
+
```
dirk-thomas: Please move this to the `SourceBuildFile` (above line 95) for now. Once any other build file supports additional configuration we can generalize it then.

jjekircp: Can do, but will then need to change ``get_repositories_and_script_generating_key_files()`` in common.py as some build_file objects passed into it will have ``custom_rosdep_urls`` defined and some will not (because that function gets called on release, source, and doc build files). Is ``hasattr`` the best way of ensuring I don't try to access an attribute that isn't defined or is there a better way that I should use instead?

dirk-thomas: Sounds good to me to use `hasattr` in this case.

doc/custom_rosdistro.rst
jjekircp: Think I have it fixed now. I didn't realize at first that the previewer I was using locally (based on pandoc) doesn't treat reStructuredText in exactly the same way as GitHub does.

dirk-thomas:
LGTM but I have not tried to run / use it. Waiting for @tfoote to get to test it as mentioned above (https://github.com/ros-infrastructure/ros_buildfarm/pull/100#issuecomment-141563839).

ros_buildfarm/config/source_build_file.py
```diff
@@ -92,6 +92,12 @@ def __init__(self, name, data):
             self.skip_ignored_repositories = \
                 bool(data['skip_ignored_repositories'])
 
+        self.custom_rosdep_urls = []
+        if '_config' in data['targets']:
+            if 'custom_rosdep_urls' in data['targets']['_config']:
+                self.custom_rosdep_urls = data['targets']['_config']['custom_rosdep_urls']
+                assert.isinstance(self.custom_rosdep_urls, list)
```
ddimarco: I think this should read `assert isinstance(...)`, i.e. without the dot.

ros_buildfarm/templates/snippet/rosdep_init.Dockerfile.em
```diff
@@ -0,0 +1,8 @@
+RUN mkdir -p /etc/ros/rosdep/sources.list.d
+
+@[for url in custom_rosdep_urls]@
+RUN (cd /etc/ros/rosdep/sources.list.d/ && wget --progress=dot @url )
+@[end for]@
+@[if not custom_rosdep_urls]@
+RUN rosdep init
```
ddimarco: I think we need `rosdep init` here regardless of whether `custom_rosdep_urls` is set.

ddimarco:
Apart from the comment above, it works for me.

ros_buildfarm/templates/snippet/rosdep_init.Dockerfile.em
jjekircp: What does `rosdep init` do other than create the /etc/ros/rosdep/sources.list.d/20-default file from the default one on ros/rosdistro? It didn't seem to me like it did anything else, which is why I did it that way.
It'd be easy enough to change it to run `rosdep init` first and then overwrite 20-default.list with the user-specified one if desired though.

tfoote: I would suggest running rosdep init either way. And not override the default sources, but add another file side by side with the additional sources. You should be able to set precidence using the number at the front of the file.

mistoll:
Tested this and it works for me.

ros_buildfarm/templates/snippet/rosdep_init.Dockerfile.em
jjekircp: Ok, I'll update that section to unconditionally run rosdep init.
I would prefer to keep the ability to override the default sources. Specifically I would like to fetch all of my rosdep sources from my forked rosdistro repo (whether or not they are identical to the public one). The primary motivation for this is that my rosdistro repo is hosted on an internal git server rather than GitHub, and so it's insulated from issues with GitHub or Internet connectivity at the site.

I can update the docs so it's clear to users that they can either use a combination of the default sources and their own custom sources file, or avoid the default sources entirely and link to the file in their forked rosdistro repo.

tfoote:
@ddimarco @mistoll thanks for testing

PR #104, opened 2015-09-25
Add support for git clone over SSH (fixes #91)

### LINKED ISSUE #91: Cannot use SSH URLs for source/release repositories in rosdistro repo
Hi all,

After noticing I couldn't use SSH URLs in my distribution.yaml, I started looking into what I'd have to change to get SSH key support inside the Docker container for the rosdistro cache job.

I referenced https://github.com/docker/docker/issues/6396 and I succeeded in forwarding the SSH agent into the container for the rosdistro cache job, but I ran into trouble with generate_all_jobs.py as I added the build-wrapper_ssh-agent snippet into templates/misc/rosdistro_cache_job.xml.em and it needs a credential ID.

That is, in scripts/misc/generate_rosdistro_cache_job.py lines 37-60 in get_job_config, I need to populate the job_data dictionary with all the values needed to expand the template, and I'm not sure where to get the credential ID from.

The release build configuration in ros_buildfarm_config has an "upload_credential_id", but in the general case that may not be the credential the user wishes to use to fetch private git repositories. Perhaps there is a need for another credential_id in the index.yaml in that repository?

I'd appreciate your feedback on how that should be implemented.

For reference, the work I've done so far is on the [ssh_git_clone](https://github.com/clearpathrobotics/ros_buildfarm/tree/ssh_git_clone) branch of my fork.

jjekircp:
At a high level, I have made the following changes to support cloning git repositories via SSH in rosdistro cache jobs, devel jobs, doc jobs, and release jobs:

* I have added a parameter git_ssh_credential_id in the index.yaml of ros_buildfarm_config. This is the ID of a credential that has already been set on the master that should be used for git+ssh cloning. This parameter is optional.
* When this parameter is defined, it is inserted into the following places:
	* Jenkins SCM plugin configuration for devel and doc jobs
	* SSH agent configuration for rosdistro cache jobs and sourcedeb jobs
* I have edited the docker run parameters for sourcedeb and rosdistro cache jobs to forward the SSH agent into the container. This allows the SSH agent configuration set in the job to be used in a git clone operation running inside the container.

One notable limitation in the code in this PR is that it is assumed that one SSH key will provide access to all repositories accessed by the buildfarm. 

This PR depends on the changes in buildfarm_deployment [#72](https://github.com/ros-infrastructure/buildfarm_deployment/pull/72) which provide the following:

* A credential on the Jenkins master containing a private SSH key
* Entries in the known_hosts file on the slave corresponding to any servers hosting repositories that will be cloned over SSH. If these entries are not present, jobs will fail with a "Host key verification failed" error.

### DIFF
--- a/ros_buildfarm/templates/misc/rosdistro_cache_job.xml.em
+++ b/ros_buildfarm/templates/misc/rosdistro_cache_job.xml.em
@@ -79,6 +79,9 @@
         ' --cidfile=$WORKSPACE/docker_generate_rosdistro_cache/docker.cid' +
         ' --net=host' +
         ' -v $WORKSPACE/rosdistro_cache:/tmp/rosdistro_cache' +
+        ' -v $HOME/.ssh/known_hosts:/etc/ssh/ssh_known_hosts:ro' +
+        ' -v $SSH_AUTH_SOCK:/tmp/ssh_auth_sock' +
+        ' -e SSH_AUTH_SOCK=/tmp/ssh_auth_sock' +
         ' rosdistro_cache_generation',
         'echo "# END SECTION"',
     ]),
@@ -103,5 +106,11 @@
 @(SNIPPET(
     'build-wrapper_timestamper',
 ))@
+@[if git_ssh_credential_id]@
+@(SNIPPET(
+    'build-wrapper_ssh-agent',
+    credential_id=git_ssh_credential_id,
+))@
+@[end if]@
   </buildWrappers>
 </project>
--- a/ros_buildfarm/templates/misc/rosdistro_cache_job.xml.em
+++ b/ros_buildfarm/templates/misc/rosdistro_cache_job.xml.em
@@ -79,9 +79,9 @@
         ' --cidfile=$WORKSPACE/docker_generate_rosdistro_cache/docker.cid' +
         ' --net=host' +
         ' -v $WORKSPACE/rosdistro_cache:/tmp/rosdistro_cache' +
-        ' -v $HOME/.ssh/known_hosts:/etc/ssh/ssh_known_hosts:ro' +
-        ' -v $SSH_AUTH_SOCK:/tmp/ssh_auth_sock' +
-        ' -e SSH_AUTH_SOCK=/tmp/ssh_auth_sock' +
+        (' -v $HOME/.ssh/known_hosts:/etc/ssh/ssh_known_hosts:ro'
+         ' -v $SSH_AUTH_SOCK:/tmp/ssh_auth_sock'
+         ' -e SSH_AUTH_SOCK=/tmp/ssh_auth_sock' if git_ssh_credential_id else '') +
         ' rosdistro_cache_generation',
         'echo "# END SECTION"',
     ]),
--- a/ros_buildfarm/templates/release/sourcedeb_job.xml.em
+++ b/ros_buildfarm/templates/release/sourcedeb_job.xml.em
@@ -114,9 +114,9 @@
         ' --net=host' +
         ' -v $WORKSPACE/ros_buildfarm:/tmp/ros_buildfarm:ro' +
         ' -v $WORKSPACE/sourcedeb:/tmp/sourcedeb' +
-        ' -v $HOME/.ssh/known_hosts:/etc/ssh/ssh_known_hosts:ro' +
-        ' -v $SSH_AUTH_SOCK:/tmp/ssh_auth_sock' +
-        ' -e SSH_AUTH_SOCK=/tmp/ssh_auth_sock' +
+        (' -v $HOME/.ssh/known_hosts:/etc/ssh/ssh_known_hosts:ro' +
+         ' -v $SSH_AUTH_SOCK:/tmp/ssh_auth_sock' +
+         ' -e SSH_AUTH_SOCK=/tmp/ssh_auth_sock' if git_ssh_credential_id else '') +
         ' sourcedeb.%s_%s_%s_%s' % (rosdistro_name, os_name, os_code_name, pkg_name),
         'echo "# END SECTION"',
     ]),
--- a/ros_buildfarm/templates/release/sourcedeb_job.xml.em
+++ b/ros_buildfarm/templates/release/sourcedeb_job.xml.em
@@ -118,7 +118,6 @@
         ' -v $SSH_AUTH_SOCK:/tmp/ssh_auth_sock' +
         ' -e SSH_AUTH_SOCK=/tmp/ssh_auth_sock' +
         ' sourcedeb.%s_%s_%s_%s' % (rosdistro_name, os_name, os_code_name, pkg_name),
-        ' sourcedeb.%s_%s_%s_%s' % (rosdistro_name, os_name, os_code_name, pkg_name),
         'echo "# END SECTION"',
     ]),
 ))@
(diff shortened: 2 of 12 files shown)


ros_buildfarm/templates/release/sourcedeb_job.xml.em
```diff
@@ -114,6 +114,10 @@
         ' --net=host' +
         ' -v $WORKSPACE/ros_buildfarm:/tmp/ros_buildfarm:ro' +
         ' -v $WORKSPACE/sourcedeb:/tmp/sourcedeb' +
+        ' -v $HOME/.ssh/known_hosts:/etc/ssh/ssh_known_hosts:ro' +
+        ' -v $SSH_AUTH_SOCK:/tmp/ssh_auth_sock' +
+        ' -e SSH_AUTH_SOCK=/tmp/ssh_auth_sock' +
+        ' sourcedeb.%s_%s_%s_%s' % (rosdistro_name, os_name, os_code_name, pkg_name),
```
dirk-thomas: Why is this line duplicated?

ros_buildfarm/templates/snippet/scm_git.xml.em
```diff
@@ -6,6 +6,9 @@
         <refspec>@ESCAPE(refspec)</refspec>
 @[end if]@
         <url>@ESCAPE(url)</url>
+@[if vars().get('git_ssh_credential_id')]@
```
dirk-thomas: Why is the condition so complicated? Shouldn't the following be enough:

    if git_ssh_credential_id

ros_buildfarm/templates/misc/rosdistro_cache_job.xml.em
```diff
@@ -79,6 +79,9 @@
         ' --cidfile=$WORKSPACE/docker_generate_rosdistro_cache/docker.cid' +
         ' --net=host' +
         ' -v $WORKSPACE/rosdistro_cache:/tmp/rosdistro_cache' +
+        ' -v $HOME/.ssh/known_hosts:/etc/ssh/ssh_known_hosts:ro' +
+        ' -v $SSH_AUTH_SOCK:/tmp/ssh_auth_sock' +
+        ' -e SSH_AUTH_SOCK=/tmp/ssh_auth_sock' +
```
dirk-thomas: Should these only be passed if necessary meaning if a git_ssh_credential_id is set?

Same below.

ros_buildfarm/templates/release/sourcedeb_job.xml.em
jjekircp: Are you referring to 118 and 119? Those are two different operations (one is a volume and one is an environment variable) though the lines look very similar.
Or have I duplicated something else?

ros_buildfarm/templates/snippet/scm_git.xml.em
jjekircp: Not all uses of this snippet set a git_ssh_credential_id, so I wanted to avoid an error in the template expansion due to an unset variable.

ros_buildfarm/templates/release/sourcedeb_job.xml.em
dirk-thomas: I am referring to line 120 and 121.

ros_buildfarm/templates/snippet/scm_git.xml.em
dirk-thomas: Ok, that makes sense. Currently all invocations always set all arguments. But I see the benefit of this instead of updating all calls with an empty argument.

ros_buildfarm/templates/misc/rosdistro_cache_job.xml.em
jjekircp: I didn't bother making those conditional as there seems to be no harm in them being there if they're not used, but I can change that if it is preferred.

mistoll: I've tested the pull and theses lines break the job, when $SSH_AUTH_SOCK is not set. So making them conditional would increase compatibility.
I didn't figure out why $SSH_AUTH_SOCK is not set though.

jjekircp: SSH_AUTH_SOCK is only ever unset if the SSH Agent plugin isn't installed. I made commits to the buildfarm_deployment repo to install that plugin, but then I discovered when I tried to deploy a new buildfarm that the plugin wasn't getting installed. It turned out that I had to explicitly install all the dependencies for the plugin as well as the plugin itself, so I updated the repo with the following commit: https://github.com/clearpathrobotics/buildfarm_deployment/commit/80aefe38887874b2fd67f8e8647d677beca84d7a

Please try checking out that version of buildfarm_deployment (at the top of reconfigure.bash in buildfarm_deployment_config you can specify the URL and branch/version to use) with that commit, reconfiguring your master, and trying again.

You've raised a good point though, in that anyone who deploys a buildfarm with buildfarm_deployment before my changes and then updates ros_buildfarm with the changes from this PR will have broken jobs until they either re-run reconfigure.bash or otherwise cause the SSH Agent plugin to be installed.

mistoll: Actually the job was not configured correctly (SSH Agent installed but not enabled). Rerunning generate_all_jobs.py on a clean checkout fixed the problem.

mistoll:
Seems to work.
Tested:
- http git
- ssh git with credentials
- dev jobs
- release jobs

ros_buildfarm/templates/misc/rosdistro_cache_job.xml.em
dirk-thomas: If this is only necessary when `git_ssh_credential_id` is set I think it should also be only passed conditionally. Because it keeps the simple case simple and implicitly documents how these arguments are being used.

jjekircp: Sorry about the delay. I see the rationale for why those arguments should only be passed conditionally and I'm happy to make that change.
Does this look like a reasonable way to conditionally add or not add these arguments to the docker run invocation?
```
        ' -v $HOME/.ssh/known_hosts:/etc/ssh/ssh_known_hosts:ro' +
        '{}'.format(' -v $SSH_AUTH_SOCK:/tmp/ssh_auth_sock -e SSH_AUTH_SOCK=/tmp/ssh_auth_sock' if git_ssh_credential_id else '') +
        ' rosdistro_cache_generation',
```

dirk-thomas: The `format` call seems unnecessary. For readability I would suggest keeping each argument on a separate line:

    (' -v $HOME/.ssh/known_hosts:/etc/ssh/ssh_known_hosts:ro'
     ' -v $SSH_AUTH_SOCK:/tmp/ssh_auth_sock' if git_ssh_credential_id else '') +



Is `known_hosts` also only necessary conditionally?

tfoote:
Thanks for the contributions! Sorry we took so long to review.

ros_buildfarm/templates/release/sourcedeb_job.xml.em
```diff
@@ -174,7 +177,7 @@
 ))@
 @(SNIPPET(
     'build-wrapper_ssh-agent',
-    credential_id=credential_id,
+    credential_id=git_ssh_credential_id,
```
dirk-thomas: Removing the `credential_id` here breaks the sourcedeb jobs. Addressed in 7112d447c0d076e076fc78fc6237f734240031a4.

PR #108, opened 2015-10-17
move utopic to old-releases

tfoote:
Parallel change to https://github.com/ros-infrastructure/buildfarm/pull/218


tfoote:
Merging to fix the builds. @dirk-thomas please comment if there's an issue.

dirk-thomas:
+1

PR #112, opened 2015-10-28
homogenize the reconfiguration process to use the file based configs

tfoote:
for all reconfigures

Tested here: http://54.183.26.131:8080/view/Manage/job/Idev_reconfigure-jobs/234/console

### DIFF
--- a/ros_buildfarm/common.py
+++ b/ros_buildfarm/common.py
@@ -353,3 +353,20 @@ def get_release_job_urls(
 
 def _get_job_url(jenkins_url, view_name, job_name):
     return '%s/view/%s/job/%s' % (jenkins_url, view_name, job_name)
+
+
+def write_groovy_script_and_configs(filename,
+                                    content,
+                                    job_configs):
+    """Write out the groovy script for reconfiguring as well as the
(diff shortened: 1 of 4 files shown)


tfoote:
@dirk-thomas comment and I'll fix it retroactively

ros_buildfarm/common.py
```diff
@@ -353,3 +353,20 @@ def get_release_job_urls(
 
 def _get_job_url(jenkins_url, view_name, job_name):
     return '%s/view/%s/job/%s' % (jenkins_url, view_name, job_name)
+
+
+def write_groovy_script_and_configs(filename,
+                                    content,
```
dirk-thomas: Please don't align wrapped lines at the open parenthesis.

ros_buildfarm/common.py
```diff
@@ -353,3 +353,20 @@ def get_release_job_urls(
...
+                                    job_configs):
+    """Write out the groovy script for reconfiguring as well as the
+     config files in a subdirectory 'configs'
```
dirk-thomas: Please follow PEP 257.

PR #126, opened 2016-01-11
update plugin version to current state

dirk-thomas:
To avoid diffs on reconfigure when jobs are being touched through the UI: http://build.ros.org:8080/job/Irel_reconfigure-jobs/5/consoleFull#console-section-5


tfoote:
+1

PR #129, opened 2016-01-12
add detailed message to assert

dirk-thomas:
While it doesn't prevent the problem from happening it provides a much better message to the user.


esteve:
+1

PR #132, opened 2016-01-12
add code style checks using travis

dirk-thomas:


tfoote:
that would be great. +1 assuming they'll pass

dirk-thomas:
Yes, as of #133.

PR #137, opened 2016-01-14
add doc for changing the default view

### LINKED ISSUE #125: make default view not All
We don't want the full list of jobs in the default page load. 

We used to use the "Manage" view, but we'd prefer to see the full build queue not the filtered queue for the "Manage" view.

dirk-thomas:
Fixes #125.


tfoote:
+1

esteve:
+1

PR #138, opened 2016-01-14
change sourcedeb job to use rosdistro cache

### LINKED ISSUE #41: GitHub CDN returns cached data
This can result in a source job rebuilding an older package version since the fetched distribution file from the rosdistro still contains the old version - even if the job was triggered because a newer version has just been released. This "fixes" itself 15 minutes later when the source job is automatically invoked again.

It can also result in the binary job to fail since it can not download the latest sourcedeb.

dirk-thomas:
Fixes #41.

This sourcedeb job continues to work with the patch
http://build.ros.org/view/Isrc_uT/job/Isrc_uT__care_o_bot_desktop__ubuntu_trusty__source/8/console#console-section-4

Since we can't easily reproduce the problem since it is based dependent on the response of the GitHub CDN that's all I could check for now.


tfoote:
+1

PR #140, opened 2016-01-15
retry import of sourcedeb once without sending notifications of failed attempt

dirk-thomas:
For each package N sourcedeb job are triggered - one for each target platform. Each job tries to import its artifact into the apt repo. All but the last one fail since other platforms are still depending on the `orig.tar.gz`.

This would not happen in the following ticket would be addressed: ros-infrastructure/bloom#150

While the failing sourcedeb jobs are automatically retriggered the notification emails for the failing runs are undesired. This is addressed by marking the sourcedeb job as `aborted` if the triggered import job failed and directly rescheduling another build (but only once to prevent infinite recursion).

This fixes ros-infrastructure/reprepro-updater#16

The debinc 1 shows the problem, while debinc 3 shows the working solution:
* http://54.183.26.131:8080/view/Isrc_uS/job/Isrc_uS__ax2550__ubuntu_saucy__source/
* http://54.183.26.131:8080/view/Isrc_uT/job/Isrc_uT__ax2550__ubuntu_trusty__source/
* http://54.183.26.131:8080/view/Manage/job/Irel_import-package-deb-inc/

### DIFF
--- a/ros_buildfarm/templates/release/binarydeb_job.xml.em
+++ b/ros_buildfarm/templates/release/binarydeb_job.xml.em
@@ -219,6 +219,7 @@
     parameters='\n'.join([
         'subfolder=%s/${JOB_NAME}__${BUILD_NUMBER}' % os_code_name,
         'debian_package_name=%s' % debian_package_name]),
+    continue_on_failure=False,
 ))@
   </builders>
   <publishers>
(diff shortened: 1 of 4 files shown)


tfoote:
+1

PR #142, opened 2016-01-15
close file handles

dirk-thomas:
Can the branch `close_files` be deleted?


tfoote:
+1

this captures the important part of that branch, deleted

PR #144, opened 2016-01-16
Various

dirk-thomas:
The first commit will get rid of the diff when configuring views: http://build.ros.org/job/Jdev_reconfigure-jobs/9/consoleFull#console-section-5

The second commit suppressed the `fatal: ...` error message in http://build.ros.org/job/Jdev_reconfigure-jobs/9/console

The third commit fixes the problem that the reconfigure job embeds the current hash into every job (http://build.ros.org/job/Jrel_reconfigure-jobs/8/console). Instead it should use the branch name which will prevent long reconfiguration times every night.


tfoote:
lets give it a try +1

PR #147, opened 2016-01-19
remove option which skips known host verification

tfoote:
This is a security fix and connects to https://github.com/ros-infrastructure/buildfarm_deployment_config/pull/12


dirk-thomas:
Lgtm

PR #151, opened 2016-01-21
ensure that folder can be removed

dirk-thomas:
This was necessary to build http://build.ros.org/view/Ibin_uT64/job/Ibin_uT64__sr_external_dependencies__ubuntu_trusty_amd64__binary/ successfully.

While the Jenkins user is still the owner he doesn't have write permissions for the directories (due to what the ROS package does internally when building the deb) which makes the `rm -fr` command fail.


tfoote:
+1

PR #152, opened 2016-01-21
fix issue with encoding of html content

dirk-thomas:
Perform the injection of the canonical url without encoding / decoding.

Fixes the unicode issue in http://build.ros.org/view/Jdoc/job/Jdoc__log4cpp__ubuntu_trusty_amd64/


dirk-thomas:
@tfoote I will go ahead with the merge to make use of the fix in the current wave of doc jobs. I can apply any feedback after the fact.

tfoote:
lgtm

PR #153, opened 2016-01-21
catch invalid rosdoc yaml

dirk-thomas:
This will let this doc job pass (as it did before): http://build.ros.org/view/Idoc/job/Idoc__asctec_mav_framework__ubuntu_trusty_amd64/4/console

For this new as well as other existing `WARNINGS` the job should actually be marked marked unstable...


dirk-thomas:
Invalid upstream rosdoc.yaml file fixed in https://github.com/ethz-asl/asctec_mav_framework#60

tfoote:
+1

PR #155, opened 2016-01-22
fix missing credential in source jobs

### LINKED ISSUE #154: sourcedeb configs are missing credentials causing failures.
We have run into an issue of sourcedeb jobs not running due to missing credentials. 

I looked at the config and it all looked "right" however when I saved it from the GUI it passed. So I diffed it to a neighboring job and we've got a lot of differences. 
Here's the diff of two similar source jobs:
https://gist.github.com/tfoote/db2d9965466c0b66d6b2

It looks like somethings changed with the config escaping mechanism.  

But somewhere the credential_id has been lost. 
Declared here: [default](https://github.com/ros-infrastructure/ros_buildfarm_config/blob/production/indigo/release-build.yaml#L61) and [armhf](https://github.com/ros-infrastructure/ros_buildfarm_config/blob/production/indigo/release-armhf-build.yaml#L116)

Loaded here: https://github.com/ros-infrastructure/ros_buildfarm/blob/a72d54a1d5206c0ee51149d8128034ef4272a7c9/ros_buildfarm/release_job.py#L549
Passed here: https://github.com/ros-infrastructure/ros_buildfarm/blob/a72d54a1d5206c0ee51149d8128034ef4272a7c9/ros_buildfarm/templates/release/sourcedeb_job.xml.em#L176
And implemented in this snippet: https://github.com/ros-infrastructure/ros_buildfarm/blob/3221d53fc8c635e63a655c22bb9e2b6ebc5af322/ros_buildfarm/templates/snippet/build-wrapper_ssh-agent.xml.em#L3

Which oddly does not have the possibility to have the empty `<string/>` which I see in all the souredeb configs. 

This reconfigure stripped the credentials: http://build.ros.org/job/Irel_arm_reconfigure-jobs/9/console

Running latest version: a5901cd9b5a1204a11ef03de4c3095ed0466cf64
The previous [run](http://build.ros.org/job/Irel_arm_reconfigure-jobs/8/consoleFull) ran: 7f16ef4898764079e500528c3b7b5a942f6cb69f

[git diff](https://github.com/ros-infrastructure/ros_buildfarm/compare/7f16ef4...a5901cd) between the above commits.

dirk-thomas:
plus some other minor stuff...

Fixes #154.

### DIFF
--- a/ros_buildfarm/templates/doc/doc_independent_job.xml.em
+++ b/ros_buildfarm/templates/doc/doc_independent_job.xml.em
@@ -151,7 +151,7 @@ repo_name = os.path.splitext(os.path.basename(repo_url))[0]
 ))@
 @(SNIPPET(
     'build-wrapper_ssh-agent',
-    credential_id=credential_id,
+    credential_ids=[credential_id],
 ))@
   </buildWrappers>
 </project>
--- a/ros_buildfarm/templates/doc/doc_job.xml.em
+++ b/ros_buildfarm/templates/doc/doc_job.xml.em
@@ -279,7 +279,7 @@
 ))@
 @(SNIPPET(
     'build-wrapper_ssh-agent',
-    credential_id=credential_id,
+    credential_ids=[credential_id],
 ))@
   </buildWrappers>
 </project>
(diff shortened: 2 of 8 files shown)


tfoote:
lgtm

PR #156, opened 2016-01-22
fix generated doc url

dirk-thomas:
Fixes ros-infrastructure/roswiki#145.

As far as I can see from the code in `jenkins_scripts` the `html` subfolder was always appended before.


dirk-thomas:
I retriggered one doc job with this branch (http://build.ros.org/view/Jdoc/job/Jdoc__robot_state_publisher__ubuntu_trusty_amd64/5/) and the result looks good: http://wiki.ros.org/robot_state_publisher

tfoote:
Lgtm

PR #158, opened 2016-01-22
add known error string to apt-get wrapper

dirk-thomas:
Should catch http://build.ros.org/view/All/job/Idev__ridgeback__ubuntu_trusty_amd64/3/console next time.

### DIFF
--- a/scripts/wrapper/apt-get.py
+++ b/scripts/wrapper/apt-get.py
@@ -7,7 +7,11 @@ from time import sleep
 
 def main(argv=sys.argv[1:]):
     max_tries = 10
-    known_error_strings = ['Failed to fetch', 'Hash Sum mismatch']
+    known_error_strings = [
+        'Failed to fetch',
+        'Hash Sum mismatch',
+        'is not what the server reported',
+    ]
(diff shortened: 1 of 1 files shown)


scripts/wrapper/apt-get.py
```diff
@@ -7,7 +7,11 @@
 
 def main(argv=sys.argv[1:]):
     max_tries = 10
-    known_error_strings = ['Failed to fetch', 'Hash Sum mismatch']
+    known_error_strings = [
+        'Failed to fetch',
```
esteve: Does this run inside a docker with `LC_ALL=C` / `LANG=C` ? If so, that's fine, but if there's a different system locale, this may fail to catch these strings.

tfoote: We use 
`RUN locale-gen en_US.UTF-8`
and
`ENV LANG en_US.UTF-8`

LC_ALL=C might be a better solution. https://unix.stackexchange.com/questions/87745/what-does-lc-all-c-do#

tfoote:
+1

PR #159, opened 2016-01-22
change trigger jobs to use distro cache

dirk-thomas:
To match behavior of source jobs: #138


tfoote:
+1 I saw a few of those overnight.

PR #161, opened 2016-01-26
use job specific ghprb commit status context

dirk-thomas:
Without this multiple jobs overwrite the same `default` status.

An example PR using this is `catkin` # 776.


tfoote:
Why the changes in lines 7, 11, 20, and 21?

dirk-thomas:
Because that is how the configuration looks like after saving the job in the web ui.

tfoote:
+1

PR #164, opened 2016-01-26
use different empy interpreter API

dirk-thomas:
The change results in a 4x - 5x speed improvement to evaluate templates:
* Before http://build.ros.org/job/Jrel_arm_reconfigure-jobs/23/ (22 min.)
* After http://build.ros.org/job/Jrel_arm_reconfigure-jobs/24/ (7 min.)


tfoote:
lgtm

I wonder what they're doing inside that's so much slower.

dirk-thomas:
Incrementally reading the file and feeding the chunks to the scanner / parser.

PR #165, opened 2016-01-26
cache tokens for each template

dirk-thomas:
Avoids the cost for repeatedly parsing each template:
* Before http://build.ros.org/job/Jrel_arm_reconfigure-jobs/27/ (7m15s)
* After http://build.ros.org/job/Jrel_arm_reconfigure-jobs/29/ (2m45s)

For reconfigure jobs with more jobs the performance improvement will be even more significant.


tfoote:
+1

PR #168, opened 2016-01-27
catch socket errors

tfoote:
This hit us a bunch last night such as: http://54.183.26.131:8080/job/Irel_release-status-page/27614/

The traceback does not show the url which failed to load.

### DIFF
--- a/ros_buildfarm/config/loader.py
+++ b/ros_buildfarm/config/loader.py
@@ -61,6 +61,8 @@ def load_url(url, retry=2, retry_period=1, timeout=10, skip_decode=False):
                 url, retry=retry - 1, retry_period=retry_period,
                 timeout=timeout)
         raise URLError(str(e) + ' (%s)' % url)
+    except socket.timeout as e:
+        raise socket.timeout(str(e) + ' (%s)' % url)
     # Python 2/3 Compatibility
     contents = fh.read()
     if isinstance(contents, str) or skip_decode:


dirk-thomas:
The build log shows `socket.timeout: The read operation timed out` but this patch catches `socket.error`?

tfoote:
socket.error is the base class for all socket exceptions, timeout, gaierror, or herror: https://docs.python.org/2/library/socket.html

dirk-thomas:
I don't think they inherit from each other: https://docs.python.org/3/library/socket.html#socket.timeout

ros_buildfarm/config/loader.py
```diff
@@ -61,6 +61,8 @@ def load_url(url, retry=2, retry_period=1, timeout=10, skip_decode=False):
                 url, retry=retry - 1, retry_period=retry_period,
                 timeout=timeout)
         raise URLError(str(e) + ' (%s)' % url)
+    except OSError as e:
+        raise OSError(str(e) + ' (%s)' % url)
```
dirk-thomas: If a subclass is raised this changes the exception type. Try something like in line 55 instead.

ros_buildfarm/config/loader.py
```diff
@@ -61,6 +61,9 @@ def load_url(url, retry=2, retry_period=1, timeout=10, skip_decode=False):
                 url, retry=retry - 1, retry_period=retry_period,
                 timeout=timeout)
         raise URLError(str(e) + ' (%s)' % url)
+    except OSError as e:
+        e.strerror += ' (%s)' % url
```
dirk-thomas: > TypeError: unsupported operand type(s) for +=: 'NoneType' and 'str'

dirk-thomas:
What about the following:

```
except socket.timeout as e:
    raise socket.timeout(str(e) + ' (%s)` % url)
```

tfoote:
Yeah, I thought it would be better to be more generic, but catching the one case I guess is best.

dirk-thomas:
+1

dirk-thomas:
The newly catched exception doesn't implement the requested retry. See #171.

Issue #114, opened 2015-11-03
binarydeb job pulls wrong sourcedeb from repo if more than one is available


jjekircp:
If the release build file includes more than one apt repository, and there is more than one sourcedeb available for a given binarydeb job, the binarydeb job will blindly take the highest version and not the one in the rosdistro.

I'm not sure if I'm running a supported configuration or not, but in my `distribution.yaml` I have my private packages alone, while retrieving all public packages directly from packages.ros.org. I have a package where I'm trying to build an older version than is currently in the indigo distribution, and its binarydeb job fails because the apt-get source command pulls the latest version available, which ends up not matching the job configuration.

I've worked around it with the following change, which obtains the version number from the rosdistro, runs apt-cache to determine the Debian package version number which matches, and then calls apt-get source with an exact version number specification.

Please take a look at the following commit and let me know if I'm misusing the custom rosdistro functionality or if I have something that I should submit for review.

https://github.com/clearpathrobotics/ros_buildfarm/commit/6e974bba01b1be39bd8850788cb8ac11a528d415

Quick detailed summary of changes:

* In scripts/release/get_sourcedeb.py I pass the rosdistro index URL into the get_sourcedeb function.
* In ros_buildfarm/templates/release/binarydeb_create_task.Dockerfile.em I pass the rosdistro index URL into get_sourcedeb.py.
* In binarydeb_job.py get_sourcedeb() I get the rosdistro index and distribution file and pull out the version corresponding to the package_name passed into the function, then call `apt-get showsrc` to find the corresponding Debian version, then change the `apt-get source` call to `package=version` instead of just `package`.

tfoote:
In general this is something that debian packaging does not support/recommend rolling old versions of packages. Typically when you want to build an older version for distribution you do something like FAKE-NEW-VERSION_OLD-VERSION-NUMBER which will sort correctly.  This might be something reasonable but I'll have to do some whiteboarding to review.

dirk-thomas:
I have implemented a modified version in #169. Can you please try if it works for your use case and comment here.

jjekircp:
@dirk-thomas: Reading the diff of #169 it looks like the exact same effect as the change I made, so I can't see it not working. 
Apologies, there's a caveat to this change I've noticed after running a buildfarm with it for a while, so I'll describe it here in case you haven't run across it. Depending on which packages people add to the buildfarm and when, you can get job failures under the following circumstances:
* User bumps package X in the rosdistro
[...]

PR #169, opened 2016-01-27
use exact sourcedeb version when building binarydeb

dirk-thomas:
I modified the patch mentioned in #114 to use the distro cache rather then the distro file directly. I also added some asserts just in case something is not as expected.

See http://build.ros.org/job/Ibin_uT64__ax2550__ubuntu_trusty_amd64__binary/10/consoleFull#console-section-10

Connect to #114.


tfoote:
+1

PR #172, opened 2016-01-27
add dry run option

dirk-thomas:
For the top level script `generate_all_jobs` I used a `--commit` option so that the default invocation is to not modify anything.


tfoote:
+1

PR #174, opened 2016-01-27
fix source jobs across build files

dirk-thomas:
In order to test this I need to be able to *dry-run* the reconfigure jobs.


dirk-thomas:
Before the arm reconfigure removes the maintainer notification from the source jobs: http://build.ros.org/job/Irel_arm_reconfigure-jobs/27/console

After the arm reconfigure keeps the maintainer notification from the source jobs: http://build.ros.org/job/Irel_arm_reconfigure-jobs/26/consoleFull

PR #175, opened 2016-01-27
add dry run option to reconfigure jobs

dirk-thomas:
Currently this only implements the option for release jobs.

### DIFF
--- a/ros_buildfarm/templates/release/release_reconfigure-jobs_job.xml.em
+++ b/ros_buildfarm/templates/release/release_reconfigure-jobs_job.xml.em
@@ -13,6 +13,16 @@
     priority=30,
 ))@
 @(SNIPPET(
+    'property_parameters-definition',
+    parameters=[
+        {
+            'type': 'boolean',
+            'name': 'dry_run',
+            'description': 'Skip the actual reconfiguration but show the diffs',
+        },
+    ],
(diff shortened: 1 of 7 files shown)


ros_buildfarm/templates/release/release_reconfigure-jobs_job.xml.em
```diff
@@ -13,6 +13,16 @@
     priority=30,
 ))@
 @(SNIPPET(
+    'property_parameters-definition',
+    parameters=[
+        {
+            'type': 'boolean',
+            'name': 'dry_run',
```
tfoote: since this will become an environment variable in the test it would be cleaner to be all caps

tfoote:
one comment, otherwise lgtm

ros_buildfarm/templates/release/release_reconfigure-jobs_job.xml.em
dirk-thomas: Almost all current parameters are lower case. Therefore I will stick to it in this PR. If we think its worth we can update the existing code in a separate PR though.

dirk-thomas:
The dry run seems to have worked as desired: http://build.ros.org/job/Irel_reconfigure-jobs/27/console

PR #176, opened 2016-01-28
add dry run option to reconfigure dev and doc jobs

dirk-thomas:


dirk-thomas:
Worked for:
* http://build.ros.org/job/Jdoc_reconfigure-jobs/28/
* http://build.ros.org/job/Jdev_reconfigure-jobs/26/

PR #182, opened 2016-01-29
support groovy to reconfigure all release parts

dirk-thomas:


ros_buildfarm/jenkins.py
```diff
@@ -100,6 +100,8 @@ def configure_view(
         template_name='generic_view.xml.em', dry_run=False):
     view_config = get_view_config(
         template_name, view_name, include_regex=include_regex)
+    if not jenkins:
```
tfoote: Why do you return the config object here?

dirk-thomas: Because it is needed outside to generate the groovy script.

tfoote: Ok, it's surprising to have it return either a `view_config`, `None`, or a `view`

tfoote:
+1

dirk-thomas:
Regression with non-release reconfigure jobs fixed in b491c74ca03fb0b4949174b0c19960fbd6ec721d.

PR #188, opened 2016-01-30
improve rsync command in doc jobs

dirk-thomas:
This avoids extensive crawling while copying the same files and not failing if the remove location doesn't exist. Instead of 10+min. and a high server load this finishes in seconds.

Thank you @gerkey for providing the "better" rsync invocation.

See http://build.ros.org/job/Jdoc__genmsg__ubuntu_trusty_amd64/16/console#console-section-7


tfoote:
+1

PR #193, opened 2016-02-02
store packages of metapackage in manifest.yaml

dirk-thomas:
Without the navigation header of a metapackage (e.g. http://wiki.ros.org/roscpp?distro=indigo) doesn't list the packages within the metapackage.

@gavanderhoorn Thank you for reporting this.


tfoote:
+1

PR #195, opened 2016-02-02
add subsections in devel builds

dirk-thomas:
http://build.ros.org/job/Idev__ros_comm__ubuntu_trusty_amd64/4/consoleFull


tfoote:
+1

PR #196, opened 2016-02-04
source devel space and force CMake when running tests

dirk-thomas:
This is everything but nice but currently afaik the only way to provide a working environment to the tests with `catkin_make_isolated`:
* before: http://build.ros.org/job/Idev__ros_comm__ubuntu_trusty_amd64/4/
* after: http://build.ros.org/job/Idev__ros_comm__ubuntu_trusty_amd64/7/

### DIFF
--- a/scripts/devel/catkin_make_isolated_and_test.py
+++ b/scripts/devel/catkin_make_isolated_and_test.py
@@ -45,7 +45,8 @@ def main(argv=sys.argv[1:]):
 
     try:
         with Scope('SUBSECTION', 'build workspace in isolation'):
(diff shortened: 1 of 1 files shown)


wjwwood:
I don't know of a better way either.

tfoote:
Yeah. I don't know of a better way either.

dirk-thomas:
The remaining failing tests of `roswtf` are related to environment variables containing nonexistent paths. I created ros/catkin#777 to avoid those entries.

dirk-thomas:
And the changelog for this in af06e3b6f4fe344da69f89542ede89e9b4da0e3c.

PR #202, opened 2016-02-06
fix usage of git wrapper script

dirk-thomas:


dirk-thomas:
And of course needed to fix the hot fix: 826965e9da6cc870c2462c12b84c7363c4f60be3

PR #206, opened 2016-02-08
use the configured mercurial installation to enable caching

tfoote:
This is a matching change to https://github.com/ros-infrastructure/buildfarm_deployment/pull/117 which will enable mercurial caching for the scm polling. 

Otherwise it relies on workspace based polling, which triggers a job anytime the workspace is cleared. (Such as when the last slave executed upon is destroyed.)


dirk-thomas:
lgtm

PR #208, opened 2016-02-11
remove non-existing job urls in generated manifest.yaml files

dirk-thomas:
Fixes ros-infrastructure/roswiki#154.

### DIFF
--- a/scripts/doc/extract_metadata.py
+++ b/scripts/doc/extract_metadata.py
@@ -64,15 +64,17 @@ def main(argv=sys.argv[1:]):
             continue
 
         # add devel job urls
-        build_files = {}
-        for build_name in source_build_files.keys():
-            build_files[build_name] = source_build_files[build_name]
         rel_pkg = distribution.release_packages[pkg_name]
         repo_name = rel_pkg.repository_name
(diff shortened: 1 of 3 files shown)


scripts/doc/extract_metadata.py
```diff
@@ -64,15 +64,19 @@ def main(argv=sys.argv[1:]):
...
-        devel_job_urls = get_devel_job_urls(
-            config.jenkins_url, build_files, args.rosdistro_name, repo_name)
-        if devel_job_urls:
-            data['devel_jobs'] = devel_job_urls
+        repo = distribution.repositories[repo_name]
+        if repo.source_repository and repo.source_repository.version:
+            build_files = {}
+            for build_name in source_build_files.keys():
+                build_files[build_name] = source_build_files[build_name]
+            rel_pkg = distribution.release_packages[pkg_name]
```
tfoote: These two lines look redundant with 67 and 68

dirk-thomas: Good catch. I removed the two redundant line in the conditional block.

tfoote:
lgtm

PR #215, opened 2016-02-29
add note about updating sync threshold

dirk-thomas:
Based on discussion in ros-infrastructure/ros_buildfarm_config#26

### DIFF
--- a/doc/ongoing_operations.rst
+++ b/doc/ongoing_operations.rst
@@ -21,6 +21,10 @@ Manually sync packages
 The *source* and *binary* are imported into the ``building`` repository and if
 the sync criteria is fulfilled automatically synced to the ``testing``
 repository.
+The maintainer should revisit the configured sync threshold to adjust it when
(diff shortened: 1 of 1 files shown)


doc/ongoing_operations.rst
```diff
@@ -21,6 +21,8 @@ Manually sync packages
 The *source* and *binary* are imported into the ``building`` repository and if
 the sync criteria is fulfilled automatically synced to the ``testing``
 repository.
+The maintainer should revisit the configured sync threshold to adjust it when
+the number of released packages changes over the lifetime of a distribution.
```
tfoote: I'd add:

The threshold should be set at a level below which a sync should not happen into testing as there has been a major regression during the build.

tfoote:
One more sentence would be great. With that added +1

doc/ongoing_operations.rst
dirk-thomas: Please go ahead and add that sentence.

PR #217, opened 2016-03-02
extract compiler warnings for devel/pr jobs

dirk-thomas:
See http://build.ros.org/view/Ipr/ and http://build.ros.org/job/Ipr__ros_comm__ubuntu_trusty_amd64/65/warnings17Result/


tfoote:
I see it on the pr jobs, is it just not on the devel jobs pending reconfigure with this chage merged?

dirk-thomas:
I manually update the view and job and based on that created this PR. After this is merged and reconfigure ran all devel and PR views and jobs will have that feature.

tfoote:
+1

PR #226, opened 2016-03-04
use gbp instead of 'git buildpackage' command

dirk-thomas:


tfoote:
+1

esteve:
+1

tfoote:
https://github.com/agx/git-buildpackage/blob/master/debian/changelog#L103

PR #227, opened 2016-03-04
add subsection in doc jobs for better outline

dirk-thomas:
The code block is only indented under the new `with` subsection without modifications.


tfoote:
+1 

thanks for the clarification on the diff. it's not well displayed.

PR #228, opened 2016-03-05
fix catkin doc job

dirk-thomas:
While it doesn't make the job pass yet the only remaining task now is to release `genmsg`.

For `genmsg` I expect devel and release jobs to require the same fix to use the same `os_codename` to generate their Dockerfiles.


tfoote:
+1

dirk-thomas:
Fixed regression with 134c136622f2e1b0a4ab5f1914338ac2fde86ad7.

PR #236, opened 2016-03-09
maintain pull request data when reconfiguring job using groovy

### LINKED ISSUE #234: pull request jobs are being triggered too often
Even for a comment on the PR.

This is likely due to the fact that the GHPRB plugin is storing information in the job configuration. But the job is being reconfigured nightly.

dirk-thomas:
Fixes #234.

Before: http://build.ros.org/job/Jdev_reconfigure-jobs/73/console#console-section-9
After: http://build.ros.org/job/Jdev_reconfigure-jobs/79/console#console-section-3


tfoote:
+1

PR #239, opened 2016-03-10
reconfigure release jobs when the rosdistro cache gets an updated manifest

### LINKED ISSUE #235: reconfigure jobs after changes to rosdistro
Currently the reconfigure jobs are only triggered once per day. With their improved performance and the ability to only reconfigure a subset of repos / packages I think they should be triggered directly after updating the rosdistro cache.

dirk-thomas:
Fixes #235.

See the description of this build http://build.ros.org/job/kinetic_rosdistro-cache/1849/ which triggered http://build.ros.org/job/Krel_reconfigure-jobs/12/console#console-section-8

And http://build.ros.org/job/kinetic_rosdistro-cache/1864/ which triggered http://build.ros.org/job/Krel_reconfigure-jobs/13/console#console-section-8

The only delay is now due to the 5 min. interval in which the distro cache is rebuild and the latency the GitHub CDN introduces serving "old" versions of the content for a few minutes.


tfoote:
This looks good. One request is can we put an echo in where it gets triggered. I don't find a record of the triggering i the console output. And the way it's being triggered the downstream reconfigure only reports the SYSTEM triggered it. So it's only traceable if you guess based on timestamps.

dirk-thomas:
Sadly no, since it is a post-build step the script can't print to the console anymore.

tfoote:
Jenkins so flexible, except in the ways we want...

Oh well, lets merge this then.

PR #240, opened 2016-03-10
fix devel and doc reconfiguration if cache is behind

dirk-thomas:
The last nightly jobs (http://build.ros.org/job/Kdev_reconfigure-jobs/13/console#console-section-5 http://build.ros.org/job/Kdoc_reconfigure-jobs/11/console#console-section-5) failed because the rosdistro cache job for Kinetic was disabled and therefore the cache was out-of-date. While that should be uncommon the race condition still exists and is worked around by this patch.


tfoote:
+1

PR #242, opened 2016-03-11
maintain the job order when reconfiguring using Groovy

dirk-thomas:
Currently the Groovy script processes the job configuration in alphabetical order. But if e.g. a new binary job is created before its related source job the job dependency is not established. (The binary job contains the information and if it is saved again, after the source job exists, the connection is established correctly.)

### DIFF
--- a/ros_buildfarm/common.py
+++ b/ros_buildfarm/common.py
@@ -396,7 +396,13 @@ def write_groovy_script_and_configs(
     job_config_dir = os.path.join(os.path.dirname(filename), 'job_configs')
     if not os.path.isdir(job_config_dir):
         os.makedirs(job_config_dir)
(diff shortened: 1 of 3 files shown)


dirk-thomas:
The same problem exists if two binary jobs are not generated in topological order. That problem is also present when the jobs are generated using the remote API. This will be addressed by the second commit.

dirk-thomas:
http://build.ros.org/job/Krel_reconfigure-jobs/23/consoleFull now shows topological order of the job generation.

tfoote:
+1

dirk-thomas:
Fix regression of package.xml is not in the cache (e.g. release entry has no version): 2351dc9c1c003f7ee111b47dd3d0c81796a11d01

PR #243, opened 2016-03-11
call save on newly created or updated jobs

dirk-thomas:
This will ensure that the modified job configuration is available for the subsequently created / updated job.


tfoote:
+1

PR #244, opened 2016-03-12
always apt-get update in devel jobs

dirk-thomas:
Should fix http://build.ros.org/job/Kdev__rosconsole_bridge__ubuntu_xenial_amd64/1/


tfoote:
+1

PR #245, opened 2016-03-12
use build, run and test dependencies for topological order

dirk-thomas:
Before `message_runtime` was reconfigured very early (http://build.ros.org/job/Krel_reconfigure-jobs/30/) since it only has a build dependency on catkin.

After it is being reconfigured much later (http://build.ros.org/job/Krel_reconfigure-jobs/31/) since its run dependencies on `roscpp_core` are being considered.


tfoote:
+1

PR #246, opened 2016-03-12
add support for arm64 on ubuntu

tfoote:


tfoote:
@dirk-thomas if you have comments I'll address them later. Merging now to get a run through on the test farm tonight. One manual job ran here: http://54.183.26.131:8080/job/Kbin_arm_uXarm64__catkin__ubuntu_xenial_arm64__binary/2/console and failed but at least it didn't upload to the wrong target.

dirk-thomas:
I do see the docker images `osrf/ubuntu_armhf` and `osrf/ubuntu_arm64`. But the existing docker image for 32bit `osrf/ubuntu_32bit` has not been renamed (and probably shouldn't be). So how is this supposed to work for 32bit?

dirk-thomas:
Example of a broken build due to this change: http://build.ros.org/view/Kbin_uW32/job/Kbin_uW32__trajectory_msgs__ubuntu_wily_i386__binary/1/

tfoote:
Fixed in f96ee6e

PR #251, opened 2016-03-13
rebuild dependency graph after reconfiguring jobs

dirk-thomas:
Replaces #243.


dirk-thomas:
While the explicit rebuilding of the dependency graph costs about one minute it ensures that all upstream and downstream relations are established when the job finished: http://build.ros.org/job/Krel_reconfigure-jobs/39/console#console-section-9

PR #252, opened 2016-03-14
change to set the locale via snippet and add support for debian

tfoote:
debian does not have locales in the base image. 

This is to fix builds like this: http://54.183.26.131:8080/job/Ksrc_debianjessie__visualization_msgs__debian_jessie__source/193/console

### DIFF
--- a/ros_buildfarm/templates/devel/devel_task.Dockerfile.em
+++ b/ros_buildfarm/templates/devel/devel_task.Dockerfile.em
@@ -11,8 +11,13 @@ MAINTAINER Dirk Thomas dthomas+buildfarm@@osrfoundation.org
 VOLUME ["/var/cache/apt/archives"]
 
 ENV DEBIAN_FRONTEND noninteractive
-RUN locale-gen en_US.UTF-8
-ENV LANG en_US.UTF-8
+
+@(TEMPLATE(
+    'snippet/setup_locale.Dockerfile.em',
+    os_name=os_name,
+    os_code_name=os_code_name,
+    timezone=timezone,
+))@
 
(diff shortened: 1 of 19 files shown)


ros_buildfarm/templates/devel/devel_task.Dockerfile.em
```diff
@@ -11,8 +11,13 @@ MAINTAINER Dirk Thomas dthomas+buildfarm@@osrfoundation.org
...
 
 ENV DEBIAN_FRONTEND noninteractive
-RUN locale-gen en_US.UTF-8
-ENV LANG en_US.UTF-8
+
+@(TEMPLATE(
+    'snippet/setup_locale.Dockerfile.em',
+    os_name=os_name,
+    os_code_name=os_code_name,
+    timezone=timezone if timezone else None,
```
dirk-thomas: Have you tried to expand this template which doesn't pass in a `timezone` variable?

tfoote: I can't test it since it's available in all templates [here](https://github.com/ros-infrastructure/ros_buildfarm/blob/master/ros_buildfarm/templates/__init__.py#L76) i'll remove the conditional loading since you're right this won't work with timezone unset.

tfoote:
updated to not be conditional on timezone since it's set by the wrapper

dirk-thomas:
LGTM. Please squash before the merge.

tfoote:
Fixup for hard coded os versions: be01b7f50c33d97f38b77534a5efd2995fb51695

PR #253, opened 2016-03-14
install python3 on debian too

tfoote:
It's not in the base image


ros_buildfarm/templates/snippet/install_python3.Dockerfile.em
```diff
@@ -1,4 +1,4 @@
-@[if os_name == 'ubuntu' and os_code_name in ['saucy', 'vivid', 'wily', 'xenial']]@
+@[if os_name == 'debian' or os_name == 'ubuntu' and os_code_name in ['saucy', 'vivid', 'wily', 'xenial']]@
```
dirk-thomas: Use `in` for checking for multiple `os_name`s.

I assume the `os_code_name` check also needs to be updated.

tfoote: They're not symmetric. All versions of debian need it, only the codenames for ubuntu. 
debian || ubuntu && (specific ubuntu codenames) 

There's a level that we could drop this and just force install on all versions. It won't cost us much and will be more generic.

dirk-thomas:
lgtm.

PR #255, opened 2016-03-15
add support for debian jessie short names, and aarch64/arm64 shorthand to v8

tfoote:
@dirk-thomas when this is merged this will it reconfigure and create duplicate jobs for everything?

### DIFF
--- a/ros_buildfarm/common.py
+++ b/ros_buildfarm/common.py
@@ -230,6 +230,7 @@ def get_short_build_name(build_name):
 
 def get_short_os_name(os_name):
     os_name_mappings = {
+        'debian': 'd',
         'ubuntu': 'u',
     }
     return os_name_mappings.get(os_name, os_name)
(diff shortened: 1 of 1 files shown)


ros_buildfarm/common.py
```diff
@@ -231,12 +231,14 @@ def get_short_build_name(build_name):
 def get_short_os_name(os_name):
     os_name_mappings = {
         'ubuntu': 'u',
+        'debian': 'd',
```
dirk-thomas: Please move the new line above `ubuntu`.

dirk-thomas:
Beside the comment lgtm. Yes, it will create the new jobs. You have to remove the old jobs e.g. using a groovy script:

```
import java.util.regex.Matcher
import java.util.regex.Pattern

pattern = Pattern.compile("MYPREFIX.+")

for (p in Jenkins.instance.projects) {
  if (!pattern.matcher(p.name).matches()) continue
  println(p.name)
  //p.delete()
}
```

tfoote:
reordered

PR #256, opened 2016-03-15
Add debian to status page

tfoote:
Tested here: http://54.183.26.131:8080/job/Krel_release-status-page/262/console


dirk-thomas:
lgtm

PR #258, opened 2016-03-16
simplify logic

dirk-thomas:
See https://github.com/ros-infrastructure/ros_buildfarm/commit/f96ee6e1578838c2ba5763fbcc1de1d8ea33a514#commitcomment-16663472

### DIFF
--- a/ros_buildfarm/templates/snippet/from_base_image.Dockerfile.em
+++ b/ros_buildfarm/templates/snippet/from_base_image.Dockerfile.em
@@ -2,11 +2,7 @@
 @[if arch == 'i386']@
 FROM osrf/ubuntu_32bit:@os_code_name
 @[else]@
-@[if arch == 'armhf']@
-FROM osrf/ubuntu_armhf:@os_code_name
-@[else]@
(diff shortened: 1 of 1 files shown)


tfoote:
+1

tfoote:
this was still on my todo list, thanks for cleaning it up. PS I've got a new generic generator for the images at https://github.com/osrf/ubuntu_32bit-docker/tree/generic which can let use produce them all. With it's current implementation it produces osrf/$os_$arch:$codename  I think it would make sense for use to homogenize the naming instead of adding a special case at both ends. also i386 is what upstream is prototyping with https://hub.docker.com/u/i386/

tfoote:
Note that I need to extend this to debian arm64 jessie too 

I just extended the test farm to build debian arm64 but it's building amd64 debs right now due to this logic: http://54.183.26.131:8080/view/Kbin_arm_dJv8/job/Kbin_arm_dJv8__catkin__debian_jessie_arm64__binary/1/

dirk-thomas:
Obsolete with #262 being merged.

PR #259, opened 2016-03-16
fix script generation with Python 2

dirk-thomas:
Since `em.Hook` is not using `object` as the base class the subclass can't use `super` with Python 2.


tfoote:
+1

PR #262, opened 2016-03-16
simplify logic

tfoote:
See https://github.com/ros-infrastructure/ros_buildfarm/commit/f96ee6e1578838c2ba5763fbcc1de1d8ea33a514#commitcomment-16663472

I suggest this instead of #258 

Before we merge this we just make sure to backfill the older images. We can either rebuild them or pull them and rename them via docker.

### DIFF
--- a/ros_buildfarm/templates/snippet/from_base_image.Dockerfile.em
+++ b/ros_buildfarm/templates/snippet/from_base_image.Dockerfile.em
@@ -1,13 +1,5 @@
-@[if os_name == 'ubuntu' and arch in ['i386', 'armhf', 'arm64']]@
-@[if arch == 'i386']@
-FROM osrf/ubuntu_32bit:@os_code_name
-@[else]@
-@[if arch == 'armhf']@
-FROM osrf/ubuntu_armhf:@os_code_name
-@[else]@
-FROM osrf/ubuntu_arm64:@os_code_name
-@[end if]@
-@[end if]@
+@[if arch in ['i386', 'armhf', 'arm64']]@
+FROM osrf/@(os_name)_@arch:@os_code_name
 @[else]@
 FROM @os_name:@os_code_name
 @[end if]@


dirk-thomas:
I don't think renaming the existing Docker 32bit image is a good idea. Existing users might rely on the current name so we would need to maintain both. Also the image is not really `i386` but `,i686` which is why I think `32bit` is the better name.

tfoote:
I agree that 32bit is more "correct" but as I mentioned in #258 `i386` is the term used for upstream builds of this type. And is also the argument to debootstrap to build a 32 bit system. I think that consistency with nomenclature with the rest of the tools used is more valuable than the "accuracy".

We can leave the existing images published, but these are not images that we've heavily publicized and we've never made any promises about their persistence. The only documentation of them is the script that generated them. And if someone really needs them they can then generate their own image using the script.

tfoote:
I've created all the appropriate images for 
* [osrf/ubuntu_i386](https://hub.docker.com/r/osrf/ubuntu_i386/tags/)
* [osrf/ubuntu_armhf](https://hub.docker.com/r/osrf/ubuntu_armhf/tags/) 
* [osrf/ubuntu_arm64](https://hub.docker.com/r/osrf/ubuntu_arm64/tags/)

As well as debian images:
* [osrf/debian_armhf](https://hub.docker.com/r/osrf/debian_armhf/tags/)
* [osrf/debian_arm64](https://hub.docker.com/r/osrf/debian_arm64/tags/)

These images leverage the new standardized script for creating the images: https://github.com/osrf/ubuntu_32bit-docker/pull/1

dirk-thomas:
+0.5 from me. When this is merged please send a notification email to the build farm mailing list to let people know that the docker image is being changed (including as of which version).

tfoote:
indigo and jade rebuilds for non-arm builds complete, merging this.

PR #263, opened 2016-03-16
disable pager for git log command

dirk-thomas:


dirk-thomas:
I will go ahead with the merge but I am happy to address any feedback afterwards.

dirk-thomas:
No Travis check I guess...

PR #264, opened 2016-03-16
add documentation how to use Travis

dirk-thomas:
After some smallish fixes and improvements it is now easily possible to automate running the jobs from the command line.

For an example see the following fork: https://github.com/dirk-thomas/genmsg/blob/indigo-devel/.travis.yml

And the resulting Travis build: https://travis-ci.org/dirk-thomas/genmsg/builds/116466985

### DIFF
--- a/doc/index.rst
+++ b/doc/index.rst
@@ -86,12 +86,16 @@ Run jobs locally
 Some of the job types can be easily run locally.
 This allows to reproduce the behavior of the build farm locally, eases
 debugging, and shortens the round-trip time for testing.
+It also means that you can run any of these jobs anywhere, e.g. on Travis.
 
 * `release jobs#run-the-release-job-locally <jobs/release_jobs.rst#run-the-release-job-locally>`_
   generate binary packages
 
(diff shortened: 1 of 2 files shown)


dirk-thomas:
I added a reference to this in the wiki beside "how to setup pull requests" so users are aware of the alternatives: http://wiki.ros.org/action/diff/buildfarm/Pull%20request%20testing?action=diff&rev1=1&rev2=2

doc/index.rst
```diff
@@ -93,6 +94,9 @@ debugging, and shortens the round-trip time for testing.
 * `devel jobs#run-the-devel-job-locally <jobs/devel_jobs.rst#run-the-devel-job-locally>`_
   build and test ROS repositories
 
+  * `devel jobs#run-the-devel-job-locally <jobs/devel_jobs.rst#run-the-devel-job-on-travis>`_
```
tfoote: copy and paste missed replace, s/locally/travis/

tfoote:
one typo otherwise +1

doc/index.rst
dirk-thomas: I have updated the PR. Please feel free to apply such trivial changes directly in the future.

PR #276, opened 2016-03-31
add more information about jenkins.ini file

dirk-thomas:
Fixes #275.

Issue #275, opened 2016-03-31
jenkins.ini format


dhood:
What is the appropriate format for `~/.buildfarm/jenkins.ini` please?

I'm out of ideas for what to try in order to flesh out the [example](https://github.com/dhood/ros_buildfarm/blob/master/doc/environment.rst#provide-credentials-for-jenkins-master).. keep getting `Could not find credentials for 'http://54.183.26.131:8080' in file '/home/dhood/.buildfarm/jenkins.ini'`. presumably my username is dhood and my password is my api token which I can get at http://build.ros.org/user/dhood/configure without issues

tfoote:
```
[DEFAULT]
username=USERNAME
password=PASSWORD/TOKEN
```

Where DEFAULT can be the host such as http://54.183.26.131:8080 or if you're only using one DEFAULT can be left.  You can have multiple entries with different credentials for different hosts. 
USERNAME and PASSWORD/TOKEN need to be updated

We should add this to the docs. There's been others confused about this.

dirk-thomas:
What is missing or unclear from the existing documentation: https://github.com/ros-infrastructure/ros_buildfarm/blob/master/doc/environment.rst#provide-credentials-for-jenkins-master

tfoote:
Yeah, that does give an example but doesn't explain the format. We should mention that there can be multiple entries for different hosts, and also extend it to explain which github tokens to use, and how to get them, if using github for authentication.

dirk-thomas:
I think the `.ini` format is a standard we don't have to describe in our document. I added a link to the wikipedia page in #276.

Regarding the login: we can't assume in our documentation that Jenkins uses GitHub for login. I added a conditional note about it with a link to the GitHub help.

PR #277, opened 2016-03-31
add doc about Jenkins API token

dirk-thomas:
As suggested in https://github.com/ros-infrastructure/ros_buildfarm/pull/276#issuecomment-204068879

### DIFF
--- a/doc/environment.rst
+++ b/doc/environment.rst
@@ -54,6 +54,9 @@ Create the `.ini <https://en.wikipedia.org/wiki/INI_file>`_ file ``~/.buildfarm/
 
 You can put multiple separate sections for different hosts into the configuration file as well as use a section ``[DEFAULT]`` which is being used if no host specific section is found.
 
+Instead of putting the plain password in this configuration you can use the *API token* of your Jenkins user.
+To get the API token (or change it) go to the *configure* page of your Jenkins user (http://YOUR-JENKINS/me/configure).
+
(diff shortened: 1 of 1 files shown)


dhood:
Can we change the example to use DEFAULT like https://github.com/dhood/ros_buildfarm/commit/48cb947d335c91a92be34497feedf9d4757cc365 ?
Sorry to drag this on since it's such low-priority, but this is what I would have needed in the documentation to prevent opening https://github.com/ros-infrastructure/ros_buildfarm/issues/275 (I thought it was a generic .ini that needed to be changed)

dirk-thomas:
I would rather not use `[DEFAULT]` in the example but keep the specific host. I think using a specific host implicitly documents the configuration and what it is used for. Otherwise when looking at the file later it is unclear from the file content alone.

dhood:
good point. https://github.com/dhood/ros_buildfarm/commit/f66b6790f23c46e8b7123329a8beb66165218444?

dhood:
good point. https://github.com/dhood/ros_buildfarm/commit/ab5319d8948f36e660cea7875489af40647901f6? if not, no problem. it can probably be figured out from the new information now

dirk-thomas:
Since the whole documentation doesn't contain the term "test farm" I would rather not mention it in that paragraph.

dhood:
ok no problem

PR #279, opened 2016-04-07
Add page showing packages blocking the next release

dhood:
On test buildfarm at 54.183.26.131:8080/job/Krel_blocked-releases-page/ws/blocked_releases_page/blocked_releases_kinetic.html
(to disable csp stuff type about:config in address bar and set `security.csp.enable` to false

### DIFF
--- a/doc/jobs/miscellaneous_jobs.rst
+++ b/doc/jobs/miscellaneous_jobs.rst
@@ -30,7 +30,9 @@ Status pages
 The following jobs generate HTML pages to visualize the status of the package
 repositories:
 
-* The ``release-status-page`` job invokes the script
+* The ``release-status-page`` job, generated on the farm by
+  **generate_release_status_page_job.py**, 
+  invokes the script
   *build_release_status_page.py*.
   The generated page shows the ROS packages of a specific ROS distribution and
   their package versions in the three repositories:
@@ -41,7 +43,9 @@ repositories:
 
   Each page shows the packages and targets defined by a *release build file*.
 
-* The ``repos-status-page`` job invokes the script
+* The ``repos-status-page`` job, generated on the farm by
+  **generate_repos_status_page_job.py**,
+  invokes the script
   *build_repos_status_page.py*.
   The generated page shows all packages for a set of three repositories:
 
@@ -51,3 +55,18 @@ repositories:
 
   Each page shows the targets defined by all the release build files using
   the same *target* repository.
+
+* The ``release-compare-page`` job, generated on the farm by
+  **generate_release_compare_page_job.py**,
+  invokes the script
+  *build_release_compare_page.py*.
+  The generated page shows, for a specific pair of ROS distributions, how the versions of the
+  repositories released in the earlier distribution compare in the more recent distribution.
+
+* The ``blocked-releases-page`` job, generated on the farm by
+  **generate_blocked_releases_page_job.py**,
+  invokes the script
+  *build_blocked_releases_page.py*.
+  The generated page shows, for a specific ROS distribution, which repositories from the previous 
+  distribution have not yet been released, and which repositories they are preventing from being 
+  released as a result.
--- a/ros_buildfarm/status_page.py
+++ b/ros_buildfarm/status_page.py
@@ -12,6 +12,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+from __future__ import print_function
+
 from collections import namedtuple
 from distutils.version import LooseVersion
 import itertools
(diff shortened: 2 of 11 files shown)


ros_buildfarm/jenkins.py
```diff
@@ -270,8 +270,8 @@ def _diff_configs(remote_config, new_config):
     if ElementTree.tostring(remote_root) == ElementTree.tostring(new_root):
         return []
 
-    xml1 = ElementTree.tostring(remote_root, encoding='unicode')
-    xml2 = ElementTree.tostring(new_root, encoding='unicode')
+    xml1 = ElementTree.tostring(remote_root)
+    xml2 = ElementTree.tostring(new_root)
```
dirk-thomas: Can you please explain why you changed this? Your commit message says:

> unicode is no longer accepted as an option in ElementTree.tostring

But the docs say something different: https://docs.python.org/3/library/xml.etree.elementtree.html#xml.etree.ElementTree.tostring

> Use encoding="unicode" to generate a Unicode string (otherwise, a bytestring is generated).

ros_buildfarm/templates/status/blocked_releases_page.html.em
```diff
@@ -0,0 +1,80 @@
...
+            @(row[col])
+        @[end for]@
+      </tr>
+      @[end for]@
+    </tbody>
+  </table>
+
+  <script type="text/javascript">window.body_done();</script>
+</body>
+</html>
```
dirk-thomas: Please use a consistent indentation throughout the file, e.g. two spaces.

ros_buildfarm/status_page.py
```diff
@@ -530,6 +530,296 @@ def _get_comparable_loose_versions(version_str1, version_str2):
...
+    valid_rosdistro_names.sort()
+    if rosdistro_name is None:
+        rosdistro_name = valid_rosdistro_names[-1]
+    print('Checking packages for "%s" distribution' % rosdistro_name)
+
+    # Find the previous distribution to the current one
+    try:
+        i = valid_rosdistro_names.index(rosdistro_name)
+    except ValueError:
+        print('Distribution key not found in list of valid distributions.')
```
dirk-thomas: Error messages should go to `stderr`. Use `from __future__ import print_function` at the top of the file and then `print(..., file=sys.stderr)`.

ros_buildfarm/common.py
```diff
@@ -10,9 +10,7 @@ class JobValidationError(Exception):
     if validation fails, e.g. because the requested package
     is not available in the specified index.yaml
     """
-
-    def __init__(self, message):
-        self.message = message
+    pass
```
dirk-thomas: Have you tried that in the case of such an exception the message is still shown as before?

ros_buildfarm/templates/status/js/setup.js
```diff
@@ -238,15 +238,24 @@ function filter_table() {
...
+              }
               break;
             }
           }
           if(is_known_query) {
-            // search in full row html
-            if (row[0].indexOf(queries[i]) == -1) return null;
+            if(is_id_query) {
+              // search only for ids matching the query
+              if (row[0].indexOf('id="'+queries[i]+'"') == -1) return null;
```
dirk-thomas: Please use spaces around operators since it helps readability.

Same above.

ros_buildfarm/status_page.py
```diff
@@ -530,6 +530,296 @@ def _get_comparable_loose_versions(version_str1, version_str2):
...
+            repo for repo in keys if _is_released(repo, prev_distro_file))
+        repo_names = prev_repo_names
+        ignored_inputs = []
+    else:
+        prev_repo_names = set(
+            repo for repo in repo_names if _is_released(repo, prev_distro_file))
+        ignored_inputs = list(set(repo_names).difference(prev_repo_names))
+        if len(ignored_inputs) > 0:
+            print(
+                'Ignoring inputs for which repository info not found in previous distribution' +
```
dirk-thomas: You can skip the trailing `+`. The strings will still be concatenated.

ros_buildfarm/status_page.py
```diff
@@ -530,6 +530,296 @@ def _get_comparable_loose_versions(version_str1, version_str2):
...
+    row['name'] = _table_cell_wrap(name_cell, name)
+
+    # whether or not the repo has been released
+    row['version'] = _table_cell_wrap(data.get('version', 'Not released'))
+
+    # num repos blocked by
+    row['num_repos_blocked_by'] = _table_cell_wrap(len(repos_blocked_by))
+
+    # repos blocked by
+    row['repos_blocked_by'] = _table_cell_wrap('<br>'.join(
```
dirk-thomas: In the case someone wants to parse the output its easier to generate valid XML. Please use `<br />`.

Same below.

ros_buildfarm/status_page.py
```diff
@@ -530,6 +530,296 @@ def _get_comparable_loose_versions(version_str1, version_str2):
...
+        if len(repos_blocked_by) > 0:
+            tags += _filter_tag_wrap('BLOCKED')
+        elif len(repos_blocking) > 0:
+            tags += _filter_tag_wrap('UNBLOCKED_BLOCKING')
+        else:
+            tags += _filter_tag_wrap('UNBLOCKED_UNBLOCKING')
+
+    url = data.get('url', None)
+    if url:
+        name_html = '<a href={0}>{1}</a>'.format(url, name)
```
dirk-thomas: The url needs to be in quotes.

ros_buildfarm/templates/status/js/setup.js
```diff
@@ -238,15 +238,24 @@ function filter_table() {
       result_rows = $.map(window.rows, function(row) {
         for (var i = 0; i < queries.length; i++) {
           var is_known_query = false;
+          var is_id_query = false;
           for(var q in QUERY_TRANSFORMS) {
-            if (queries[i] === QUERY_TRANSFORMS[q]) {
+            if (RegExp("^("+QUERY_TRANSFORMS[q]+")$").test(queries[i])) {
               is_known_query = true;
+              if (q === "id") {
```
dirk-thomas: Currently only a single query can search for ids since it is being identified by the key. If this would check e.g. if the value starts with `id=` it would be possible to add arbitrary id -based queries in the future.

ros_buildfarm/templates/status/js/setup.js
```diff
@@ -238,15 +238,24 @@ function filter_table() {
       result_rows = $.map(window.rows, function(row) {
         for (var i = 0; i < queries.length; i++) {
           var is_known_query = false;
+          var is_id_query = false;
           for(var q in QUERY_TRANSFORMS) {
-            if (queries[i] === QUERY_TRANSFORMS[q]) {
+            if (RegExp("^("+QUERY_TRANSFORMS[q]+")$").test(queries[i])) {
```
dirk-thomas: I am torn if we need to use a regexp for this and also interpret all other plain text queries as regexp. Not sure how the alternative would look like though...

ros_buildfarm/templates/status/js/setup.js
dhood: Yeah, I had tried it that way but ran into issues when there were special characters involved
e.g. I wanted [this](https://github.com/ros-infrastructure/ros_buildfarm/pull/279/files#diff-e404e3f34fac28ad7e08164c36763016R38) to link to `?q=id=metapackages` but it would get cut short because of the two `=`s, so I resorted to using dots

FWIW, this is just the name of the query, not the query itself https://github.com/ros-infrastructure/ros_buildfarm/pull/279/files#diff-e404e3f34fac28ad7e08164c36763016R14

ros_buildfarm/templates/status/js/setup.js
dhood: yeah, I did it this way to keep it compatible with the other status pages that are using setup.js but perhaps it makes sense to duplicate the file and make it specific to the blocked repos page (which is only using regex queries)

ros_buildfarm/templates/status/js/setup.js
dirk-thomas: What about `id:xyz`?

ros_buildfarm/templates/status/js/setup.js
dirk-thomas: I would rather avoid duplication since the file is big and it will be quite some effort to exact this portion into a separate file.

It's ok to keep the regexp. I was more think about impact where it isn't needed. Have you compared how it impacts e.g. the filtering on the huge Indigo page?

ros_buildfarm/templates/status/js/setup.js
dhood: tried it :disappointed:

ros_buildfarm/jenkins.py
dhood: this was to get it to work with python 2.7 (before I realised it was targeting python 3) so I will revert it and https://github.com/dhood/ros_buildfarm/commit/e4b4cbca1ecf40a4f2e58a8aa4f48851309f1890

ros_buildfarm/templates/status/js/setup.js
dhood: yeah I was testing with it too, the rest of the pages' queries are here https://github.com/dhood/ros_buildfarm/blob/add-blocking-script/ros_buildfarm/templates/status/js/setup.js#L7 but I suppose there is a risk that in the future someone will want to use a query called id for a different purpose

dirk-thomas:
What are the remaining items for this:

* Should the resulting html be published somewhere else? Or should be keep it within the jobs workspace to make it easier to delete it when not needed anymore
* I guess we don't want to generate the job by default (based on the build farm config). But It needs some documentation that it exists and how to invoke it.
* What about the CSP? Is that still necessary if we move it to the official build farm (or not since it has a real domain name)?

dhood:
For the third point, other pages on test build farm have the same issue so it either gets fixed automatically when moving to the official build farm or it's something that someone knows how to fix

For the other points, all I can add is that I _think_ these questions can be extended to the other status pages added in this commit: https://github.com/ros-infrastructure/ros_buildfarm/commit/d7b0ecd5d5d759363e89bd7d3ea4e83645fcd8ca because I was trying to copy it as closely as possible

dirk-thomas:
The other status pages are being generated by the `generate_all` script and are mentioned in the docs: https://github.com/ros-infrastructure/ros_buildfarm/blob/master/doc/jobs/miscellaneous_jobs.rst#status-pages.

Since the other pages work fine on the official farm it might "fix itself" when being moved. But the other pages are served by Apache rather then Jenkins (if we keep it within the jobs workspace).

dhood:
OK I mentioned it in the documentation

I have discovered that http://54.183.65.232/status_page/blocked_releases_kinetic.html is live, and doesn't require disabling CSP. I think it might have been [this line](https://github.com/ros-infrastructure/ros_buildfarm/pull/279/files#diff-e92455f36410fb0669e723716f2f292aR91). Filtering by repo id doesn't work on that page because it's using the old setup.js. But I think this covers points 1 and 3

ros_buildfarm/templates/status/blocked_releases_page.html.em
```diff
@@ -0,0 +1,80 @@
...
+      <input type="text" name="q" id="q" />
+      <p>
+        <a href="?q=" title="Show all repos">all</a>,
+        <a href="?q=_RELEASED" title="Repositories which have already been released">released</a>,
+        <a href="?q=_UNRELEASED" title="Repositories which have not been released">unreleased</a>,
+        <a href="?q=_BLOCKED" title="Repositories which are blocked from being released because of unreleased dependencies">blocked</a>,
+        <a href="?q=_UNBLOCKED" title="Repositories which can be released">releasable</a>,
+        <a href="?q=_UNBLOCKED_BLOCKING" title="Repositories which can be released and are preventing others from being released">releasable and blocking</a>,
+        <a href="?q=_UNBLOCKED_UNBLOCKING" title="Repositories which can be released and are not preventing others from being released">releasable and not blocking</a>,
+        <a href="?q=.metapackages" title="Repositories which are dependencies of the desktop-full metapackage">desktop-full</a>
```
jacquelinekay: I think the . needs to be removed from this query string. When I visit http://54.183.65.232/status_page/blocked_releases_kinetic.html and click on "desktop_full", the search bar gets `.metapackages` and no results are returned, but if I remove the dot it seems to work fine.

dhood: if you try a different package, without the dot it just searches everywhere in the row, so doesn't really do what it's supposed to. that site you linked to is only partially up-to-date (using an non-updated javascript file needed for the dot queries)

jacquelinekay: Got it. From brief inspection of the `setup.js` it looks like search strings will search across all columns by default and a regex is needed to filter by a column (e.g. repo name), is that correct? Maybe there should be a guide to the regex keys somewhere near the top of the page. Also sorry if this is outdated feedback based on the state of the site right now, I'm not generating it locally.

jacquelinekay:
Is it ok if I send an email to ros-release with a link to the test site, even if this PR isn't merged yet? It has a lot of useful info for maintainers. I can put a disclaimer that it's not the final version and tell everyone to report bugs if they find them.

ros_buildfarm/templates/status/blocked_releases_page.html.em
dhood: If you load the page in [this comment](https://github.com/ros-infrastructure/ros_buildfarm/pull/279#issue-146475447) instead you'll get the one using the right `setup.js`

dhood:
give me a minute and I'll try to get that `setup.js` file to update on http://54.183.65.232/status_page/blocked_releases_kinetic.html, then all of the functionality should be there

dhood:
this one's working fine http://54.183.65.232/status_page/blocked_releases/blocked_releases_kinetic.html 
the display's not great on smaller screens but I can iterate on that

jacquelinekay:
Nice

ros_buildfarm/status_page.py
```diff
@@ -530,6 +531,298 @@ def _get_comparable_loose_versions(version_str1, version_str2):
...
+
+
+def _is_released(repo, dist_file):
+    return repo in dist_file.repositories and \
+        dist_file.repositories[repo].release_repository is not None and \
+        dist_file.repositories[repo].release_repository.version is not None
+
+
+def _get_blocked_releases_info(
+        config_url, rosdistro_name, repo_names=None, depth=1):
```
scpeters: I think using `depth=1` keeps the cumulative total from being accurate. Can we make the default 10?

dhood:
@dirk-thomas I'm going to make some changes on this branch and will ping you when it's stable again to continue review

ros_buildfarm/status_page.py
dhood: @scpeters do you see what you expect for libccd now? http://54.183.65.232/status_page/blocked_releases/blocked_releases_kinetic.html?q=fcl

scpeters: this is absolutely glorious! yes, it's what I was hoping to see :)

dhood:
on my side we're good to continue thanks @dirk-thomas

ros_buildfarm/templates/status/blocked_releases_page.html.em
```diff
@@ -0,0 +1,111 @@
...
+      'maintainers_of_repos_blocked_by',
+      'num_repos_recursively_blocked',
+      'num_repos_blocked',
+      'repos_blocked',
+      ]]@
+      <td @(' class="thincol"' if col in ['version',] or col.startswith('num') else '')>@(row[col])</td>
+    @[end for]@
+    </tr>
+  @[end for]@
+    </tbody>
```
dhood: @dirk-thomas I've arranged the template code like this so that the resulting html looks correctly indented. is it better to have the template code indented differently and run an html beautifier on the generated file?

dirk-thomas: I wouldn't use a html beautifier since it just another indirection. These small snippets can easily be written with clean html markup without needing cleanup afterwards.

I haven't checked the generated html but the template itself looks inconsistent (e.g. the indentation of the `tbodys and `tr` tags). It might be fine the the template has been expanded (e.g. due to the spaces in front of `@[for`) but it is hard to see from looking at the template. I would recommend to update the template the following way which makes it easy to see the structure and correct indentation in the template and allows to easily conclude how the generated output will look like:

* all line with `@[...]@` should have no leading spaces, to keep the visual indentation of the template logic you can put spaces between `@[` and the keyword `if` / `for` (same as ros2/rosidl_dds#32)
* then all lines with html code can be indented easily to line up opening and closing tags

```
    <tbody>
      <script type="text/javascript">window.tbody_ready();</script>
@[for row in repos_data]@
      <tr>
@[  for col in [...]]@
        <td @(' class="thincol"' if col in ['version',] or col.startswith('num') else '')>@(row[col])</td>
@[  end for]@
      </tr>
@[end for]@
    </tbody>
```

dhood: perfect, thanks for that. I was only moving the `@` to the beginning of the line so was getting errors. have requested membership in that team so I get pinged in the future

dhood:
anything other reviews for this @dirk-thomas?

doc/jobs/miscellaneous_jobs.rst
```diff
@@ -51,3 +55,18 @@ repositories:
...
+  The generated page shows, for a specific pair of ROS distributions, how the versions of the
+  repositories released in the earlier distribution compare in the more recent distribution.
+
+* The ``blocked-releases-page`` job, generated on the farm by
+  **generate_blocked_releases_page_job.py**,
+  invokes the script
+  *build_blocked_releases_page.py*.
+  The generated page shows, for a specific ROS distribution, which repositories from the previous 
+  distribution have not yet been released, and which repositories they are preventing from being 
+  released as a result.
```
dirk-thomas: Maybe mention for these two paragraphs that they are *not* part of the generate-all script?

ros_buildfarm/status_page.py
```diff
@@ -530,6 +531,303 @@ def _get_comparable_loose_versions(version_str1, version_str2):
...
+        h.write(html)
+
+    additional_resources(output_dir, copy_resources=copy_resources)
+
+
+def _div_wrap(data, name=None):
+    if name:
+        return '<div id="{0}">{1}</div>'.format(name, data)
+    else:
+        return '<div>{0}</div>'.format(data)
```
dirk-thomas: The main logic in both branches is the same. This could be changed into something without duplication like:

    if_attr = ' id="%s"' % name if name is not None else ''
    return '<div{0}>{1}</div>'.format(if_attr, data)

ros_buildfarm/templates/status/blocked_releases_page.html.em
```diff
@@ -0,0 +1,111 @@
...
+  </script>
+  <script type="text/javascript" src="js/setup.js?@(resource_hashes['setup.js'])"></script>
+  <script type="text/javascript">
+    var QUERY_TRANSFORMS = {
+      'id': 'id=".*"',
+      'label': 'label=".*"',
+    };
+
+    function encoded_query_a_tag(query, title, content) {
+      document.write("<a href=\"?q=" + encodeURIComponent(query) + "\" title=\"" + title + "\">" + content + "</a>")
```
dirk-thomas: By using single quotes for the outside strings you avoid requiring escaping.

Same below.

doc/jobs/miscellaneous_jobs.rst
dhood: I think it might just apply to the second paragraph https://github.com/ros-infrastructure/ros_buildfarm/blob/master/scripts/generate_all_jobs.py#L116
I didn't have a particular reason for not including the blocked-releases-page job in the generate-all script, I just didn't see it. add it?

dirk-thomas: If the blocked releases job can be generates as part of generate-all that would be even better. I am not sure if it needs any further configuration options though.

dhood: ok think that should be done now

dhood:
Thanks for your patience with this, I know it's been "a while"!
This is live for kinetic at http://repositories.ros.org/status_page/blocked_releases_kinetic.html and I believe good to merge now.

dhood:
note that the different status page jobs are over-writing the `setup.js` file, so the page will only filter correctly if it was the last of the status page jobs to have run.

I've temporarily put it at http://repositories.ros.org/status_page/blocked_releases_page/blocked_releases_kinetic.html so that the files don't conflict, for easier review. I'll remove that directory once merged.

ros_buildfarm/templates/status/blocked_releases_page.html.em
```diff
@@ -0,0 +1,121 @@
...
+        <script language="JavaScript">encoded_query_a_tag('label="UNRELEASED"', "Repositories which have not been released", "unreleased")</script>,
+        <script language="JavaScript">encoded_query_a_tag('label="BLOCKED"', "Repositories which are blocked from being released because of unreleased dependencies", "blocked")</script>,
+        <script language="JavaScript">encoded_query_a_tag('label="UNBLOCKED"', "Repositories which can be released", "releasable")</script>,
+        <script language="JavaScript">encoded_query_a_tag('label="UNBLOCKED_BLOCKING"', "Repositories which can be released and are preventing others from being released", "releasable and blocking")</script>,
+        <script language="JavaScript">encoded_query_a_tag('label="UNBLOCKED_UNBLOCKING"', "Repositories which can be released and are not preventing others from being released", "releasable and not blocking")</script>,
+        <script language="JavaScript">encoded_query_a_tag('id="metapackages"', "Repositories which are dependencies of the desktop-full metapackage", "desktop-full")</script>
```
dirk-thomas: The `desktop-full` filter is actually filtering for `metapackages` which contains other variants. I would suggest to rename the filter to match the actual criteria.

ros_buildfarm/templates/status/blocked_releases_page.html.em
```diff
@@ -0,0 +1,121 @@
...
+
+  <div class="table-div">
+    <table>
+      <thead>
+        <tr>
+          <!-- warning: if titles run onto two lines, first row of table data may be affected -->
+          <th class="sortable"><div>Repository</div></th>
+          <th class="sortable"><div>Version</div></th>
+          <th class="sortable">
+            <div title="Number of repositories which are directly blocking this one by not being released">
```
dirk-thomas: I found the current title to be difficult to understand - especially since "by not being released" follows at the end but refers to the first part of the sentence. I would suggest the following: 

> Number of unreleased repositories which are directly blocking this one

I will add similar comments to the other columns.

ros_buildfarm/templates/status/blocked_releases_page.html.em
```diff
@@ -0,0 +1,121 @@
...
+              # blocking release
+            </div>
+          </th>
+          <th class="sortable">
+            <div title="Repositories which are directly blocking this one by not being released">
```
dirk-thomas: Unreleased repositories which are directly blocking this one

ros_buildfarm/templates/status/blocked_releases_page.html.em
```diff
@@ -0,0 +1,121 @@
...
+              Blocking repos
+            </div>
+          </th>
+          <th class="sortable">
+            <div title="Maintainers of the repositories which are directly blocking this one by not being released">
```
dirk-thomas: Maintainers of the unreleased repositories which are directly blocking this one

ros_buildfarm/templates/status/blocked_releases_page.html.em
```diff
@@ -0,0 +1,121 @@
...
+              Maintainers of blocks
+            </div>
+          </th>
+          <th class="sortable">
+            <div title="Number of repositories which are being directly or indirectly blocked by this one not being released">
```
dirk-thomas: Number of unreleased repositories which are directly or indirectly blocking this one

ros_buildfarm/templates/status/blocked_releases_page.html.em
```diff
@@ -0,0 +1,121 @@
...
+              # recursively blocked
+            </div>
+          </th>
+          <th class="sortable">
+            <div title="Number of repositories which are being directly blocked by this one not being released">
```
dirk-thomas: Number of repositories which are being directly blocked by this

ros_buildfarm/templates/status/blocked_releases_page.html.em
```diff
@@ -0,0 +1,121 @@
...
+              # directly blocked
+            </div>
+          </th>
+          <th class="sortable">
+            <div title="Repositories which are being directly blocked by this one not being released">
```
dirk-thomas: Repositories which are being directly blocked by this one

ros_buildfarm/templates/status/blocked_releases_page.html.em
```diff
@@ -0,0 +1,123 @@
...
+          <th class="sortable">
+            <div title="Repositories that are being directly blocked by this one">
+              Directly blocked repos
+            </div>
+          </th>
+        </tr>
+      </thead>
+      <tbody>
+        <!-- Originally sort the table by the repos with the most recursively blocked dependencies -->
+        <script type="text/javascript">window.sort=6; window.reverse=1;</script>
```
dirk-thomas: I would suggest a secondary column to sort by (the name) since there can be rows with the same values and depending in the selected filter all rows might have no blocked deps.

ros_buildfarm/templates/status/blocked_releases_page.html.em
```diff
@@ -0,0 +1,123 @@
...
+        <script language="JavaScript">encoded_query_a_tag('label="UNRELEASED"', "Repositories that have not been released", "unreleased")</script>,
+        <script language="JavaScript">encoded_query_a_tag('label="BLOCKED"', "Repositories that are blocked from being released because of unreleased dependencies", "blocked")</script>,
+        <script language="JavaScript">encoded_query_a_tag('label="UNBLOCKED"', "Repositories that can be released", "releasable")</script>,
+        <script language="JavaScript">encoded_query_a_tag('label="UNBLOCKED_BLOCKING"', "Repositories that can be released and are preventing others from being released", "releasable and blocking")</script>,
+        <script language="JavaScript">encoded_query_a_tag('label="UNBLOCKED_UNBLOCKING"', "Repositories that can be released and are not preventing others from being released", "releasable and not blocking")</script>,
+        <script language="JavaScript">encoded_query_a_tag('id="metapackages"', "Repositories that are dependencies of the metapackages repository", "metapackages")</script>
```
dirk-thomas: The replacement of `which` with `that` in e0715a3a6d63c1747bee7f7f258dbf60d1dc1754 doesn't make the text more readable imo.

dhood: it's an english correction, it wasn't for readability (despite what the commit message says - I just lumped the changes together)

ros_buildfarm/templates/status/blocked_releases_page.html.em
dhood: thanks for the suggestion: https://github.com/ros-infrastructure/ros_buildfarm/pull/279/commits/a7ca2c6d23e71b4ef4cdfe171ce88c1a19d7ae21

dirk-thomas:
After the pending comment has been addressed it would be good to regenerate the page to see the new sorting in action.

dhood:
seems there's a difference in behaviour for chrome and firefox: for firefox, the sort on the #blocked dependencies is stable and doesn't affect the alphabetical order of the package names. For chrome, the sort is unstable, and the package names are not in alphabetical order once sorted on #blocked dependencies.
I've added a todo for this as we discussed offline it's probably not worth addressing it at the moment

dirk-thomas:
I made the sort `stable` in 8ad02634d468d8e02535e3bad35cf4bb24caf4e5. The change is also deployed to the [current status page](http://repositories.ros.org/status_page/blocked_releases_page/blocked_releases_kinetic.html).

dhood:
looks great, thanks!

dirk-thomas:
I think this is ready to be squashed and merged :ship:

dhood:
:champagne: :tada: thank you @dirk-thomas and @tfoote for your help with this - I learnt a lot!
I'll merge this, remove the temporary `blocked_releases_page/` subdirectory on the repositories server, and push the new page-generation jobs to the farm.

dhood:
ah, nevermind, I thought I have write access these days. Well then can someone merge this please and I'll do the other two tasks?

dirk-thomas:
I have done the squash & merge and also invited you to the org unit.

dhood:
thank you. the jobs have been pushed and the following pages are live now in their final location:
http://repositories.ros.org/status_page/blocked_releases_indigo.html
http://repositories.ros.org/status_page/blocked_releases_jade.html
http://repositories.ros.org/status_page/blocked_releases_kinetic.html

dirk-thomas:
@dhood Can you please take a look why [this job](http://build.ros.org/job/Lrel_blocked-releases-page/326/console) failed.

dhood:
thanks for the heads up - I will look into it

dhood:
Please see https://github.com/ros-infrastructure/ros_buildfarm/pull/407

PR #280, opened 2016-04-07
Add locale to locale.gen before setting the locale in Debian

jacquelinekay:
According to the warnings on the failing Kinetic job reported by @dlu, it appears that the locale is not set properly on Debian.

According to these guides, the line must be added before running locale-gen.

https://wiki.debian.org/Locale
https://people.debian.org/~schultmc/locales.html


jacquelinekay:
Closing in favor of Dirk's change

PR #281, opened 2016-04-07
fix locale on Debian

jacquelinekay:
Running this on the farm now to see if it fixes the locale issues for Debian: http://build.ros.org/job/Ksrc_dJ__map_msgs__debian_jessie__source/108


jacquelinekay:
The build was a success! I will change the branch name back to master on the job configuration and merge this.

Issue #285, opened 2016-04-15
[docker-based prerelease] --level argument not working?


130s:
In the following, I thought I excluded downstream testing by `--level 0`, with which I expect 0 packages to be tested in `catkin_workspace_overlay`, but I see this many pkgs being tested (and thus the prerelease test seems never-ending on my machine). Am I wrong?

```
$ generate_prerelease_script.py   https://raw.githubusercontent.com/ros-infrastructure/ros_buildfarm_config/production/index.yaml   indigo default ubuntu trusty amd64   moveit_core   --level 0   --output-dir ./
$ ./prerelease.sh 
```
```
$ ls catkin_workspace/src/
moveit_core  moveit_msgs  moveit_ros
$ ls catkin_workspace_overlay/src/
abb_irb2400_moveit_config          fetch_moveit_config            pepper_moveit_config
abb_irb6640_moveit_config          force_torque_sensor_calib      play_motion
atlas_moveit_config                grasping_msgs                  pr2_arm_kinematics
atlas_v3_moveit_config             hironx_moveit_config           pr2eus_moveit
baxter_moveit_config               husky_ur5_moveit_config        pr2_moveit_config
baxter_sim_kinematics              industrial_trajectory_filters  pr2_moveit_tutorials
carl_moveit                        iri_wam_moveit_config          pr2_teleop_general
clam_moveit_config                 jaco_moveit_config             r2_moveit_generated
cob_collision_monitor              jsk_ik_server                  robot_calibration
cob_grasp_generation               jsk_interactive_marker         romeo_moveit_actions
cob_kinematics                     jsk_pcl_ros                    romeo_moveit_config
cob_lookat_action                  jsk_pcl_ros_utils              simple_grasping
cob_moveit_config                  katana                         sr_robot_msgs
cob_obstacle_distance              mico_moveit_config             staro_moveit_config
cob_obstacle_distance_moveit       moveit_commander               tiago_moveit_config
cob_pick_place_action              moveit_core                    turtlebot_arm_block_manipulation
fanuc_lrmate200ic5h_moveit_config  moveit_full                    turtlebot_arm_moveit_config
fanuc_lrmate200ic5l_moveit_config  moveit_planners_ompl           turtlebot_arm_moveit_demos
fanuc_lrmate200ic_moveit_config    moveit_python                  ur10_moveit_config
fanuc_m10ia_moveit_config          moveit_setup_assistant         ur3_moveit_config
[...]

PR #286, opened 2016-04-15
[prerelease] Remove remains from previous prerelease test run if anything.

130s:
`prerelease.sh` actually warns users as `This script should be executed either in an empty folder or in a folder only containing the prerelease scripts`. But remaining folders from the previous test may harm the next test (e.g. #285).

With this change, when users choose `yY` then the script removes the remaining folders.

### DIFF
--- a/ros_buildfarm/templates/prerelease/prerelease_script_check.sh.em
+++ b/ros_buildfarm/templates/prerelease/prerelease_script_check.sh.em
@@ -2,12 +2,12 @@ set +e
 _ls_prerelease_scripts=`ls | grep -v "prerelease.*\.sh"`
 set -e
 if [ "$_ls_prerelease_scripts" != "" ]; then
-  echo "This script should be executed either in an empty folder or in a folder only containing the prerelease scripts" 1>&2
+  echo "Files from previous prerelease tests are found. These files help speeding things up if you mean to repeat the test for the same repository. If you're testing different repository than the one in previous test, you need to remove them as this script should be executed either in an empty folder or in a folder only containing the prerelease scripts" 1>&2
   if [ "$1" != "-y" ]; then
     read -p "Do you wish to continue anyway, this might overwrite existing files and folders? (y/N) " answer
     case $answer in
       [yY]* ) ;;
-      * ) exit 1;;
+      * ) echo "Exiting. If you're testing difrrent repo, you might want to remove the following folders: ${_ls_prerelease_scripts}." && exit 1;;
     esac
   fi
   echo ""


dirk-thomas:
By running the script in a not empty folder a repeated invocation can be faster then the first invocation. Therefore the script should not delete the content of the folder.

130s:
I wasn't aware of that tip.
Ok, I updated the commit. This does:
- Elaborate how remaining files from previous test can effect the next test for the same repo.
- If user quits the program, list the folders s/he needs to delete.

```
$ ./prerelease.sh 
Prerelease script

Use workspace: /tmp/prerelease_mock

Files from previous prerelease tests are found. These files help speeding things up if you mean to repeat the test for the same repository. If you're testi
ng different repository than the one in previous test, you need to remove them as this script should be executed either in an empty folder or in a folder o
nly containing the prerelease scripts
Do you wish to continue anyway, this might overwrite existing files and folders? (y/N) 

(Entered N)

Exiting. If you're testing difrrent repo, you might want to remove the following folders: catkin_workspace
catkin_workspace_overlay
docker_build_and_install
docker_build_and_test
docker_generating_dockers
keys
ros_buildfarm
wrapper_scripts.
```

ros_buildfarm/templates/prerelease/prerelease_script_check.sh.em
```diff
@@ -2,12 +2,12 @@ set +e
 _ls_prerelease_scripts=`ls | grep -v "prerelease.*\.sh"`
 set -e
 if [ "$_ls_prerelease_scripts" != "" ]; then
-  echo "This script should be executed either in an empty folder or in a folder only containing the prerelease scripts" 1>&2
+  echo "Files from previous prerelease tests are found. These files help speeding things up if you mean to repeat the test for the same repository. If you're testing different repository than the one in previous test, you need to remove them as this script should be executed either in an empty folder or in a folder only containing the prerelease scripts" 1>&2
```
dirk-thomas: This makes the assumption that the folder does contain "files from previous prerelease tests". But that doesn't have to be the case. The user might have just called the script the first time but passed a non-empty folder. So I think this needs to be rephrased more carefully.

ros_buildfarm/templates/prerelease/prerelease_script_check.sh.em
```diff
@@ -2,12 +2,12 @@ set +e
...
   if [ "$1" != "-y" ]; then
     read -p "Do you wish to continue anyway, this might overwrite existing files and folders? (y/N) " answer
     case $answer in
       [yY]* ) ;;
-      * ) exit 1;;
+      * ) echo "Exiting. If you're testing difrrent repo, you might want to remove the following folders: ${_ls_prerelease_scripts}." && exit 1;;
```
dirk-thomas: The variable contains the names of all shell scripts. So the message seems to not make sense to me.

dirk-thomas:
Are you planning to update the PR some time in the future? If not I would rather close this then keeping it open.

PR #293, opened 2016-05-17
add option for compiler warning in devel jobs to make it unstable

dirk-thomas:


tfoote:
+1, though the change in the rst looks unrelated and will be inaccurate until we release again

dirk-thomas:
I removed the changes to the travis instruction from this PR.

But they are not inaccurate and don't require a new release. The old not is about 1.0.1 and we have long released 1.1.0 already.

dirk-thomas:
The logic in the `publisher_warnings` snippet needed fixing to pass the argument value `0` correctly: b25438fe9185178713d5085a5c900af1a9676f09

PR #294, opened 2016-05-18
disable git pager when performing 'git log' on ros_buildfarm.

felixduvallet:
This prevents the build job from blocking when `.gitconfig` contains a `pager = ` entry.

I believe this is in line with other instances of calling `git log`, for example in:

ros_buildfarm/ros_buildfarm/templates/devel/devel_script_clone.sh.em


tfoote:
lgtm

dirk-thomas:
Thanks for the patch.

PR #299, opened 2016-05-23
More compare status pages

wjwwood:
This change does a few things:

First it changes the generation of compare status pages to do a page for all previous ros distros, and a pair for each previous ros distro against the current one. For example:

- distro: kinetic
  - old distros: indigo, jade
  - generates:
    - indigo vs jade vs kinetic
    - indigo vs kinetic
    - jade vs kinetic
  - previously generated:
    - jade vs kinetic
- distro: jade
  - old distros: indigo
  - generates:
    - indigo vs jade
  - previously generated:
    - indigo vs jade
- distro: indigo
  - old distros: none
  - generates:
    - none
  - previously generated:
    - none

I was concerned because when jade goes EOL, then the kinetic job would stop updating the "indigo vs jade vs kinetic" and the packages in the kinetic and indigo columns would get out of date.
So, this pr also adds an age box to both the compare status page and the distro status pages which displays the time since it the page was generated. The box will start out as green (ok), turn yellow after 15 minutes (somewhat stale), and after 1 day it turns red (there's a problem). It does this by embedding a unix timestamp in the html file when it is generated and comparing to the current time when the page is rendered. I imported a js library called `moment.js` to do this: http://momentjs.com/

I ran the `./scripts/status/build_release_status_page.py` and the `./scripts/status/build_release_compare_page.py` scripts locally and they work, but I haven't tested the generate jobs scripts yet.

The need for this was so I can link to a "kinetic vs indigo" page from the Kinetic installation instructions (so people can quickly check what packages are missing in Kinetic that they might care about).


wjwwood:
Here are some screenshots of the new age box:

screen shot 2016-05-23 at 2 51 11 pm

screen shot 2016-05-23 at 2 51 21 pm

wjwwood:
We fixed some bugs in the job generation while deploying it: 8ac6ec91fc4a0e414aea556f15463895b8f0166f

ros_buildfarm/templates/status/release_compare_page.html.em
```diff
@@ -14,7 +15,12 @@
   <link rel="stylesheet" type="text/css" href="css/compare_page.css?@(resource_hashes['compare_page.css'])" />
 </head>
 <body>
-  <script type="text/javascript">window.body_ready();</script>
+@{
+import time
```
dirk-thomas: The template already has a `start_time`. So this should not generate a new timestamp but use the existing one.

wjwwood: `start_time` is a formatted string. I'm expecting a unix timestamp in floating point seconds.

wjwwood: It's also `time.localtime()`, so I'd need a new variable anyways to capture `time.time()`. I can move it into a template variable I suppose though.

wjwwood: https://github.com/ros-infrastructure/ros_buildfarm/pull/300

PR #300, opened 2016-05-24
use template variable for unix start time

wjwwood:
Follow up to https://github.com/ros-infrastructure/ros_buildfarm/pull/299/files/9e051679a3fa3e992352b04c52fc7d28018909c4#r64372539

### DIFF
--- a/ros_buildfarm/status_page.py
+++ b/ros_buildfarm/status_page.py
@@ -26,8 +26,7 @@ def build_release_status_page(
     from rosdistro import get_cached_distribution
     from rosdistro import get_index
 
-    start_time = time.localtime()
-    unix_start_time = time.time()
+    start_time = time.time()
 
     config = get_config_index(config_url)
     release_build_files = get_release_build_files(config, rosdistro_name)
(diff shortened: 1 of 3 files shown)


dirk-thomas:
I would suggest to only pass a single `start_time = time.time()` into the template. The template can then use the raw time as well as derive a local string representation from that.

wjwwood:
Why? Is the goal to minimize the number of template variables or the number of times we get the current time? In both cases, it doesn't seem worth the extra work...

dirk-thomas:
The arguments passed between the "controller" and the "view" is considered to be the interface. And commonly the goal is to minimize the interface of two separate components since it increases decoupling.

wjwwood:
I don't think we should do more work in the view to reduce the variables sent from the controller. If we send just the `time.time()` timestamp we need to create the other (`time.localtime()`) with it in each of the views.

I'm not convinced this pr needs to be changed, but feel free to add more commits if you want.

dirk-thomas:
I have added 9328cfe0b8b3988a5e0ea4625ab3bdef9ff2edc4 to this PR. @wjwwood Does that look ok to you?

wjwwood:
👍

PR #302, opened 2016-05-27
Remove hard coded os_name ubuntu in devel jobs

tfoote:
This allows prerelease jobs to run on debian jessie. And theoretically we could turn on devel jobs on debian jessie as well now too.


dirk-thomas:
lgtm

PR #310, opened 2016-07-14
fix the build dependencies to include recursive run dependencies from packages within the workspace

dirk-thomas:
Reported in https://github.com/ros-simulation/gazebo_ros_pkgs/pull/473#issuecomment-232178189

I ran a test job with the branch of the pull request and `gazebo` was added as a build dependency as desired.

### DIFF
--- a/scripts/devel/create_devel_task_generator.py
+++ b/scripts/devel/create_devel_task_generator.py
@@ -94,7 +94,7 @@ def main(argv=sys.argv[1:]):
 
     # get build dependencies and map them to binary packages
     build_depends = get_dependencies(
-        pkgs.values(), 'build', _get_build_dependencies)
+        pkgs.values(), 'build', _get_build_and_recursive_run_dependencies)
     debian_pkg_names_building = resolve_names(build_depends, **context)
     debian_pkg_names_building -= set(debian_pkg_names)
     debian_pkg_names += order_dependencies(debian_pkg_names_building)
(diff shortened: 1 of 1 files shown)


scripts/devel/create_devel_task_generator.py
```diff
@@ -161,11 +161,33 @@ def get_dependencies(pkgs, label, get_dependencies_callback):
...
+    depends = pkg.build_depends + pkg.buildtool_depends
+    # include recursive run dependencies on other pkgs in the workspace
+    # if pkg A in the workspace build depends on pkg B in the workspace
+    # then the recursive run dependencies of pkg B need to be installed
+    # in order to build the workspace
+    other_pkgs_by_names = dict([(p.name, p) for p in pkgs if p.name != pkg.name])
+    run_depends_in_pkgs = set([d.name for d in depends if d.name in other_pkgs_by_names])
+    while run_depends_in_pkgs:
+        # pick first element from sorted order to ensure deterministic results
+        pkg_name = sorted(run_depends_in_pkgs).pop(0)
```
tfoote: This would be more efficient to call 

```
run_depends_in_pkgs.sort()
run_depends_in_pkgs.pop(0)
```

And then you can get rid of `run_depends_in_pkgs.remove(pkg_name)`

Sorting an already sorted list is more efficient and this will avoid the copy as well.

tfoote:
One possible optimization but otherwise looks good to me.

Seeing this in the script kind of makes me wonder if there might be a more centralized location for this sort of logic that might be shareable between the different scripts. But it is a kind of application specific sorting so might not be worth trying to centralize.

scripts/devel/create_devel_task_generator.py
dirk-thomas: The variable is currently a `set` in order to avoid duplicates but a set can't be sorted. I am not sure if converting the data structure to a list (which can be sorted) makes sense since then it will require manual logic to deduplicate.

tfoote: Nevermind, I was thinking it was a list. The logic to deduplicate is definitely worth avoiding at the cost of a sort/copy.

PR #315, opened 2016-08-08
disable sourcedeb after 5 consecutive failures

### LINKED ISSUE #306: automatically disabling jobs such as sourcedebs could avoid a lot of unnecessary emails
We sent several thousand emails to @carlos3dx this weekend because we didn't notice the bad release https://github.com/ros/rosdistro/pull/11989 for most of the weekend.

This plugin could be configured to disable a job after N or so failed builds in a row. 
https://wiki.jenkins-ci.org/display/JENKINS/Disable+Failed+Job+Plugin

Each night the reconfigure would bring it back online, but at least it would only cause N emails per package instead of 84 or so per package. I'd venture that an N around 5 would be reasonable.

dirk-thomas:
Fixes #306.


tfoote:
+1

PR #316, opened 2016-08-09
make the queue path configurable

### LINKED ISSUE #298: import_package_job has hard coded repo path
https://github.com/ros-infrastructure/ros_buildfarm/blob/master/ros_buildfarm/templates/release/import_package_job.xml.em#L49

This should be parameterized and passed in.

dirk-thomas:
Fixes #298.


tfoote:
+1

PR #317, opened 2016-08-09
add more default labels

dirk-thomas:
Replaces #68.

Since each build file can already override the label onyl in the case where the default is being used additional labels are being used. Beside the default (`buildslave`) another label is being added which is created from the following three parts:

* rosdistro name
* build type (`devel`, `doc`, `binarydeb`, `sourcedeb`)
* name of the build file

Since the user can choose the level of granularity of the build files he can therefore influence the granularity of the default labels.

### DIFF
--- a/ros_buildfarm/common.py
+++ b/ros_buildfarm/common.py
@@ -444,3 +444,18 @@ def topological_order_packages(packages):
         if pkg_path is None:
             raise RuntimeError('Circular dependency in: %s' % pkg)
     return ordered_pkg_tuples
+
+
+def get_node_label(config_job_label, default_label=None):
+    if config_job_label is not None:
+        return config_job_label
+    if default_label is None:
+        default_label = get_default_node_label()
+    return default_label
+
+
+def get_default_node_label(additional_label=None):
+    label = 'buildslave'
+    if additional_label:
+        label += ' ' + additional_label
+    return label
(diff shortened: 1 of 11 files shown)


ros_buildfarm/common.py
```diff
@@ -444,3 +444,14 @@ def topological_order_packages(packages):
...
     return ordered_pkg_tuples
+
+
+def get_node_label(
+    config_job_label, rosdistro_name=None, job_type=None, build_name=None
+):
+    if config_job_label is not None:
+        return config_job_label
+    label = 'buildslave'
+    if rosdistro_name and job_type and build_name:
```
tfoote: Why do you support all but not some of the optional arguments? It doesn't make much sense to me to return the default label if 2 of the 3 optional arguments are set.

dirk-thomas: Only when all the arguments are provided for the second label it is being added.

tfoote: I guess it makes sense not to have a code path we're not using. However looking forward it seems like it would be cleaner to support one, two, or three optional arguments, so that in the future when one of those use cases comes up we don't need to worry about someone relying on the all or nothing behavior encoded here.

dirk-thomas:
I modified the API of `get_node_label` to an argument `default_label` and added a function `get_default_node_label`. This should avoid the discussion about partial labels since each caller has full control about how the additional default labels are constructed.

tfoote:
+1

dirk-thomas:
Fix regression in: 2205ad2a81f07204c02a995b9636ea4500b3d0eb

dirk-thomas:
The sourcedeb labels should contain the build file name since they might be shared by multiple. Instead of setting multiple labels with different build file name suffixes I just stripped it: ff92d25f063314acc6d0d1aaf5953ce5cfdad0d5

PR #318, opened 2016-08-09
Refactor wrapper scripts

### LINKED ISSUE #301: get_wrapper_scripts is not robust to being run on an installed version of ros_buildfarm
I used a version of ros_buildfarm installed in a virtual env to reconfigure the job and it caused a regression of the doc independent jobs. 

The method is here: https://github.com/ros-infrastructure/ros_buildfarm/blob/master/ros_buildfarm/templates/__init__.py#L149

The delta in the config is here: http://build.ros.org:8080/job/doc_independent-packages/jobConfigHistory/showDiffFiles?timestamp1=2016-04-20_11-47-03&timestamp2=2016-05-23_16-23-03

It replaced
```
echo "#!/usr/bin/env python3\n\nimport subprocess\nimport sys\nfrom time import sleep\n\n\ndef main(argv=sys.argv[1:]):\n    max_tries = 10\n    known_error_strings = [\n        'Connection timed out',\n    ]\n\n    command = argv[0]\n    if command == 'clone':\n        rc, _, _ = call_git_repeatedly(\n            argv, known_error_strings, max_tries)\n        return rc\n    else:\n        assert \"Command '%s' not implemented\" % command\n\n\ndef call_git_repeatedly(argv, known_error_strings, max_tries):\n    command = argv[0]\n    for i in range(1, max_tries + 1):\n        if i > 1:\n            sleep_time = 5 + 2 * i\n            print(\"Reinvoke 'git %s' (%d/%d) after sleeping %s seconds\" %\n                  (command, i, max_tries, sleep_time))\n            sleep(sleep_time)\n        rc, known_error_conditions = call_git(argv, known_error_strings)\n        if rc == 0 or not known_error_conditions:\n            break\n        print('')\n        print('Invocation failed due to the following known error conditions: '\n              ', '.join(known_error_conditions))\n        print('')\n        # retry in case of failure with known error condition\n    return rc, known_error_conditions, i\n\n\ndef call_git(argv, known_error_strings):\n    known_error_conditions = []\n\n    cmd = ['git'] + argv\n    print(\"Invoking '%s'\" % ' '.join(cmd))\n    proc = subprocess.Popen(\n        cmd, stdout=subprocess.PIPE, stderr=subprocess.STDOUT)\n    while True:\n        line = proc.stdout.readline()\n        if not line:\n            break\n        line = line.decode()\n        sys.stdout.write(line)\n        for known_error_string in known_error_strings:\n            if known_error_string in line:\n                if known_error_string not in known_error_conditions:\n                    known_error_conditions.append(known_error_string)\n    proc.wait()\n    rc = proc.returncode\n    return rc, known_error_conditions\n\n\nif __name__ == '__main__':\n    sys.exit(main())" > wrapper_scripts/git.py
```

with 

```
echo "#!/home/tfoote/work/dbf/deploy_venv/bin/python\n# EASY-INSTALL-SCRIPT: 'ros-buildfarm===1.1.1-master','git.py'\n__requires__ = 'ros-buildfarm===1.1.1-master'\n__import__('pkg_resources').run_script('ros-buildfarm===1.1.1-master', 'git.py')" > wrapper_scripts/git.py
```

These wrapper scripts should probably be stored using pkg_resources instead of being in the scripts directory.

dirk-thomas:
Fixes #301.


ros_buildfarm/wrapper/apt_get.py
```diff
@@ -0,0 +1,142 @@
+#!/usr/bin/env python3
+
```
tfoote: This and below should get a license header.

dirk-thomas: None of the other source files in this repo has any so I wouldn't add it for these two files which have only been moved in this PR.

tfoote:
That's a nice solution. One comment about the license headers, otherwise +1

It looks like we should actually go through and fill in licenses in general.

ros_buildfarm/wrapper/apt_get.py
tfoote: Well we should add it to all the files. I just noticed for these files because I saw the full diff. We can merge this as is I've opened a new ticket #320

dirk-thomas:
Fix regression in e8f01f734f053de2cceb2575d62dd1ba45757ae5.

PR #319, opened 2016-08-09
Improve devel job ssh support

### LINKED ISSUE #314: dev jobs ignoring ssh authentication
Hi,
I try to fix some final things before finally doing a PR for making the buildfarm work with private GH repos. The src and release jobs take care of the git_ssh_credential_id (given in the buildfarm config index.yaml) and set the 'SSH Agent Credentials' in the jenkins job configuration correctly. The dev jobs on the other hand ignore the setting and one needs to manually fix every job entry. Anyone has a hint where/what to fix so that the dev jobs can be configured the same way as the other jobs?
Best,
 Manuel

tfoote:
This should add the ssh-agent and pass through the ssh known hosts into docker.

fixes #314


dirk-thomas:
lgtm

PR #322, opened 2016-08-10
only require a source entry for the target repo

### LINKED ISSUE #92: Documentation: note that external apt repositories used by buildfarm must have source packages enabled
Apologies for the slightly imprecise title. I tried building a private ROS package that depended on a system .deb package only available in one of my internal apt repositories.

I added that apt repository to the `<distribution>`/release-build.yaml file in the ros_buildfarm_config fork, then noticed that the apt-get update was failing during the binary job.

It turned out that I had to go into the reprepro configuration for my internal apt repository and add "source" to the list of architectures in the conf/distributions file.

I realised when I looked at templates/snippet/add_distribution_repositories.Dockerfile.em in the ros_buildfarm repository that the "deb-src" entry is always added, irrespective of whether or not the given apt repository is set up for source packages. 

I don't think anything is actually broken here, but it would be helpful to have a note on this in the documentation, particularly for buildbot_ros users who aren't expecting extra apt repositories to be added to the sources.list with a deb-src entry. 

For reference, this is the error message given by apt-get update in the binary job:
```
Unable to find expected entry 'main/source/Sources' in Release file
```

dirk-thomas:
Fixes #92.

Instead of requiring source entries for all repos this patch only uses a source entry for the target repo since that is the only repo we ever fetch a sourcedeb from (which we have pushed there before).

### DIFF
--- a/ros_buildfarm/argument.py
+++ b/ros_buildfarm/argument.py
@@ -154,6 +154,12 @@ def add_argument_distribution_repository_key_files(parser):
              'corresponding URLs')
 
 
+def add_argument_target_repository(parser):
+    parser.add_argument(
+        '--target-repository',
+        help='The target repository where generated packages are pushed to')
+
+
 def add_argument_custom_rosdep_urls(parser):
     parser.add_argument(
         '--custom-rosdep-urls',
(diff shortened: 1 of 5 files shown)


tfoote:
+1

PR #324, opened 2016-08-16
call apt clean after apt install

dirk-thomas:
Deploy on idle farm and reconfigure all jobs.


tfoote:
+1

dirk-thomas:
`apt clean` is not available on older Ubuntu distros: 7956fb2cb38d76e85cae759dd23657f6be30adcf

dirk-thomas:
Until we stop supporting Saucy we need to stick with `apt-get`: 7bb36cc6378c8effefca9a7cdf048027225eb664 76cbdc8a456cbff104ecf887d88c23dece07fc9f

PR #327, opened 2016-08-18
allow status page jobs to run on master as well as slave_on_master

tfoote:
It's common when long running reconfigure jobs are going that the status pages cannot be updated since they are sharing the same executor. 

Reviewing the jobs there's no reason that the status page jobs cannot run in parallel. They upload the same resources into the same target directory, but they upload the exact same files and which ever one writes last will be fine. 

This leaves the reconfigure jobs tied to a single executor as well as the release trigger jobs which are good to wait for the reconfigure jobs to finish.

### DIFF
--- a/ros_buildfarm/templates/status/release_compare_page_job.xml.em
+++ b/ros_buildfarm/templates/status/release_compare_page_job.xml.em
@@ -24,7 +24,7 @@
     refspec=None,
 ))@
   <scmCheckoutRetryCount>2</scmCheckoutRetryCount>
-  <assignedNode>slave_on_master</assignedNode>
+  <assignedNode>master||slave_on_master</assignedNode>
   <canRoam>false</canRoam>
   <disabled>false</disabled>
   <blockBuildWhenDownstreamBuilding>false</blockBuildWhenDownstreamBuilding>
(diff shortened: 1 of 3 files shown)


dirk-thomas:
See https://github.com/ros-infrastructure/buildfarm_deployment/pull/133#issuecomment-240616430 for why I don't think the jobs on master should run in parallel. And based on that the status pages must not be run on the same executor. Therefore -1 to let them run on the master.

tfoote:
So, looking a little closer at the status jobs they could be allowed to run on the executors as well. Maybe instead of adding the label `master` we could add `buildslave`. I validated that they work sucessfully on a generic buildslave: http://build.ros.org/job/Jrel_arm_release-status-page/20769/

This would allow one to run on the slave_on_master, but many could run on the executors if they're available. And if they're backlogged it will be no slower than running them single threaded on the master. 

Note to do this we will want to resolve #331 first.

dirk-thomas:
If you simply want the status pages to be able to run in parallel to potentially long running reconfigure jobs without blocking the very important rosdistro cache jobs there is another way to achieve that: create another executor on the master with a specific label and assign that label to all status page jobs.

tfoote:
It would work to add another executor with a different label, but that will make the deployment significantly more complicated as it will add yet another jenkins slave on the master machine. I'd really like to remove the `slave_on_master` executor already as it one of the biggest complications to deploying the buildfarm on a single machine already.

dirk-thomas:
See https://github.com/ros-infrastructure/ros_buildfarm/issues/331#issuecomment-251174703

PR #328, opened 2016-08-18
Allow token urls

ayrton04:
For some background, the issue is that our Jenkins PR check jobs were not firing. Looking at the logs in Jenkins, I found these lines:

`INFO: GitHub project property is missing the URL, cannot start ghprb trigger for job Ipr__robot_localization__ubuntu_trusty_amd64`

Looking around online and looking at the manual configuration for the job, I noticed this:

image

By checking the "GitHub project" box and manually filling out the project URL, the jobs were successfully triggered on subsequent PRs. I tracked the issue down to the fact that the `get_github_project_url` was returning None. This was due to the fact that we were using URLs in our rosdistro that started with OAuth tokens, as we need to clone private repos, but `get_github_project_url `was assuming that the URLs would start with `https://github.com/`. 

I've addressed the issue by using a regex instead. Please note that my testing of this has only involved manually checking the regex and running these methods in isolation. If there's a way for me to force our build server to point to our `ros_buildfarm` fork instead of the OSRF one, I can do that to verify that it works. Here are the tests I ran for the regex:

```
>>> pattern = re.compile("https:\/\/([a-f0-9]{40}@)?github\.com\/.*")
>>> url = "https://3dc7fb42354ce39edbc4f92b44786313cc53abe5@github.com/"
>>> print pattern.match(url)  # Patten.match returns a Match object if the pattern is found 
<_sre.SRE_Match object at 0x7f2272390eb8>
>>> url = "https://3dc7fb42354ce39edbc4f92b44786313cc53abe5@github.com"
>>> print pattern.match(url)  # Pattern.match returns None if no pattern is found
None
>>> url = "https://3dc7fb42354ce39edbc4f92b44786313cc53abe5@github.com/whatevER92994/aa"
>>> print pattern.match(url)
<_sre.SRE_Match object at 0x7f22722fa120>
>>> url = "https://3dc7fb42354ce39edbc4f92b44786313cc53abe5aaaaa@github.com/"
>>> print pattern.match(url)
None
>>> url = "https://3dc7fb42354ce39edbc4f92b44786313cc53abe@github.com/"
>>> print pattern.match(url)
None
>>> url = "https://3dc7fb42354ce39edbc4f92b44786313cc53abe5@github.com/"
>>> print pattern.match(url)
<_sre.SRE_Match object at 0x7f2272390eb8>
>>> url = "https://3dc7fb42354ce39edbc4f92b44786313cc53abE5@github.com/"
>>> print pattern.match(url)
None
>>> url = "https://3dc7fb42354ce39edbc4f92b44786313cc53abe5@github.com/"
>>> print pattern.match(url)
<_sre.SRE_Match object at 0x7f2272390eb8>
>>> url = "https://3dc7fb42354ce39edbc4f92b44786313cc53abg5@github.com/"
>>> print pattern.match(url)
None
>>> url = "https:/3dc7fb42354ce39edbc4f92b44786313cc53abf5@github.com/"
>>> print pattern.match(url)
None
>>> url = "http://3dc7fb42354ce39edbc4f92b44786313cc53abf5@github.com/"
>>> print pattern.match(url)
None
>>> url = "https://3dc7fb42354ce39edbc4f92b44786313cc53abf5github.com/"
>>> print pattern.match(url)
None
>>> url = "https://3dc7fb42354ce39edbc4f92b44786313cc53abf5@github.com/"
>>> print pattern.match(url)
<_sre.SRE_Match object at 0x7f2272390eb8>
>>> url = "https://github.com/"
>>> print pattern.match(url)
<_sre.SRE_Match object at 0x7f2272390eb8>
>>> url = "https://github.com"
>>> print pattern.match(url)
None
>>> url = "http://github.com/"
>>> print pattern.match(url)
None
>>> url = "https://@github.com"
>>> print pattern.match(url)
None
>>> url = "https://@github.com/"
>>> print pattern.match(url)
None
>>> url = "https://github.com/"
>>> print pattern.match(url)
<_sre.SRE_Match object at 0x7f22722fa120>
```

### DIFF
--- a/ros_buildfarm/common.py
+++ b/ros_buildfarm/common.py
@@ -278,7 +278,9 @@ def get_short_arch(arch):
 def git_github_orgunit(url):
     if github_url_pattern.match(url) is None:
         return None
-    path = url[len(prefix):]
+    git_domain = 'github.com/'
+    path_index = url.find(git_domain)
+    path = url[path_index + len(git_domain):]
     index = path.index('/')
     return path[:index]
 
--- a/ros_buildfarm/common.py
+++ b/ros_buildfarm/common.py
@@ -44,6 +44,7 @@ class Scope(object):
 
 
 Target = namedtuple('Target', 'os_name os_code_name arch')
+github_url_pattern = re.compile("https:\/\/([a-f0-9]{40}@)?github\.com\/.*")
 
 
 def get_repositories_and_script_generating_key_files(
@@ -275,9 +276,7 @@ def get_short_arch(arch):
 
 
 def git_github_orgunit(url):
-    pattern = re.compile("https:\/\/([a-f0-9]{40}@)?github\.com\/.*")
-
-    if pattern.match(url) is None:
+    if github_url_pattern.match(url) is None:
         return None
     path = url[len(prefix):]
     index = path.index('/')
@@ -285,9 +284,7 @@ def git_github_orgunit(url):
 
 
 def get_github_project_url(url):
-    pattern = re.compile("https:\/\/([a-f0-9]{40}@)?github\.com\/.*")
-
-    if pattern.match(url) is None:
+    if github_url_pattern.match(url) is None:
         return None
     git_suffix = '.git'
     if not url.endswith(git_suffix):
--- a/ros_buildfarm/common.py
+++ b/ros_buildfarm/common.py
@@ -14,7 +14,7 @@
 
 from collections import namedtuple
 import os
-
+import re
 
 class JobValidationError(Exception):
     """
@@ -274,8 +274,10 @@ def get_short_arch(arch):
 
 
 def git_github_orgunit(url):
-    prefix = 'https://github.com/'
-    if not url.startswith(prefix):
+    pattern = re.compile("https:\/\/([a-z0-9]{40}@)?github\.com\/.*")
+    pattern.match(url)
+
+    if pattern.match(url) is None:
         return None
     path = url[len(prefix):]
     index = path.index('/')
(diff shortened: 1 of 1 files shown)


dirk-thomas:
Thanks for the patch. I am not sure if using a regex for this task and then afterward searching for the github domain is the best thing to do. I would propose to use `urlparse` instead. Please have a look at 6cda6ecf930fd393ec8e0a895170a3706e5167d4 for an alternative patch. That patch also accepts any kind of auth token.

That being said I am not sure if embedding the auth token into the url is a good idea. This information might be visible in several locations. Maybe you should consider a different approach?

Btw. I hope the auth token you posted above is just an example and not valid otherwise you might want to revoke it to prevent abuse.

ayrton04:
Yeah, I made a token up for that example, no worries.

I got the auth token idea from Danial Di Marco's [ROSCon talk](https://vimeo.com/142151194) at around 15:45. Our rosdistro Git server is private, and we've locked down access to our build server, so we should be set. That being said, if there's a better approach for accessing private repos, I'd be happy to give it a shot.

Your patch is much better, agreed. :+1:

dirk-thomas:
Interesting, it seems like they are already using it like that so it seems they must have a similar local patch. It would be nice if they would contribute it back :worried:

Anyway I just created #329 with my modified patch so I will close this. Thank you for creating the issue and helping to make the build farm useful in more scenarios.

tfoote:
I believe the Bosch team worked around it by having the read access open but only available via a VPN, and they only needed the token based access for the rosdistro and other write access

PR #330, opened 2016-08-18
don't use shallow cloning when using merge-before-build

dirk-thomas:
Based on the discussion in http://discourse.ros.org/t/clone-depth-for-pr-devel-jobs/419 about failing PR jobs.

Consider the following scenario:
* a job uses a shallow clone as well as "merge before build" (which currently all PR jobs do)
* the branch is not built on top of the latest master
* a newer version of the Jenkins Git plugin actually applies the depth option for the shallow clone correctly

The "merge before build" fails since it doesn't have enough history to determine how to perform the merge. Since we can't set a depth which will work in all cases this patch disables the use of shallow clone for the case of PR jobs. Other jobs which don't use "build before merge" can continue to use shallow clones.

### DIFF
--- a/ros_buildfarm/templates/devel/devel_job.xml.em
+++ b/ros_buildfarm/templates/devel/devel_job.xml.em
@@ -55,8 +55,7 @@ if pull_request:
     path='catkin_workspace/src/%s' % source_repo_spec.name,
     git_ssh_credential_id=git_ssh_credential_id,
 ))@
-@[end if]@
-@[if pull_request]@
+@[else]@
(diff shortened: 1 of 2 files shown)


tfoote:
+1

Might we want to keep the block though and disable the shallow clone conditionally. And while we're here stop pulling tags since we're not using them?

dirk-thomas:
`devel` jobs might use tags which we can't check for. I am not sure if the no-tags option for PR jobs is worth it. I don't expect it to save noticeable data.

PR #334, opened 2016-08-24
run single apt call for folded dependencies

dirk-thomas:
Before the folded commands were just concatenated. Now the apt calls are being merged into a single invocation.


tfoote:
+1

I tested a little bit and the install targets only cost a few second extra to call sequentially, but the main advantage of this is not calling update between the installs on the same line. 

It would be a little better for debugging errors if they were separate, but with the `pkgProblemResolver` enabled this should be ok to debug. And we really shouldn't be getting to this point if dependencies are broken.

PR #336, opened 2016-08-25
change status page threshold

dirk-thomas:
The status page jobs run every 15 minutes. With the current threshold being identical the page is already marked in orange if the job finishes a second slower than the previous run. Therefore this patch adds some buffer to the threshold.


tfoote:
+1

PR #338, opened 2016-08-26
Updating diffutils URL

ayrton04:
Our reconfigure jobs were failing. It looks like Google moved the location of the diffutils-1.2.1.jar file. Verified that `wget` succeeds with the new URL.


dirk-thomas:
Addresses #339. Please see comment there.

PR #340, opened 2016-08-30
use upstream CrumbRequester if available

### LINKED ISSUE #333: use CrumbRequester from upstream
Upstream has integrated the `CrumbRequester` class in https://github.com/pycontribs/jenkinsapi/commit/06e03d845edf9076dfe561390b0a67593b5518fb. Once there is a new release the class in the repo should be removed and upstream being used instead.

dirk-thomas:
Fixes #333.


tfoote:
+1

PR #342, opened 2016-09-26
fix newline expansion for some shells

dirk-thomas:
If the `sh` shell actually some custom shell like `bash` the `echo` command doesn't expand the `\n` to newline but outputs them as-is. 

Fixes ros/rosdistro#12824.


dirk-thomas:
Merging since it has been reported to address the problem: https://github.com/ros/rosdistro/issues/12824#issuecomment-249709548

PR #343, opened 2016-09-28
always use constructed base image name

dirk-thomas:
Support native arm64 jobs on a TX1.

Example builds:
* Devel http://build.ros.org/view/Kdev_uxv8/job/Kdev_uxv8__rospack__ubuntu_xenial_arm64/12/
* Binarydeb http://build.ros.org/view/Kbin_uxv8_uXv8/job/Kbin_uxv8_uXv8__desktop_full__ubuntu_xenial_arm64__binary/7/

### DIFF
--- a/ros_buildfarm/common.py
+++ b/ros_buildfarm/common.py
@@ -14,6 +14,7 @@
 
 from collections import namedtuple
 import os
+import platform
 try:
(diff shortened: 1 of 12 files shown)


dirk-thomas:
Ready for review.

PR #355, opened 2016-11-29
install EmPy from PyPI

dirk-thomas:


tfoote:
That's great! Are there other places we could make this switch?

dirk-thomas:
I am not aware of any other places.

PR #357, opened 2016-12-05
use utf-8 instead of unicode

mikaelarguedas:
`'unicode'` is not a known encoding type for python2

### DIFF
--- a/ros_buildfarm/jenkins.py
+++ b/ros_buildfarm/jenkins.py
@@ -239,8 +239,9 @@ def _diff_configs(remote_config, new_config):
     if ElementTree.tostring(remote_root) == ElementTree.tostring(new_root):
         return []
 
-    xml1 = ElementTree.tostring(remote_root, encoding='utf-8')
-    xml2 = ElementTree.tostring(new_root, encoding='utf-8')
+    encoding_type = 'utf-8' if sys.version_info[0] < 3 else 'unicode'
+    xml1 = ElementTree.tostring(remote_root, encoding=encoding_type)
+    xml2 = ElementTree.tostring(new_root, encoding=encoding_type)
     lines1 = xml1.splitlines()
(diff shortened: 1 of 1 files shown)


dirk-thomas:
This patch fails with Python 3.

mikaelarguedas:
The reason for this patch initially was because when running `generate_all_jobs.py` using the python2 version `python-rosbuildfarm` the command failed when trying to show the diff between the job on jenkins and the newly generated job configuration.

simple reproducible example:
```python
from xml.etree import ElementTree
new_config = """
<triggers>
  <hudson.triggers.TimerTrigger>
    <spec>0 11 * * *</spec>
  </hudson.triggers.TimerTrigger>
</triggers>
"""
new_root = ElementTree.fromstring(new_config)
xml1 = ElementTree.tostring(new_root, encoding='unicode')
```
and the resulting error message:
```Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/lib/python2.7/xml/etree/ElementTree.py", line 1126, in tostring
    ElementTree(element).write(file, encoding, method=method)
  File "/usr/lib/python2.7/xml/etree/ElementTree.py", line 817, in write
    self._root, encoding, default_namespace
  File "/usr/lib/python2.7/xml/etree/ElementTree.py", line 884, in _namespaces
    add_qname(tag)
  File "/usr/lib/python2.7/xml/etree/ElementTree.py", line 868, in add_qname
    qnames[qname] = encode(qname)
  File "/usr/lib/python2.7/xml/etree/ElementTree.py", line 843, in encode
    return text.encode(encoding)
LookupError: unknown encoding: unicode
```

https://github.com/ros-infrastructure/ros_buildfarm/pull/357/commits/0019797bee53d5ac03af8eff4c89a9d058437d81 seems to fix it

dirk-thomas:
Thanks!

PR #358, opened 2016-12-19
[prerelease] Add arg example for --custom-repo option.

130s:
Argument value for `--custom-repo` option of `generate_prerelease_script.py` isn't very clear to me so adding a concrete example.
I'm not sure if the example here is correct, but on my computer it seems the script is able to fetch the specified branch and commit.

### DIFF
--- a/scripts/prerelease/generate_prerelease_script.py
+++ b/scripts/prerelease/generate_prerelease_script.py
@@ -79,7 +79,8 @@ def main(argv=sys.argv[1:]):
         type=_repository_name_and_type_and_url_and_branch,
         default=[],
         metavar='REPO_NAME:REPO_TYPE:REPO_URL:BRANCH_OR_TAG_NAME',
-        help='The name, type, url and branch / tag name of a repository')
+        help='The name, type, url and branch / tag name of a repository, '
+             'e.g. "common_tutorials:git:https://github.com/ros/common_tutorials:pullrequest-1"')
 
     add_overlay_arguments(parser)
 


scripts/prerelease/generate_prerelease_script.py
```diff
@@ -79,7 +79,10 @@ def main(argv=sys.argv[1:]):
         type=_repository_name_and_type_and_url_and_branch,
         default=[],
         metavar='REPO_NAME:REPO_TYPE:REPO_URL:BRANCH_OR_TAG_NAME',
-        help='The name, type, url and branch / tag name of a repository')
+        help='The name, type, url and branch / tag name of a repository. ' +
+             'e.g. For forked repo https://github.com/ros/common_tutorials ' +
+             'with a branch called pullrequest-1, it will be ' +
+             'common_tutorials:git:https://github.com/ros/common_tutorials:pullrequest-1')
```
130s: Agreed. I updated as you suggested.

scripts/prerelease/generate_prerelease_script.py
```diff
@@ -79,7 +79,8 @@ def main(argv=sys.argv[1:]):
         type=_repository_name_and_type_and_url_and_branch,
         default=[],
         metavar='REPO_NAME:REPO_TYPE:REPO_URL:BRANCH_OR_TAG_NAME',
-        help='The name, type, url and branch / tag name of a repository')
+        help='The name, type, url and branch / tag name of a repository.' +
+             'e.g. "common_tutorials:git:https://github.com/ros/common_tutorials:pullrequest-1"')
```
dirk-thomas: There is currently no space between the two strings. The first one also ends with `.` but the second starts with lower case.

I would suggest to replace the trailing `.` with `, `.

130s: Updated.
And `+` removed as suggested at https://github.com/ros-infrastructure/ros_buildfarm/pull/358#discussion_r93327818.

PR #360, opened 2017-01-11
added lunar targets

mikaelarguedas:
I assumed it's ok to have duplicate `os_code_mapping` values because they're always associated with the `os_short_name` so it shouldn't be ambiguous. Is that correct ?


dirk-thomas:
On the status pages the tooltip of the column will show the long string so that should be ok. The only problem would be if a single ROS distro would span across both `Saucy` and `Stretch` since the Jenkins views would get merged. But that is not the case so I think this is good as is.

PR #363, opened 2017-01-20
add multiverse to source.list for amd64

k-okada:
I have trouble on buidling package at http://build.ros.org/job/Isrc_uT__darknet__ubuntu_trusty__source/,
whcih it can compile on i386, but not on amd64. The problem is it fail to find `nvidia-cuda-toolkit` as

```
23:13:21 Invoking 'debchange -v 2016.11.27-1trusty-20170119-231321-0800 -p -D trusty -u high -m Append timestamp when binarydeb was built.' in '/tmp/binarydeb/ros-indigo-darknet-2016.11.27'
23:13:21 # END SUBSECTION
23:13:24 Looking for the '.dsc' file of package 'ros-indigo-darknet' with version '2016.11.27-1'
23:13:27 Traceback (most recent call last):
23:13:27   File "/usr/lib/python3/dist-packages/apt/cache.py", line 198, in __getitem__
23:13:27     return self._weakref[key]
23:13:27   File "/usr/lib/python3.4/weakref.py", line 125, in __getitem__
23:13:27     o = self.data[key]()
23:13:27 KeyError: 'nvidia-cuda-toolkit'
23:13:27 
23:13:27 During handling of the above exception, another exception occurred:
23:13:27 
23:13:27 Traceback (most recent call last):
23:13:27   File "/tmp/ros_buildfarm/scripts/release/create_binarydeb_task_generator.py", line 162, in <module>
23:13:27     main()
23:13:27   File "/tmp/ros_buildfarm/scripts/release/create_binarydeb_task_generator.py", line 84, in main
23:13:27     apt_cache, debian_pkg_names)
23:13:27   File "/tmp/ros_buildfarm/ros_buildfarm/common.py", line 134, in get_binary_package_versions
23:13:27     pkg = apt_cache[debian_pkg_name]
23:13:27   File "/usr/lib/python3/dist-packages/apt/cache.py", line 205, in __getitem__
23:13:27     raise KeyError('The cache has no package named %r' % key)
23:13:27 KeyError: "The cache has no package named 'nvidia-cuda-toolkit'"
23:13:27 Build step 'Execute shell' marked build as failure
23:13:27 [ssh-agent] Stopped.
23:13:27 [description-setter] Could not determine descri
```

What I found is amd64 was build on `ubuntu:trusty` Docker but i386 was using `osrf/ubuntu_i386:trusty`.
[...]

PR #364, opened 2017-01-24
enable additional repo for binary jobs

dirk-thomas:
Replaces #363.

E.g. on Ubuntu enable multiverse.


dirk-thomas:
@tfoote You could remove `multiverse` from the custom `armhf` Docker images (to be consistent with the other Docker images). Currently the armhf jobs will have an output that the repo is in there twice but continue without a problem.

Issue #233, opened 2016-03-07
sourcedeb syncs occasionally fail


tfoote:
This leaves the sourcedebs empty on the testing repo. 

The problem is that when syncing the sourcedebs we first clear them all, and re-add them all. However if one fails to import the whole import stops leaving the repo empty. 

The error occurs when the orig.tar.gz is inconsistent. This happens when there's a debian increment and bloom recreates the tarball instead of creating a deb diff patch.  

A failed job can be seen here: http://build.ros.org/view/Ksrc_uW/job/Krel_sync-packages-to-testing_wily_amd64/4/console 
```
19:23:50 Creating distributions file /var/repos/ubuntu/testing/conf/distributions
19:23:50 running command ['reprepro', '-v', '-b', '/var/repos/ubuntu/testing', '--noskipold', 'update', 'wily']
19:23:50 Warning: No VerifyRelease line in 'ros-kinetic-wily-source' or any rule it includes via 'From:'.
19:23:50 Release.gpg cannot be checked unless you tell which key to check with.
19:23:50 (To avoid this warning and not check signatures add 'VerifyRelease: blindtrust').
19:23:50 aptmethod got 'file:///var/repos/ubuntu/building/dists/wily/InRelease'
19:23:50 aptmethod got 'file:///var/repos/ubuntu/building/dists/wily/main/source/Sources.gz'
19:23:50 File "pool/main/r/ros-kinetic-catkin/ros-kinetic-catkin_0.7.0.orig.tar.gz" is already registered with different checksums!
19:23:50 md5 expected: 93770d0ded61ab813df8ac629190e740, got: 6a43f7d9ee21dfd5046ab75fd117ff01
19:23:50 sha1 expected: 938f96dd5a5fe6208ca4ddf4c71fd51689693dd5, got: c7242b70b483212ca7604653e460b5e90727e0b7
19:23:50 sha256 expected: 4d16cd79f1c1a847ea0428ce6ea6b7a1f860274a91beba4510e7a75f9baa3f56, got: c87cbb94eacd59e99e32a262fd193d8f3210f1faaa6a855f238d6e50844e719b
19:23:50 size expected: 197515, got: 197531
19:23:50 There have been errors!
```

However once the xenial job removed the other references to orig.tar.gz http://build.ros.org/job/Krel_sync-packages-to-testing_xenial_amd64/3/console

```
19:25:20 removing 'ros-kinetic-catkin' from 'xenial|main|source'...
[...]

PR #374, opened 2017-02-16
Orig tarball

tfoote:
First commit worked here: http://build.ros.org/view/Lsrc_uZ/job/Lsrc_uZ__catkin__ubuntu_zesty__source/5/console

2nd commit adds build file based parameterization and passes here: http://build.ros.org/view/Lsrc_uZ/job/Lsrc_uZ__catkin__ubuntu_zesty__source/7/console and http://build.ros.org/view/Lsrc_uX/job/Lsrc_uX__catkin__ubuntu_xenial__source/3/console

I manually modified the config to use this branch.

I believe that this will fix the issue with debian increments as well.

### DIFF
--- a/ros_buildfarm/sourcedeb_job.py
+++ b/ros_buildfarm/sourcedeb_job.py
@@ -22,7 +22,7 @@ from ros_buildfarm.release_common import dpkg_parsechangelog
 
 def get_sources(
         rosdistro_index_url, rosdistro_name, pkg_name, os_name, os_code_name,
-        sources_dir):
+        sources_dir, debian_repository_urls):
     from rosdistro import get_cached_distribution
     from rosdistro import get_index
     index = get_index(rosdistro_index_url)
@@ -64,17 +64,19 @@ def get_sources(
     origtgz_version = pkg_version.split('-')[0]
     debian_package_name = get_debian_package_name(rosdistro_name, pkg_name)
     filename = '%s_%s.orig.tar.gz' % (debian_package_name, origtgz_version)
-    # TODO(tfoote) remove the hardcoded repo url
-    URL_TEMPLATE='http://repositories.ros.org/ubuntu/building/pool/main/r/%s/%s'
-    url = URL_TEMPLATE % (debian_package_name, filename)
 
-    try:
-        output_file = os.path.join(sources_dir, '..', filename)
-        urlretrieve(url, output_file)
-        print("Fetched %s to %s" % (url, output_file))
-    except Exception as ex:
-        print(ex)
-        print("No tarball found at %s, that's ok it will be rebuilt" % url)
+    for repo in debian_repository_urls:
+        URL_TEMPLATE = '%s/pool/main/r/%s/%s'
+        url = URL_TEMPLATE % (repo, debian_package_name, filename)
+
+        try:
+            output_file = os.path.join(sources_dir, '..', filename)
+            urlretrieve(url, output_file)
+            print("Found matching original tarball, "
+                  "downloading %s to %s" % (url, output_file))
+            continue
+        except:
+            print("No tarball found at %s, that's ok it will be rebuilt" % url)
 
     # output package version for job description
     print("Package '%s' version: %s" % (pkg_name, source_version))
--- a/ros_buildfarm/sourcedeb_job.py
+++ b/ros_buildfarm/sourcedeb_job.py
@@ -77,7 +77,7 @@ def get_sources(
                   (url, output_file))
             break
         except:
-            print("No tarball found at '%s'." % url)
+            print("No tarball found at '%s'" % url)
 
     # output package version for job description
     print("Package '%s' version: %s" % (pkg_name, source_version))
--- a/ros_buildfarm/sourcedeb_job.py
+++ b/ros_buildfarm/sourcedeb_job.py
@@ -66,8 +66,9 @@ def get_sources(
     filename = '%s_%s.orig.tar.gz' % (debian_package_name, origtgz_version)
 
     for repo in debian_repository_urls:
-        URL_TEMPLATE = '%s/pool/main/r/%s/%s'
-        url = URL_TEMPLATE % (repo, debian_package_name, filename)
+        URL_TEMPLATE = '%s/pool/main/%s/%s/%s'
+        prefix = debian_package_name[0]
+        url = URL_TEMPLATE % (repo, prefix, debian_package_name, filename)
 
         try:
             output_file = os.path.join(sources_dir, '..', filename)
--- a/ros_buildfarm/sourcedeb_job.py
+++ b/ros_buildfarm/sourcedeb_job.py
@@ -65,19 +65,20 @@ def get_sources(
     debian_package_name = get_debian_package_name(rosdistro_name, pkg_name)
     filename = '%s_%s.orig.tar.gz' % (debian_package_name, origtgz_version)
 
+    URL_TEMPLATE = '%s/pool/main/%s/%s/%s'
+    prefix = debian_package_name[0]
     for repo in debian_repository_urls:
-        URL_TEMPLATE = '%s/pool/main/%s/%s/%s'
-        prefix = debian_package_name[0]
         url = URL_TEMPLATE % (repo, prefix, debian_package_name, filename)
 
+        output_file = os.path.join(sources_dir, '..', filename)
         try:
-            output_file = os.path.join(sources_dir, '..', filename)
             urlretrieve(url, output_file)
-            print("Found matching original tarball, "
-                  "downloading %s to %s" % (url, output_file))
-            continue
+            break
         except:
-            print("No tarball found at %s, that's ok it will be rebuilt" % url)
+            print("No tarball found at '%s'." % url)
+        else:
+            print("Downloaded original tarball '%s' to '%s'" %
+                  (url, output_file))
 
     # output package version for job description
     print("Package '%s' version: %s" % (pkg_name, source_version))
--- a/ros_buildfarm/sourcedeb_job.py
+++ b/ros_buildfarm/sourcedeb_job.py
@@ -14,6 +14,7 @@
 
 import os
 import subprocess
+from urllib.request import urlretrieve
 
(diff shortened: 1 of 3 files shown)


tfoote:
This should actually fix: #233 and obviate the need for #373 and #375

dirk-thomas:
I don't think this should replace the other PRs. E.g. #375 makes a lot of sense even when this gets applied. Otherwise if Yakkety is the first platform to build the tarball contains current time stamps and can't be recreated without a diff.

Anyway imo this needs extensive testing on the test farm including, first time releases, new patch releases, new debinc releases, etc. (I will use the "more-information-needed" label for this.)

tfoote:
This is a high priority, deploying this should allow us to turn off the EOL'd distros which we've wanted to do for a long time. The most recent thread is [here](https://discourse.ros.org/t/drop-support-for-wily/727) And had planned to do it for even longer. 

After working on manually packaging debian increments I believe that this is the better way to package everything. From the [debian handbook section](https://debian-handbook.info/browse/stable/sect.source-package-structure.html) the orig.tar.gz is not supposed to be modified by the packaging. 

@jspricke @j-rivero Can you confirm that this is a correct approach for building the packages?

For reference we've discovered that xenial, yakkety, and zesty produce different tarballs since they use different options to tar which means they conflict on trying to upload the orig.tar.gz. #373 and #375 workaround a change in standard and a bug respectively. 

I agree #375 is less than desireable behavior. we could leave it in. We'll just have to remove that code path once we stop supporting Yakkety. And it's a very fragile fix if there's ever a change to those lines (such as debian makes a patch release to fix the bug, though quite unlikely), the sed line will fail. And with this fix we will never run into this problem since we won't regenerate the orig.tar.gz files. Completely reproducable builds are valuable, but I'd argue that adding the complexity to our system is unnecessary since it's already fixed upstream and should remain so for future versions.

dirk-thomas:
Regarding #375 it is clearly not a great solution. I am happy if anyone suggests a "cleaner" workaround. We could e.g. create our own version of the Debian package and put it in our repo. But since that involves more effort I went for the workaround I am able to implement within minutes.

Anyway I disagree with your conclusion of this being especially fragile and increases complexity unnecessarily.

* First of all the modification is only performed for Yakkety. It doesn't need to be removed since once Yakkety is not being targeted anymore it just never gets used anymore. It can be cleaned up though (if we know nobody using our code uses it to build for Yakkety anymore).
* The sed line applies the exact upstream patch. Only if the same line is being modified in a new release in Yakkety the sed command would turn into a no-op since the regex would match anymore. The line could be changed for two reasons (both probably very unlikely):

  * The Debian package gets updated to patch the actual bug. Then since the sed line doesn't apply anymore the build continues to work.
  * The Debian package gets updated with a change to this line which doesn't fix the bug. Then we will immediately notice the failing builds and have to update the workaround to continue working.

jspricke:
Hi Tully,

* Tully Foote <notifications@github.com> [2017-02-16 23:15]:
> After working on manually packaging debian increments I believe that this is the better way to package everything. From the [debian handbook section](https://debian-handbook.info/browse/stable/sect.source-package-structure.html) the orig.tar.gz is not supposed to be modified by the packaging. 

For Debian the orig.tar.gz should be the one the upstream project
provides¹, so it shouldn't make a difference if you take the one from
the repo or from upstream². As you are not Debian, I think it's up to
you to decide how to handle this. I agree with Dirk, that gbp is buggy,
and it should be fixed on their side (or worked around as in #375, while
you rely on a broken version). Hopefully this doesn't happen again so
often (otherwise your idea is a viable way to work around these).

Cheers Jochen

¹: Except when there is a good reason to repack it, like non-free
content.
²: Except when upstream changes released versions, where I think this is
a nice check for.

dirk-thomas:
This is currently blocking the Lunar packages for Debian targets. Note that those are actually creating different archives since the newer `dpkg` version adds a new `.buildinfo` file. So it would be non-deterministic which of the two variations of the archive makes it into the repo depending on which job gets processed first.

mikaelarguedas:
@tfoote @j-rivero any input on how to handle the difference between debian and ubuntu for the orig tarballs ? (c.f. https://github.com/ros-infrastructure/ros_buildfarm/pull/380#issue-209213286)

tfoote:
@mikaelarguedas @dirk-thomas the `.buildinfo` doesn't go into the orig.tar.gz that will go into the debian.tar.gz which is layered on top, and will be regenerated any time a sourcedeb job is generated. Thus this is independent of that change and #380 will stand on it's own to fix that.

dirk-thomas:
@tfoote Thanks for clarifying.

tfoote:
I've rebased on the latest. It now builds on Stretch, Xenial, Yakkety, and Zesty for Lunary. 

I ran all 4 of these debian increments in parallel: 
http://build.ros.org:8080/view/Lsrc_dS/job/Lsrc_dS__angles__debian_stretch__source/7/console
http://build.ros.org:8080/view/Lsrc_uX/job/Lsrc_uX__angles__ubuntu_xenial__source/7/console
http://build.ros.org:8080/view/Lsrc_uY/job/Lsrc_uY__angles__ubuntu_yakkety__source/3/console
http://build.ros.org:8080/view/Lsrc_uZ/job/Lsrc_uZ__angles__ubuntu_zesty__source/3/console

And they all passed. 

This job passes: http://build.ros.org:8080/view/Lsrc_uX/job/Lsrc_uX__angles__ubuntu_xenial__source/6/ after the old style Yakkety job changed the tarball underneath this job running: http://build.ros.org:8080/job/Lrel_import-package/164/console
http://build.ros.org:8080/view/Lsrc_uX/job/Lsrc_uX__angles__ubuntu_xenial__source/5/

It now has built both our use cases, a new releases and debian increment releases. 

There is currently one known issue which is that if two are triggered simultaneously on a new release they might both generate the tarball. And if they generate different ones(aka different behaviors between versions of dpkg like #375 ) In which case it will hit our existing retry logic and whichever job uploaded 2nd will need to retry. It will be better than the current situation where we require that all arches run before any individual one can succeed again.

Note that this also fixes the Stretch builds RE: https://github.com/ros-infrastructure/ros_buildfarm_config/pull/75

ros_buildfarm/sourcedeb_job.py
```diff
@@ -59,6 +60,25 @@ def get_sources(
...
+    debian_package_name = get_debian_package_name(rosdistro_name, pkg_name)
+    filename = '%s_%s.orig.tar.gz' % (debian_package_name, origtgz_version)
+
+    for repo in debian_repository_urls:
+        URL_TEMPLATE = '%s/pool/main/%s/%s/%s'
+        prefix = debian_package_name[0]
+        url = URL_TEMPLATE % (repo, prefix, debian_package_name, filename)
+
+        try:
+            output_file = os.path.join(sources_dir, '..', filename)
```
dirk-thomas: Please move this line before the try / except block.

ros_buildfarm/sourcedeb_job.py
```diff
@@ -59,6 +60,25 @@ def get_sources(
...
+            urlretrieve(url, output_file)
+            print("Found matching original tarball, "
+                  "downloading %s to %s" % (url, output_file))
+            continue
```
dirk-thomas: `continue` seems to be unnecessary here. First it does loop anyway. Second should it actually try the other repo urls or `break` instead?

ros_buildfarm/sourcedeb_job.py
```diff
@@ -59,6 +60,25 @@ def get_sources(
...
             (source_version, pkg_version))
 
+    # If a tarball already exists reuse it
+    origtgz_version = pkg_version.split('-')[0]
+    debian_package_name = get_debian_package_name(rosdistro_name, pkg_name)
+    filename = '%s_%s.orig.tar.gz' % (debian_package_name, origtgz_version)
+
+    for repo in debian_repository_urls:
+        URL_TEMPLATE = '%s/pool/main/%s/%s/%s'
+        prefix = debian_package_name[0]
```
dirk-thomas: Please move these two line before the loop.

ros_buildfarm/sourcedeb_job.py
```diff
@@ -59,6 +60,25 @@ def get_sources(
...
+        url = URL_TEMPLATE % (repo, prefix, debian_package_name, filename)
+
+        try:
+            output_file = os.path.join(sources_dir, '..', filename)
+            urlretrieve(url, output_file)
+            print("Found matching original tarball, "
+                  "downloading %s to %s" % (url, output_file))
```
dirk-thomas: Please move this line into an `else` block.

Please put single quotes around the placeholders.

At this point the download should already be complete Therefore the message should be something like this:

> Downloaded original tarball %s from %s

ros_buildfarm/sourcedeb_job.py
```diff
@@ -59,6 +60,25 @@ def get_sources(
...
+            continue
+        except:
+            print("No tarball found at %s, that's ok it will be rebuilt" % url)
```
dirk-thomas: Please put single quotes around the placeholders.

The part of the message ("it will be rebuilt") might be misleading since it might still be downloaded from another repo. Therefore I would suggest to leave that part out.

tfoote:
@dirk-thomas Good points, comments addressed. The continue was supposed to be a break.

ros_buildfarm/sourcedeb_job.py
dirk-thomas: The first sentence still applies:

> Please move this line into an else block.

dirk-thomas:
@j-rivero has added a +1 to the above comment https://github.com/ros-infrastructure/ros_buildfarm/pull/374#issuecomment-280832332

dirk-thomas:
We will merge this before the automatic reconfigure of the source jobs tonight.

dirk-thomas:
Manually squashed into 06e7b6c1227cfa93168fb549c31466bb8b45c58e

dirk-thomas:
Just for the record: this patch seems to be the reason for this regression: https://github.com/ros/rosdistro/pull/14091#issuecomment-284465164

PR #376, opened 2017-02-17
a patch that allowed generate_release_script to run locally

tfoote:
In testing #374 I had to do this to successfully run generate_release_script. I'm not sure it's quite the right fix but it at least ran.


dirk-thomas:
Replaced by #386.

PR #378, opened 2017-02-17
Customizable age threshold

dirk-thomas:
Regenerate the blocked and compare pages only every six hours (16 pages). The other status pages every 20 minutes (14 pages). This should reduce the load by more than half. Hopefully other management jobs won't starve anymore.

Update the age threshold logic to be configurable per page and match the new schedule.

### DIFF
--- a/ros_buildfarm/templates/status/blocked_releases_page.html.em
+++ b/ros_buildfarm/templates/status/blocked_releases_page.html.em
@@ -29,6 +29,7 @@
 import time
 }@
   <script type="text/javascript">
+    window.age_threshold_green = moment.duration(7, 'hours');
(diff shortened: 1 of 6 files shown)


dirk-thomas:
I think it is better to trigger them "5 min past" because then they have 10 min. to finish until the next round of trigger jobs are scheduled (every 15 min.).

dirk-thomas:
Lets merge and see if it helps with the starvation and consider updating the intervals if needed.

mikaelarguedas:
Sounds good

tfoote:
+1 note that if we want the blocked page at a specific time, at least in the building we can manually trigger them and have the results in 10 minutes instead of increasing the steady state load. Once we're out of the release process it might even be worth backing them off to nightly.

PR #381, opened 2017-02-22
check for test dependencies as well

mikaelarguedas:
Devel jobs fail because packages marked as `releasable` don't have all test dependencies satisfied. e.g. http://build.ros.org/job/Ldev__angles__ubuntu_xenial_amd64/1/

This adds test dependencies to the deps to check

### DIFF
--- a/ros_buildfarm/status_page.py
+++ b/ros_buildfarm/status_page.py
@@ -768,7 +768,7 @@ def _get_blocked_releases_info(
             # Accumulate all dependencies for those packages
             for package in packages:
                 recursive_dependencies = dependency_walker.get_recursive_depends(
-                    package, ['build', 'run', 'buildtool'], ros_packages_only=True,
+                    package, ['build', 'buildtool', 'run', 'test'], ros_packages_only=True,
                     limit_depth=depth)
                 package_dependencies = package_dependencies.union(
                     recursive_dependencies)


tfoote:
This looks reasonable. As we talked briefly it might be worth doing a 2nd query and separating out those that are blocked by test dependencies only. Since they are releasable, it's just that the devel jobs will fail until the dependencies are met.

dhood:
For my understanding, would somebody please explain if we encourage maintainers to:

1. release as soon as possible by the "buildable+runnable" metric so that fewer packages are blocking others, or 
1. only release testable packages (presumably with passing tests)

or perhaps it depends on significance of the package? i.e. what was the metric we used before this page was created?

mikaelarguedas:
I think that we encourage people to release as soon as possible for a new distro. So buildable+runnable could be enough.
Then it's still great to know that package A does work (testable and passing) before asking maintainers to release packages that depend on A. Otherwise any fix will result in a rebuild of A and all downstream packages instead of just a rebuild of A. The lower the package in the stack the longer it takes to rebuild.

Not sure what the metric was before so I'm gessing it was something like: release `ros_base`, then poke all maintainers of `desktop-full` to release their packages. It was up to the maintainers to ensure that their dependencies were available. I think that was the main motivation for creating this page

dhood:
thanks for the response @mikaelarguedas! OK, the decision was probably left up to the maintainers before. As you pointed out it makes sense to put stricter requirements on packages lower down the stack to avoid rebuilds: thanks for the info.

mikaelarguedas:
If it's ok for everyone, I'm going to merge this as is given that the lunar beta cycle is about to start and that I don't think I'll have time to do a smarter query + display as suggested in https://github.com/ros-infrastructure/ros_buildfarm/pull/381#issuecomment-281761707 by then

dirk-thomas:
Sounds good to me.

PR #383, opened 2017-02-22
use python-rosdistro-modules where possible

dirk-thomas:
Similar to #379.

The only location requiring `python3-rosdistro` is https://github.com/ros-infrastructure/ros_buildfarm/blob/ec218740829145a27b414b66e586d42255213fae/ros_buildfarm/templates/misc/rosdistro_cache_task.Dockerfile.em#L35 since the job invokes one of the rosdistro scripts: `rosdistro_build_cache`.

### DIFF
--- a/stdeb.cfg
+++ b/stdeb.cfg
@@ -1,6 +1,6 @@
 [DEFAULT]
-Depends: python-argparse, python-catkin-pkg-modules, python-configparser, python-empy, python-rosdistro (>= 0.4.0), python-yaml, python3-empy
-Depends3: python3-catkin-pkg-modules, python3-empy, python3-rosdistro (>= 0.4.0), python3-yaml
+Depends: python-argparse, python-catkin-pkg-modules, python-configparser, python-empy, python-rosdistro-modules, python-yaml, python3-empy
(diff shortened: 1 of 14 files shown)


stdeb.cfg
```diff
@@ -1,6 +1,6 @@
 [DEFAULT]
-Depends: python-argparse, python-catkin-pkg-modules, python-configparser, python-empy, python-rosdistro (>= 0.4.0), python-yaml, python3-empy
-Depends3: python3-catkin-pkg-modules, python3-empy, python3-rosdistro (>= 0.4.0), python3-yaml
+Depends: python-argparse, python-catkin-pkg-modules, python-configparser, python-empy, python-rosdistro-modules, python-yaml, python3-empy
+Depends3: python3-catkin-pkg-modules, python3-empy, python3-rosdistro-modules-modules, python3-yaml
```
tfoote: Double modules here.

dirk-thomas: Great catch. I must have been overly eager with search-and-replace. Thx.

dirk-thomas:
The only usage of any of the scripts provided by rosdistro was the above mentioned `rosdistro_build_cache`. So this should good to go now.

PR #386, opened 2017-02-24
Fix generate release script

dirk-thomas:
Replaces #376.

This regressed in #164 which replaced the `Interpreter.file()` call with a `Interpreter.string()` call.

### DIFF
--- a/ros_buildfarm/templates/__init__.py
+++ b/ros_buildfarm/templates/__init__.py
@@ -96,7 +96,9 @@ def expand_template(template_name, data, options=None):
 
         with open(template_path, 'r') as h:
             content = h.read()
+            interpreter.invoke('beforeFile', name=template_name, file=h)
         interpreter.string(content, template_path, locals=data)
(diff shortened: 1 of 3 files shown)


ros_buildfarm/templates/__init__.py
```diff
@@ -96,7 +96,9 @@ def expand_template(template_name, data, options=None):
 
         with open(template_path, 'r') as h:
             content = h.read()
+            interpreter.invoke('beforeFile', name=template_name, file=h)
         interpreter.string(content, template_path, locals=data)
+        interpreter.invoke('afterFile')
```
tfoote: Is this a standard hook? I don't see an implementation like `beforeFile`

dirk-thomas: Yes, the base class has a lot of empty methods for various hooks: https://github.com/dirk-thomas/empy/blob/0df7f44312071b976d202a3f8a5fd2ce6aacd174/em.py#L971-L1035

Issue #369, opened 2017-02-14
debian builds are failing on KeyError: "The cache has no package named 'apt-src'"


tfoote:
KeyError: "The cache has no package named 'apt-src'"

This is happening across many jobs, some are flagged in #368 I've also seen completely unreleated jobs: http://build.ros.org/job/Kbin_dj_dJ64__cl_tf2__debian_jessie_amd64__binary/13/console

```
Traceback (most recent call last):
  File "/usr/lib/python3/dist-packages/apt/cache.py", line 201, in __getitem__
    return self._weakref[key]
  File "/usr/lib/python3.4/weakref.py", line 125, in __getitem__
    o = self.data[key]()
KeyError: 'apt-src'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/tmp/ros_buildfarm/scripts/release/create_binarydeb_task_generator.py", line 163, in <module>
    main()
  File "/tmp/ros_buildfarm/scripts/release/create_binarydeb_task_generator.py", line 85, in main
    apt_cache, debian_pkg_names)
  File "/tmp/ros_buildfarm/ros_buildfarm/common.py", line 134, in get_binary_package_versions
    pkg = apt_cache[debian_pkg_name]
  File "/usr/lib/python3/dist-packages/apt/cache.py", line 208, in __getitem__
    raise KeyError('The cache has no package named %r' % key)
KeyError: "The cache has no package named 'apt-src'"
```

dirk-thomas:
If the `apt update` command uses `http://http.debian.net` the apt cache doesn't contain an entry for the `apt-src` package (and our script fails with the `KeyError`). If it uses `http://deb.debian.org` the apt cache has the expected entry and the build succeeds.

Could it be that we are randomly hitting an inconsistent Debian mirror?

dirk-thomas:
#370 switches to use `deb.debian.org`. We will merge that and see in the next days if the problem happens again.

tfoote:
Unfortunately it seems that there's still a random mirror failing. I just noticed another failure here and it only pulls from http://deb.debian.org : http://build.ros.org/view/Kbin_djv8_dJv8/job/Kbin_djv8_dJv8__agvs_description__debian_jessie_arm64__binary/8/console

dirk-thomas:
We will merge this before the automatic reconfigure of the source jobs tonight.

dirk-thomas:
[...]

PR #387, opened 2017-02-27
workaround sporadically missing apt-src on Debian

dirk-thomas:
Workaround for #369.

### DIFF
--- a/ros_buildfarm/templates/release/binarydeb_create_task.Dockerfile.em
+++ b/ros_buildfarm/templates/release/binarydeb_create_task.Dockerfile.em
@@ -57,6 +57,11 @@ RUN echo "@today_str"
     os_name=os_name,
     os_code_name=os_code_name,
 ))@
+@(TEMPLATE(
+    'snippet/install_apt-src.Dockerfile.em',
+    os_name=os_name,
+    os_code_name=os_code_name,
+))@
 
 RUN python3 -u /tmp/wrapper_scripts/apt.py update-install-clean -q -y devscripts dpkg-dev python3-apt python3-catkin-pkg-modules python3-empy python3-rosdistro-modules python3-yaml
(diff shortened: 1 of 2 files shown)


dirk-thomas:
Currently ros-infrastructure/buildfarm_deployment#145 is blocking this to be tested.

mikaelarguedas:
@dirk-thomas this can be tested now that https://github.com/ros-infrastructure/buildfarm_deployment/issues/145 is resolved

dirk-thomas:
All recently built Lunar Debian Stretch jobs (for amd64 as well as armv8) finished without any of the builds failing with this problem. Therefore I am leaning to not merge this PR and close it (until we see the problem again).

mikaelarguedas:
Sounds good to me, almost done rebuilding all the Stretch packages and all is green so far

dirk-thomas:
This has failed again for Debian Jessie armv8 (http://build.ros.org/view/Kbin_djv8_dJv8/job/Kbin_djv8_dJv8__depth_image_proc__debian_jessie_arm64__binary/35/console). Since it hasn't happened for Stretch in the past rebuild but does happen for Jessie maybe this should be updated to only apply the workaround on Jessie?

tfoote:
I see you reopened this. For the record I'm seeing about 10-20 instances of this per day over the weekend.

dirk-thomas:
Indeed, I took care of manually retriggering all of them for the sync.

Did you see them only on Jessie (that's what the patch targets now) or also any on Stretch or other platforms?

tfoote:
Searching the notifications I only see them on Jessie armv8 since Feb 16th. 
Before that I see both Jessie amd64 and armv8.

These errors appear to have started Jan 24th. 
I'm guessing the Jessie amd64

PR #377, opened 2017-02-17
add magic docker fix env var

dirk-thomas:
Fixes http://build.ros.org/view/Kbin_dj_dJ64/job/Kbin_dj_dJ64__mvsim__debian_jessie_amd64__binary/52/

No questions asked :zipper_mouth_face:

PR #388, opened 2017-03-03
Adding more padding to try to miss alignment issue

tfoote:
This is a follow up to #377 which avoided the 512 * 2*n boundary issue for one value. Until we upgrade I think we'll get random other collisions.

### DIFF
--- a/ros_buildfarm/templates/release/binarydeb_task.Dockerfile.em
+++ b/ros_buildfarm/templates/release/binarydeb_task.Dockerfile.em
@@ -64,7 +64,7 @@ RUN echo "@today_str"
 ))@
 
 # magic https://github.com/docker/docker/issues/14203
-ENV DOCKER_FIX ' '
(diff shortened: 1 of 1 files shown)


dirk-thomas:
I don't think we need to merge this since it would only move the problem from one job to a random other one.

Please use this branch to make the failing job pass.

tfoote:
Switching to this branch got it to pass. http://build.ros.org/view/Kbin_uX32/job/Kbin_uX32__camera_info_manager__ubuntu_xenial_i386__binary/28/console  

I'll close this for now, but there may be another random collision we'll need to revive this for.

dirk-thomas:
Since this workaround won't ever fix the problem I will more likely remove the dummy environment variable. If the problem reappears the specific failing job has to be triggered with a custom branch.

I expect the problem to be addressed as soon as ros-infrastructure/buildfarm_deployment#116 is solved.

PR #395, opened 2017-03-08
add support for changing upstream content

tfoote:
This will auto commit changes to the upstream tarball as diffs for the debian increment.
Addresses issue in #374

It's not recommended, but there are a lot of times that this comes up in our community. This will make the sourcedebs robust to that.

This fixed the build: http://build.ros.org/view/Kbin_uxhf_uXhf/job/Ksrc_uX__openhrp3__ubuntu_xenial__source/18/console

### DIFF
--- a/ros_buildfarm/sourcedeb_job.py
+++ b/ros_buildfarm/sourcedeb_job.py
@@ -117,7 +117,7 @@ def build_sourcedeb(sources_dir, os_name=None, os_code_name=None):
         # dpkg-buildpackage args
         '-us', '-uc',
         # set the option for dpkg-source to auto-commit upstream changes
-        # This is needed for debian incrments where the upstream had changed.
+        # This is needed for Debian increments where the upstream has changed.
         # It's not the best practice but we have people doing it a bunch.
         '--hook-source=\'bash -c "echo auto-commit > debian/source/options"\'',
         # debuild args for lintian
--- a/ros_buildfarm/sourcedeb_job.py
+++ b/ros_buildfarm/sourcedeb_job.py
@@ -119,7 +119,8 @@ def build_sourcedeb(sources_dir, os_name=None, os_code_name=None):
         # set the option for dpkg-source to auto-commit upstream changes
         # This is needed for Debian increments where the upstream has changed.
         # It's not the best practice but we have people doing it a bunch.
-        '--hook-source=\'bash -c "echo >> debian/source/options && echo auto-commit >> debian/source/options"\'',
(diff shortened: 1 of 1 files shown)


ros_buildfarm/sourcedeb_job.py
```diff
@@ -116,6 +116,10 @@ def build_sourcedeb(sources_dir, os_name=None, os_code_name=None):
     cmd += [
         # dpkg-buildpackage args
         '-us', '-uc',
+        # set the option for dpkg-source to auto-commit upstream changes
+        # This is needed for Debian increments where the upstream has changed.
+        # It's not the best practice but we have people doing it a bunch.
+        '--hook-source=\'bash -c "echo auto-commit > debian/source/options"\'',
```
dirk-thomas: Since the file `debian/source/options` might already exist it would be better to append `auto-commit` rather than overwrite the content.

tfoote: I updated it to append instead of write. There's still some edge cases, like what happens if it's already defined. But I think those are fewer and a good config file parser should handle that.

ros_buildfarm/sourcedeb_job.py
```diff
@@ -116,6 +116,10 @@ def build_sourcedeb(sources_dir, os_name=None, os_code_name=None):
     cmd += [
         # dpkg-buildpackage args
         '-us', '-uc',
+        # set the option for dpkg-source to auto-commit upstream changes
+        # This is needed for Debian increments where the upstream has changed.
+        # It's not the best practice but we have people doing it a bunch.
+        '--hook-source=\'bash -c "echo auto-commit >> debian/source/options"\'',
```
dirk-thomas: This only works when an existing file has a trailing newline. Please update it to also work when an existing file doesn't.

tfoote: injected new line

dirk-thomas:
I retriggered http://build.ros.org/view/Kbin_djv8_dJv8/job/Ksrc_dJ__openhrp3__debian_jessie__source/18/ Once all downstream jobs have finished *Kinetic* can be synced again. Then we can merge pending rosdistro PRs.

dirk-thomas:
@tfoote This patch is failing new source jobs on Zesty and Stretch. Can you please at this asap, e.g.:

* http://build.ros.org/job/Lsrc_uZ__genpy__ubuntu_zesty__source/7/
* http://build.ros.org/job/Lsrc_dS__genpy__debian_stretch__source/12/

PR #396, opened 2017-03-08
run local jobs during tests

dirk-thomas:
With this patch Travis runs:
* nosetests to check style
* run a devel job for `roscpp_core`
* run a doc job for `roscpp_core`
* run a release job for `rostime`
* run a prerelease job for the underlay `cpp_common` and `roscpp_core` and the overlay `roscpp`

That should at least catch some future regressions before trying the changes on Jenkins.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,77 @@ language: python
 python:
   - "3.4"
 sudo: false
+env:
+  global:
+    - ROS_DISTRO_NAME=kinetic
+    - OS_NAME=ubuntu
+    - OS_CODE_NAME=xenial
+    - ARCH=amd64
+matrix:
+  include:
+    - language: python
+      python: "3.4"
+      sudo: required
+      services:
+        - docker
+      env: JOB_TYPE=devel REPOSITORY_NAME=roscpp_core
+      before_script:
+        # install catkin for test results
+        - sudo sh -c 'echo "deb http://packages.ros.org/ros/ubuntu $(lsb_release -sc) main" > /etc/apt/sources.list.d/ros.list'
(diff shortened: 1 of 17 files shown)


.travis.yml
```diff
@@ -2,6 +2,78 @@ language: python
...
+      before_script:
+        # install catkin for test results
+        - sudo sh -c 'echo "deb http://packages.ros.org/ros/ubuntu $(lsb_release -sc) main" > /etc/apt/sources.list.d/ros.list'
+        - sudo apt-key adv --keyserver hkp://ha.pool.sks-keyservers.net:80 --recv-key 421C365BD9FF1F717815A3895523BAEEB01FA116
+        - sudo apt-get update -qq
+        - sudo apt-get install ros-indigo-catkin -y
+        - python setup.py install
+        - mkdir job && cd job
+      script:
+        - generate_prerelease_script.py https://raw.githubusercontent.com/ros-infrastructure/ros_buildfarm_config/production/index.yaml $ROS_DISTRO_NAME default $OS_NAME $OS_CODE_NAME $ARCH $UNDERLAY_REPOSITORY_NAMES --level 0 --pkg $OVERLAY_PACKAGE_NAMES --output-dir .
```
tfoote: It would be great to do at least one level of overlay to get those code paths. Though the sample package might want to be changed to something that won't have too many one level dependencies.

dirk-thomas: I would rather not use `depth` for the Travis test since the amount of work can increase unnoticed when the rosdistro changes over time. Therefore I picked a deterministic set.

If you want to check the `depth` option I would recommend writing a unit test for the code path which computes the list of packages to be selected using a mock distro.

PR #398, opened 2017-03-11
Don't use hooks on saucy and trusty.

tfoote:
dpkg-buildpackage does not have the hooks option on those versions.

Follow up: https://github.com/ros-infrastructure/ros_buildfarm/pull/397#issuecomment-285828348

### DIFF
--- a/ros_buildfarm/sourcedeb_job.py
+++ b/ros_buildfarm/sourcedeb_job.py
@@ -114,22 +114,29 @@ def build_sourcedeb(sources_dir, os_name=None, os_code_name=None):
         # do not sign the .buildinfo file, since dpkg 1.18.19
         cmd.append('-ui')
 
-    def older_gbp():
-        yield ('debian', 'jessie')
-        for distro_name in ['saucy', 'trusty', 'utopic', 'vivid',
-                            'wily', 'xenial', 'yakkety']:
+    def no_hook_gbp():
+        for distro_name in ['saucy', 'trusty']:
             yield ('ubuntu', distro_name)
 
-    hook_quote = "'" if (os_name, os_code_name) in older_gbp() else ''
+    def older_gbp():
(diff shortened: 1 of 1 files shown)


tfoote:
It passed here with my testing: http://build.ros.org/job/Jsrc_uT__message_filters__ubuntu_trusty__source/8/console

ros_buildfarm/sourcedeb_job.py
```diff
@@ -116,20 +116,27 @@ def build_sourcedeb(sources_dir, os_name=None, os_code_name=None):
 
     def older_gbp():
         yield ('debian', 'jessie')
-        for distro_name in ['saucy', 'trusty', 'utopic', 'vivid',
+        for distro_name in ['saucy', 'utopic', 'vivid',
```
dirk-thomas: Should `saucy` be removed from this?

tfoote: fixed, i forgot to save the change locally. It's not explicitly relevant, since it would be excluded anyway.

dirk-thomas:
I fixed the style errors.

dirk-thomas:
If `saucy` and `trusty` don't have the hook argument what happens if they download an existing tarball but than have a diff?

tfoote:
This means for them we need to find another way to insert that auto-commit file.

dirk-thomas:
Do you want to go ahead an merge this as is in order to unblock the farm? And work on another fix separately?

tfoote:
yeah, this will let it keep building, and I'll need to start digging through man pages again.

tfoote:
I'll try to find a solution that works for all the versions and doesn't use the hooks at all which will get rid of all this special casing.

dirk-thomas:
I will merge this now (only the prerelease test is ongoing which is unaffected by the change).

PR #405, opened 2017-03-16
a few prerelease improvements

dirk-thomas:
* Change the default of the `--level` argument to 0 since it is the most common value (2 doesn't make much sense in many cases).
* Skip processing the overlay workspace if it is empty anyway (saving quite some time).
* Fix the prerelease script when using only external repos which don't have any source entries in the distro.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -80,6 +80,19 @@ matrix:
       sudo: required
       services:
         - docker
+      env: JOB_TYPE=external_prerelease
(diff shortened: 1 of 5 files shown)


ipa-mdl:
works like a charm (https://github.com/ros-industrial/industrial_ci/pull/145)

dirk-thomas:
@ipa-mdl Thank you for the feedback.

PR #408, opened 2017-04-03
avoid switching between dst and non-dst timezone

dirk-thomas:
Instead of using DST or non-DST when generating the jobs (which flip flops twice a year) always use the non-DST timezone. While this will result in the same local time it keeps the jobs configuration stable.


tfoote:
It looks like time.timezone gives the local (non-DST) timezone: https://docs.python.org/2/library/time.html#time.timezone and tzname gives both the local and DST names.

dirk-thomas:
Thanks, looks much more concise: 10755d70ade7a8790af1be9e7c3e66d023569d73

Issue #409, opened 2017-04-07
Doc jobs don't use custom rosdep URLs


ayrton04:
We are having issues resolving dependencies in our doc jobs. We have a custom rodistro setup, and so I decided to compare the issue with some dev jobs, which are able to resolve dependencies just fine. Below is some sample output (server URLs redacted):

**Dev job:**
```
06:11:12   ENV ROSDISTRO_INDEX_URL http://our-rosdistro-server/path/to/master/index.yaml
06:11:12  
06:11:12   RUN rosdep init
06:11:12   RUN (cd /etc/ros/rosdep/sources.list.d/ && wget --no-verbose -O $(basename http://our-rosd-stro-server/path/to/master/rosdep/sources.list.d/20-default.list )  http://our-rosd-stro-server/path/to/master/rosdep/sources.list.d/20-default.list )
(several steps later, we call rosdep update)
```

**Doc job:**
```
05:58:22 Step 19 : ENV ROSDISTRO_INDEX_URL http://our-rosdistro-server/path/to/master/index.yaml
05:58:22  ---> Running in 3077aad905bf
05:58:22  ---> 3396b2975a6c
05:58:22 Removing intermediate container 3077aad905bf
05:58:22 Step 20 : RUN rosdep init
05:58:22  ---> Running in b7a11d655a35
05:58:23 Wrote /etc/ros/rosdep/sources.list.d/20-default.list
(several steps later, we call rosdep update)
```

Looking at their templates, the [devel job](https://github.com/ros-infrastructure/ros_buildfarm/blob/2d14e80509a9055ff55ed22a6ccda18d8533ce5f/ros_buildfarm/templates/devel/devel_create_tasks.Dockerfile.em#L54) uses a [snippet](https://github.com/ros-infrastructure/ros_buildfarm/blob/2d14e80509a9055ff55ed22a6ccda18d8533ce5f/ros_buildfarm/templates/snippet/rosdep_init.Dockerfile.em) for `rosdep init`, whereas the [doc job](https://github.com/ros-infrastructure/ros_buildfarm/blob/2d14e80509a9055ff55ed22a6ccda18d8533ce5f/ros_buildfarm/templates/doc/doc_create_task.Dockerfile.em#L48) just calls it.

Are they intentionally different? If not, I can submit a PR.

tfoote:
That looks like an oversight. A PR would be appreciated.

ayrton04:
Will do!

[...]

PR #410, opened 2017-04-07
Adding support for custom rosdep URLs in doc builds

ayrton04:
This addresses #409 by adding support for custom rosdep URLs in doc jobs. I still need to try to test this against our build farm, but I want to open it up for comments in the meantime.

Note that I adapted these changes from the [PR](https://github.com/ros-infrastructure/ros_buildfarm/pull/100) that added the same functionality for devel jobs.

### DIFF
--- a/doc/configuration_options.rst
+++ b/doc/configuration_options.rst
@@ -315,3 +315,14 @@ The following options are valid in version ``2`` (beside the generic options):
 * ``upload_user``: The username to use to rsync the resultant files.
   This should match the config ``upload::docs::user`` in the buildfarm_deployment_config.
   The default is ``jenkins-slave``
+
+The following options are valid as keys in the ``_config`` dict under
+``targets``:
+
+* ``custom_rosdep_urls``: a list of URLs containing rosdep sources.list.d entry
+  files that are downloaded into /etc/ros/rosdep/sources.list.d at the beginning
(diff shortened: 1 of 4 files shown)


dirk-thomas:
Thank you you for adding support for custom rosdep URLs for doc builds.

I just committed 7307d8811c79cd0f6e3c7ef1221409231a1bc474 to this PR which should address the failing unit test. Other than that the patch looks good to me. Let's see what Travis will say...

dirk-thomas:
The job tests fail currently but I it looks like that these [recently added tests](#396) don't do the right thing when the PR branch is in a forked repo (rather than within this repo as it was the case for the PRs since these tests have been added). I will look into it since it would be good have them pass before this is merged.

dirk-thomas:
I think I have addressed the problem in #411. In order to affect Travis for your PR can you please rebase your branch against the current master.

ayrton04:
Yes, will do.

ayrton04:
Whoops, rebase fail. I merged the upstream master into my fork and rebased from that. Will fix.

ayrton04:
Travis failed because of a `pip install` command that experienced a "connection reset by peer" error:

https://travis-ci.org/ros-infrastructure/ros_buildfarm/jobs/220487175#L360

dirk-thomas:
Thank you!

Issue #412, opened 2017-04-07
generate_prerelease_script.py command fails on template expansion


mikaelarguedas:
... when trying to run prerelease for `jsk_recognition` package.
log:
```
$ generate_prerelease_script.py \
>   https://raw.githubusercontent.com/ros-infrastructure/ros_buildfarm_config/production/index.yaml \
>   kinetic default ubuntu xenial amd64 \
>   jsk_recognition \
>   --level 1 \
>   --output-dir ./
Fetching buildfarm configuration...
Fetching rosdistro cache...
Evaluating job templates...
AttributeError in template 'devel/devel_script_clone.sh.em': 'NoneType' object has no attribute 'name'
```
Stumbled upon that when trying to reproduce another bug in the prerelease process. Didn't have time to investigate why some `repo_spec`s passed to this template are `None`.

dirk-thomas:
It fails because the requested repo doesn't have a source entry (see https://github.com/ros/rosdistro/blob/3956fb60080d5c4db8df80265efbf86ba2414f04/kinetic/distribution.yaml#L2341-L2361) therefore a prerelease is not possible for it.

I will make a patch to fail with a better error message though.

dirk-thomas:
You can try running the command with the patch proposed in #413. It should provide a clear error message now.

dirk-thomas:
Did you generate the command with the prerelease website? If yes, that should probably be fixed to not generate that command in the first place.

mikaelarguedas:
yep I generated it via the prerelease website

PR #413, opened 2017-04-07
improve error message when trying a prerelease for a released repo without a source entry

dirk-thomas:
Improves the error message in case of #412.


mikaelarguedas:
I tested it, it's much clearer now, thanks!

Issue #414, opened 2017-04-12
Xenial Jobs Failing with New Ubuntu Docker Image


evenator:
The `latest` tag of the Ubuntu Xenial image at https://hub.docker.com/r/library/ubuntu/ has just been updated to a new image. On our internal build farm, we're seeing all Xenial source jobs failing in Step 4 of the `Build Dockerfile - generate sourcedeb` stage with this failure:

```
Step 4 : RUN locale-gen en_US.UTF-8
 ---> Running in b44c39e2a893
[91m/bin/sh: 1: locale-gen: not found
[0mRemoving intermediate container b44c39e2a893
The command '/bin/sh -c locale-gen en_US.UTF-8' returned a non-zero code: 127
Build step 'Execute shell' marked build as failure
```

It would seem that the new image doesn't have `locale-gen` pre-installed. As far as I can tell, this hasn't hit the main ROS buildfarm at https://build.ros.org (perhaps due to caching).

With the previous `xenial`, `locale-gen` runs ok:

```bash
$ sudo docker run -it --rm ubuntu:xenial-20170214
root@78055af4ac09:/# locale-gen
Generating locales (this might take a while)...
Generation complete.
```

But with the current `xenial:latest` image, it fails:

```bash
$ sudo docker run -it --rm ubuntu:xenial
root@1eaa25e24244:/# locale-gen
bash: locale-gen: command not found
```

mikaelarguedas:
thanks for reporting this. Docker images on the buildfarm are invalidated daily, that would explain why it hasn't hit the main ROS buildfarm yet.

mikaelarguedas:
Should be fixed by #415 .

Thanks again @evenator for the heads up!

PR #415, opened 2017-04-12
ubuntu doesn't ship with locales anymore

mikaelarguedas:
As of today, `xenial` and `zesty` don't ship with locales anymore.
This PR installs it for all ubuntu platforms because I'm expecting the new `yakkety` and `trusty` image to not have it either in the next image.

Addresses https://github.com/ros-infrastructure/ros_buildfarm/issues/414

### DIFF
--- a/ros_buildfarm/templates/devel/devel_create_reconfigure_task.Dockerfile.em
+++ b/ros_buildfarm/templates/devel/devel_create_reconfigure_task.Dockerfile.em
@@ -8,8 +8,6 @@ ENV DEBIAN_FRONTEND noninteractive
 
 @(TEMPLATE(
     'snippet/setup_locale.Dockerfile.em',
-    os_name='ubuntu',
-    os_code_name='trusty',
     timezone=timezone,
 ))@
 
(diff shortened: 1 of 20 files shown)


mikaelarguedas:
@ros-infrastructure/ros_team It would be good to review / iterate on this quick because that may hit us on any new machine we spawn on the buildfarm

wjwwood:
Is there any downside to doing this (other than wasted effort on machines/images that already have it)?

Looks reasonable to me, but I'm not super familiar with the issue.

mikaelarguedas:
I don't see any downside except the one you mentioned. But it's also not my main area of expertise.
I noticed that the ROS2 buildfarm is doing it without condition for the xenial image and it didn't cause unexpected behaviours (that I know of)

mikaelarguedas:
> It looks like the argument os_name isn't used anymore in the template. So it should be removed from all invocations of the template.

Done in https://github.com/ros-infrastructure/ros_buildfarm/pull/415/commits/bce51283480685111a45a7767e57f12a95f047d9

I saw the first occurrences of this on several jobs last night: (e.g. http://build.ros.org/job/Ldoc__robot_state_publisher__ubuntu_xenial_amd64/1/console#console-section-11, http://build.ros.org/job/Ldev__ros_control__ubuntu_xenial_amd64/4/console#console-section-8)

ayrton04:
+1 for this

dirk-thomas:
@mikaelarguedas I will go ahead and merge this. Worst case we will iterate on it in a new PR.

PR #416, opened 2017-04-13
Adding wget to doc_create_task package installs

ayrton04:
The `rosdep_init` snippet requires `wget` for custom rosdep URLs. Verified this change against some jobs on our build farm.


mikaelarguedas:
Thanks for pointing it out. Indeed the rosdep snippet needs `wget` and it is installed in the other template using it but not in this one :+1:

PR #418, opened 2017-04-14
update plugin versions and fix groovy failures

### LINKED ISSUE #417: New source jobs not created
Looks like new entries in the rosdistro file don't result in new jobs being created (http://repositories.ros.org/status_page/ros_lunar_default.html?q=RED).
I'm not sure where it came from but it seems to have appeared a few days ago. What changed on the buildfarm (that I know of) since then:
- update of all the jenkins plugins and whitelisted all groovy scripts manually
- #415 
- #416

dirk-thomas:
For now this patch only updates the plugin version.

**But** the plugin configurations still need to be reviewed and updated (especially for `groovy`) in order to be parsable with the new plugin versions.

Fixes #417.

### DIFF
--- a/ros_buildfarm/templates/snippet/reconfigure_jobs.groovy.em
+++ b/ros_buildfarm/templates/snippet/reconfigure_jobs.groovy.em
@@ -61,7 +61,7 @@ if (views.size() != @(expected_num_views)) {
     throw new AbortException("Wrong number of view configs")
 }
 
-views.each{
+for (it in views) {
     found_view = false
     view_name = it.getName()
     view_config = new File(it.path).getText('UTF-8')
@@ -165,7 +165,7 @@ def mergePullRequestData(job_name, current_file, job_config) {
     return format_xml(document2)
 }
 
-jobs.each{
+for (it in jobs) {
     found_project = false
     job_name = it.getName()
     // remove leading serial number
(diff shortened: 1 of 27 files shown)


mikaelarguedas:
Got Lrel_reconfigure-jobs working finally. I'm going to merge this and regenerate / deploy all the jobs and see how it goes

dirk-thomas:
Have you checked the configuration from the other updated plugins? While they might not have such breaking changes they still need to be reviewed / updated.

mikaelarguedas:
agreed, but I won't have time to do this today. If this fix allows us to get the buildfarm up for the weekend I'd be in favor of merging and deploying this and iterate on the rest of the plugins as soon as someone has time to do it

ros_buildfarm/templates/snippet/reconfigure_jobs.groovy.em
```diff
@@ -61,7 +61,7 @@ if (views.size() != @(expected_num_views)) {
     throw new AbortException("Wrong number of view configs")
 }
 
-views.each{
+for (it in views) {
```
tfoote: It is really odd that the iterator was not resolving. But at least the for loop works.

dirk-thomas:
This can be merged once CI has passed.

But the remaining problems have to be followed up on asap:

* check the configuration of the remaining plugins
* update the buildfarm deployment logic to install the diffutils archive
* document the signatures which require approval (if it can't be automated)

mikaelarguedas:
yes I'm planning on creating an issue for 1 and one issue for 2 and 3 with a gist referring all the approval once the list is complete

mikaelarguedas:
Note: this branch should be kept until all jobs have been reconfigured

PR #423, opened 2017-04-21
avoid iterating all items if not necessary

dirk-thomas:
While debugging a problem I noticed that these loop can be optimized which will safe some ms for each job.


dirk-thomas:
One of the style checks is failing due to https://gitlab.com/pycqa/flake8-docstrings/issues/19 I will merge this anyway to benefit from the performance improvement.

PR #424, opened 2017-04-21
update config of plugins to minimize diff

### LINKED ISSUE #420: Check configuration of all plugins update recently
https://github.com/ros-infrastructure/ros_buildfarm/pull/418 fixes the breaking issues encountered after updating groovy 2 but all the other plugins should configuration should also be checked to make sure we didn't change behavior in a more subtle way

dirk-thomas:
Update the configuration for all plugins to match the latest versions.

Fixes #420.

### DIFF
--- a/ros_buildfarm/templates/release/import_package_job.xml.em
+++ b/ros_buildfarm/templates/release/import_package_job.xml.em
@@ -9,6 +9,10 @@
     num_to_keep=10000,
 ))@
 @(SNIPPET(
+    'property_job-priority',
+    priority=-1,
+))@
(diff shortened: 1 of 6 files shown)


dirk-thomas:
One difference appearing in the reconfigure logs seems to be related to a problem in Jenkins form handling: https://issues.jenkins-ci.org/browse/JENKINS-43776

dirk-thomas:
One of the style checks is failing due to https://gitlab.com/pycqa/flake8-docstrings/issues/19 I will merge this anyway.

PR #425, opened 2017-04-25
modify compare page to list packages

dirk-thomas:
Currently the page lists repositories but when a package is being moved to a different repo that can't be shown. While the list of packages is longer it handles that case easily. Commonly the viewer filters the list by package / repo / maintainer name anyway.

### DIFF
--- a/doc/jobs/miscellaneous_jobs.rst
+++ b/doc/jobs/miscellaneous_jobs.rst
@@ -27,11 +27,11 @@ Dashboard
 Status pages
 ------------
 
-The following jobs generate HTML pages to visualize the status of the package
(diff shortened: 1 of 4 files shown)


mikaelarguedas:
thanks for doing this! While the list will be longer that makes a lot more sense than comparing only repos. Did you run a job to produce one of these "compare repos" status page ?

dirk-thomas:
Not on Jenkins - only locally.

dirk-thomas:
One of the style checks is failing due to https://gitlab.com/pycqa/flake8-docstrings/issues/19 I will merge this anyway.

PR #430, opened 2017-04-27
improve performance to collect recursive dependencies

dirk-thomas:
E.g. for Indigo it speeds up collecting all recursive dependencies (which is the longest running part to order topologically) by approximately 25%.

Similar to ros-infrastructure/catkin_pkg#166.


dirk-thomas:
One of the style checks is failing due to https://gitlab.com/pycqa/flake8-docstrings/issues/19 This won't block merging this.

PR #433, opened 2017-05-05
Git credentials

marc-hanheide:
In order to facilitate builds for private repositories (on github or gitlab), suitable credentials need to be passed on to jenkins for the devel jobs. As I couldn't find a suitable solution in the current build farm I implemented this. Not sure if this is useful for others, but it might well be. Also happy for pointers to other solutions.

This assigns credentials for a git-based repository using the dictionary `git_credentials` in `index.yaml` of the buildfarm configuration. E.g.:
   ```
   git_credentials:
     "https://github.com/.*": credential_id_1
     "https://gitlab.oru.se/.*": credential_id_1
   ```
The key in this dictionary is a regular expression (test case-insensitive) and the value is the respective credential ID in jenkins which needs to be available in the jenkins credential store.

### DIFF
--- a/ros_buildfarm/config/index.py
+++ b/ros_buildfarm/config/index.py
@@ -99,6 +99,12 @@ class Index(object):
             self.git_ssh_credential_id = data['git_ssh_credential_id']
             assert isinstance(self.git_ssh_credential_id, str)
 
+        # find any URL->credential mappings to set credentials for repos
+        self.git_credentials = {}
+        if 'git_credentials' in data:
+            self.git_credentials = data['git_credentials']
+            assert isinstance(self.git_credentials, dict)
+
         self.jenkins_url = data['jenkins_url']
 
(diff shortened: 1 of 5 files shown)


tfoote:
For private repos/private build farms using https the much simpler solution is to use a deployment key or oauth token and embedding it into the urls: https://developer.github.com/guides/managing-deploy-keys/ https://github.com/blog/1270-easier-builds-and-deployments-using-git-over-https-and-oauth

Clearly using this would require the rosdistro to be private as well though. But I think several users are already doing this.

The regexes seem a little bit fragile is there a more canonical way to do this in jenkins? I see that the credentials plugin configs themselves already have regexes built in.

marc-hanheide:
thanks for the pointers, I shall have a look. Not an expert myself. I'm sure there are more elegant ways ;-)

PR #434, opened 2017-05-13
add ORPHANED shortcut link

k-okada:
see https://discourse.ros.org/t/releasing-repositories-form-other-people/1797/15


dirk-thomas:
I will close this PR for now since #435 has been merged. If you would like to support `OR` queries to filter e.g. "unmaintained" as well as "end of life" package please implement it a generic way which allows user to enter a query using that logic.

PR #435, opened 2017-05-13
add UNMAINTAINED shortcut link

k-okada:
simpler version of #434

### DIFF
--- a/ros_buildfarm/templates/status/js/setup.js
+++ b/ros_buildfarm/templates/status/js/setup.js
@@ -6,7 +6,8 @@ var QUERY_TRANSFORMS = {
   'GRAY': '<a class="i"></a>',
   'RED1': '<td><a class="m"></a>',
   'RED2': '</a><a class="m"></a><a',
-  'RED3': '<a class="m"></a></td>'
+  'RED3': '<a class="m"></a></td>',
+  'UNMAINTAINED' : '<span class="unmaintained"'
 };
 
 window.body_ready = function() {
--- a/ros_buildfarm/templates/status/release_status_page.html.em
+++ b/ros_buildfarm/templates/status/release_status_page.html.em
@@ -60,6 +60,7 @@
       <a href="?q=ORANGE">ORANGE</a>,
       <a href="?q=YELLOW">YELLOW</a>,
       <a href="?q=GRAY">GRAY</a>
+      <a href="?q=UNMAINTAINED" title="Filter packages with unmaintained status">UNMAINTAINED</a>
     </p>
     <form action="?">
       <input type="text" name="q" id="q" />


k-okada:
travis failed due to 
```
0.05s$ sudo sh -c 'echo "deb http://packages.ros.org/ros/ubuntu $(lsb_release -sc) main" > /etc/apt/sources.list.d/ros.list'
120.11s$ sudo apt-key adv --keyserver hkp://ha.pool.sks-keyservers.net:80 --recv-key 421C365BD9FF1F717815A3895523BAEEB01FA116
Executing: gpg --ignore-time-conflict --no-options --no-default-keyring --homedir /tmp/tmp.HdYQVsoq8k --no-auto-check-trustdb --trust-model always --keyring /etc/apt/trusted.gpg --primary-keyring /etc/apt/trusted.gpg --keyring /etc/apt/trusted.gpg.d/apt.postgresql.org.gpg --keyserver hkp://ha.pool.sks-keyservers.net:80 --recv-key 421C365BD9FF1F717815A3895523BAEEB01FA116
gpg: requesting key B01FA116 from hkp server ha.pool.sks-keyservers.net
gpg: keyserver timed out
gpg: keyserver receive failed: keyserver error
The command "sudo apt-key adv --keyserver hkp://ha.pool.sks-keyservers.net:80 --recv-key 421C365BD9FF1F717815A3895523BAEEB01FA116" failed and exited with 2 during .
Your build has been stopped.
```

ros_buildfarm/templates/status/release_status_page.html.em
```diff
@@ -60,6 +60,7 @@
       <a href="?q=ORANGE">ORANGE</a>,
       <a href="?q=YELLOW">YELLOW</a>,
       <a href="?q=GRAY">GRAY</a>
+      <a href="?q=UNMAINTAINED" title="Filter packages whth unmaintained or end-of-life status">UNMAINTAINED</a>
```
dirk-thomas: Spelling: `whth` -> `with`

This query does not show packages with status "end of life". It only shows packages with status "unmaintained".

ros_buildfarm/templates/status/js/setup.js
```diff
@@ -6,7 +6,8 @@ var QUERY_TRANSFORMS = {
   'GRAY': '<a class="i"></a>',
   'RED1': '<td><a class="m"></a>',
   'RED2': '</a><a class="m"></a><a',
-  'RED3': '<a class="m"></a></td>'
+  'RED3': '<a class="m"></a></td>',
+  'UNMAINTAINED' : '<span class="unmaintained"'
```
dirk-thomas: The query could just use the lower case search "unmaintained". I don't expect that token to appear in other parts of the status page beside the status column. Therefore this change is not necessary.

k-okada: If I understand the code correctly, it is not likely happens, but in theory if someone make `foo_unmaintaned_pkg` package, then it does not work correctly,

k-okada:
#434 matches both `end-of-life` and `unmaintained`,
#435 matches only `unmaintained`
I personally think #435 is more helpful, but it requires a hack to search OR condition https://github.com/ros-infrastructure/ros_buildfarm/pull/434/files#diff-02b0305ea7fbc00c508896b48b82952bR271

dirk-thomas:
Thanks.

ros_buildfarm/templates/status/release_status_page.html.em
```diff
@@ -60,6 +60,7 @@
       <a href="?q=ORANGE">ORANGE</a>,
       <a href="?q=YELLOW">YELLOW</a>,
       <a href="?q=GRAY">GRAY</a>
+      <a href="?q=UNMAINTAINED" title="Filter packages with unmaintained status">UNMAINTAINED</a>
```
dirk-thomas: Fixed missing comma in 4df621c68837560e013d813e08d4fffcb66ab549.

PR #436, opened 2017-05-14
support OR syntax

k-okada:
implementd OR logic in a generic way

as suggested in https://github.com/ros-infrastructure/ros_buildfarm/pull/434#issuecomment-301268579

### DIFF
--- a/ros_buildfarm/templates/status/js/setup.js
+++ b/ros_buildfarm/templates/status/js/setup.js
@@ -271,14 +271,16 @@ function filter_table() {
               break;
             }
           }
+          // replace OR to RegExp syntax(|)
+          queries[i] = queries[i].replace(/\s*\OR\s*/, '|');
           if (is_known_query) {
             // search in full row html
-            if (row[0].indexOf(queries[i]) == -1) return null;
+            if (! RegExp(queries[i]).test(row[0])) return null;
           } else {
             // search in plain text of each column
             match = false;
             for (var j = 1; j < row.length; j++) {
-              if (row[j] && row[j].indexOf(queries[i]) != -1) {
+              if (row[j] && RegExp(queries[i]).test(row[j])) {
                 match = true;
                 break;
               }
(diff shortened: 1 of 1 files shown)


ros_buildfarm/templates/status/js/setup.js
```diff
@@ -271,14 +271,16 @@ function filter_table() {
               break;
             }
           }
+          // replace OR to RegExp syntax(|)
+          queries[i] = queries[i].replace(/\s*\OR\s*/, '|');
```
dirk-thomas: This logic seems wrong since it would replace `OR` within `MORE`.

ros_buildfarm/templates/status/js/setup.js
```diff
@@ -271,14 +271,16 @@ function filter_table() {
...
           if (is_known_query) {
             // search in full row html
-            if (row[0].indexOf(queries[i]) == -1) return null;
+            if (! RegExp(queries[i]).test(row[0])) return null;
```
dirk-thomas: Switching the query string into a regex might affect strings which have special meaning within a regex. I am not sure if that is desired in all cases? If so it should probably be mentioned in the help / hover text of the search field.

dirk-thomas:
In general I think `OR` is not used much since the user can simply run two queries and look at both results to get the union. Imo the more common use case is `AND` since match two criteria at the same time is not easily done with separate queries. So supporting `AND` is probably more helpful for users.

k-okada:
Thanks for comment, it seems that 'AND' is already supported as `+` syntax https://github.com/ros-infrastructure/ros_buildfarm/blob/master/ros_buildfarm/templates/status/js/setup.js#L244, this means `A+B` searches `A` and `B`, but `A   +   B` searches `A(space)(space)` and `(space)(space)B`

I first thought that situation where `A|B` and `A   |    B` behaves differently seems strange, but if we look into current `+` syntax, I feel like it is ok.
So, I updated PR to support regex for input query. Now we support `A+B` and `A|B`,

If you still prefer solution support `AND` (plus `+`) and `OR`, please let me know.

dirk-thomas:
I added a title attribute to the input fields to provide the user with more information what can be entered as a query: 92e49d6d5b4940cfbd425ad7c5105ec83bc0fba1

dirk-thomas:
Thanks for the patch!

PR #437, opened 2017-05-15
explicitly reschedule aborted builds

dirk-thomas:
The implicit rescheduling done by the `requeue` plugin seems to not happen anymore (see e.g. http://build.ros.org/view/Lbin_uX64/job/Lbin_uX64__rostime__ubuntu_xenial_amd64__binary/2/). Might be a change within Jenkins since the plugin hasn't changed.

### DIFF
--- a/ros_buildfarm/templates/snippet/builder_check-docker.xml.em
+++ b/ros_buildfarm/templates/snippet/builder_check-docker.xml.em
@@ -17,6 +17,7 @@ else:
     'builder_system-groovy',
     command=
 """// DISABLE SLAVE IF OUTPUT INDICATES DOCKER RUN PROBLEMS
+import hudson.model.ParametersAction
 import hudson.slaves.OfflineCause.UserCause
 import java.io.BufferedReader
 import java.util.regex.Pattern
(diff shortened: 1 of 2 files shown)


tfoote:
I saw at least one successfully requeue this morning when I took the machine offline: http://build.ros.org:8080/job/Lbin_dsv8_dSv8__cpp_common__debian_stretch_arm64__binary/2/console

dirk-thomas:
> I saw at least one successfully requeue this morning when I took the machine offline: http://build.ros.org:8080/job/Lbin_dsv8_dSv8__cpp_common__debian_stretch_arm64__binary/2/console

That case is different since there is no Groovy script involved which marks the current build as aborted.

dirk-thomas:
> I guess the requeue plugin is really focused on machines going offline then.

Sure, but that was also the case before. What has changed is that an aborted job seems to not be rescheduled anymore by that plugin. Imo that makes sense since we decided in the job to abort it. I would expect the rescheduling to continue work if we wouldn't abort the build. But since we want to suppress the notification we have to.

dirk-thomas:
Fix missing import: 95c5a8dc8568e364e5f140526a332916370fbaef

PR #438, opened 2017-05-17
check version in a way that supports Python 2.6

wjwwood:
Similar to this change: https://github.com/ros-infrastructure/catkin_pkg/pull/165

I ran into this because our web server is still using Python 2.6 and we cannot update it ourselves.


wjwwood:
I has to be installed into the virtualenv for the prerelease website.

PR #439, opened 2017-05-17
add ORPHANED that shows both end-of-life and unmaintaned

k-okada:
follow up of https://github.com/ros-infrastructure/ros_buildfarm/pull/434, since we have generic OR method in https://github.com/ros-infrastructure/ros_buildfarm/pull/436#issuecomment-301931055, but we already have UNMAINTAINED keyword, so this might be duplicates.

### DIFF
--- a/ros_buildfarm/templates/status/js/setup.js
+++ b/ros_buildfarm/templates/status/js/setup.js
@@ -7,7 +7,6 @@ var QUERY_TRANSFORMS = {
   'RED1': '<td><a class="m"></a>',
   'RED2': '</a><a class="m"></a><a',
   'RED3': '<a class="m"></a></td>',
-  'UNMAINTAINED' : '<span class="unmaintained"',
   'ORPHANED' : '<span class="unmaintained"|<span class="end-of-life"'
 };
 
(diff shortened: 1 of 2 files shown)


dirk-thomas:
Please remove the "UNMAINTAINED" keyword in favor of this new one. I don't think we should add multiple keywords which are so similar (especially since "unmaintained" can easily be types manually).

k-okada:
done

--
◉ Kei Okada


2017-05-17 11:46 GMT+09:00 Dirk Thomas <notifications@github.com>:

ros_buildfarm/templates/status/js/setup.js
```diff
@@ -7,7 +7,7 @@ var QUERY_TRANSFORMS = {
   'RED1': '<td><a class="m"></a>',
   'RED2': '</a><a class="m"></a><a',
   'RED3': '<a class="m"></a></td>',
-  'UNMAINTAINED' : '<span class="unmaintained"'
+  'ORPHANED' : '<span class="unmaintained"|<span class="end-of-life"',
```
dirk-thomas: Please remove the trailing comma.

k-okada:
sorry, fixed

--
◉ Kei Okada


2017-05-17 12:02 GMT+09:00 Dirk Thomas <notifications@github.com>:

dirk-thomas:
Thanks.

PR #445, opened 2017-05-25
use Xenial Docker images instead of Trusty

dirk-thomas:


dirk-thomas:
I would like to deploy that on a workday in the early morning to leave time to catch potential regressions. So once it has been review I would merge it on the next workday morning.

PR #446, opened 2017-05-25
add option to trigger only not-failed jobs

dirk-thomas:
This adds the new option for being manually triggered.

In a follow up pull request I will create a new nightly job to run this option automatically.

### DIFF
--- a/ros_buildfarm/templates/snippet/builder_system-groovy_verify-upstream.xml.em
+++ b/ros_buildfarm/templates/snippet/builder_system-groovy_verify-upstream.xml.em
@@ -1,3 +1,6 @@
+@{
+code = TEMPLATE('snippet/check_recursive_upstream_projects.groovy.em')
+}@
 @(SNIPPET(
     'builder_system-groovy',
     command=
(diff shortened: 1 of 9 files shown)


dirk-thomas:
Waiting for review...

ros_buildfarm/templates/snippet/builder_system-groovy_verify-upstream.xml.em
```diff
@@ -1,3 +1,6 @@
+@{
+code = TEMPLATE('snippet/check_recursive_upstream_projects.groovy.em')
+}@
```
nuclearsandwich: Should this block be inside the Groovy snippet? This seems to be created malconfigured builds.

I saw this first on my ROS 2 test farm but some recent jobs on build.ros.org like [this one](http://build.ros.org/view/Lbin_uxv8_uXv8/job/Lbin_uxv8_uXv8__libphidget21__ubuntu_xenial_arm64__binary/) which are currently failing due to script approval also look like something isn't quite right.

dirk-thomas: Absolutely. Thanks for reporting this. The line was supposed to assign the content of the file to the variable `code`. But instead it outputs it in-place.

Please see #448 for the fix.

Issue #431, opened 2017-05-02
new job to enumerate repeatedly failing jobs


dirk-thomas:
For each ROS distro there should be a job which e.g. runs nightly and outputs all jobs which have been failing for "a long time". The list should help the person in charge of a ROS distro to make an informed decision what to do about it (e.g. remove the entry from the rosdistro, rollback a recent change, etc.).

The new job type is most likely a sibling of the rosdistro cache job. It should contain a system Groovy script running within the Jenkins masters vm.

dirk-thomas:
@j-rivero In what timeframe are you planning to work on this?

j-rivero:
> @j-rivero In what timeframe are you planning to work on this?

depending on your urgency, probably when we have the setup for ARIAC in place and working fine (means: no this week).

dirk-thomas:
It is not an urgent feature but we want to have it at some point. I don't mind about this week but when do you plan to work on it? E.g. in two weeks, or more likely four? Any estimate would be fine.

j-rivero:
> It is not an urgent feature but we want to have it at some point. I don't mind about this week but when do you plan to work on it? E.g. in two weeks, or more likely four? Any estimate would be fine.

Give me three weeks or ping me if we change the priority and some point.

mikaelarguedas:
@j-rivero Do you have an updated ETA on this ?

j-rivero:
as soon as ARIAC and SRCsim finish, next week?

j-rivero:
Is there anyway of testing locally what the python scripts generates as Jenkins configurations?

dirk-thomas:
> Is there anyway of testing locally what the python scripts generates as Jenkins configurations?

For which jobs? You can run the scripts to generate the maintenance jobs without pushing them to Jenkins. For existing jobs it shows the diff. I am not sure if it shows the full xml for new jobs though (but that could be added for testing if necessary).

j-rivero:
> For which jobs? You can run the scripts to generate the maintenance jobs without pushing them to Jenkins. For existing jobs it shows the diff. I am not sure if it shows the full xml for new jobs though (but that could be added for testing if necessary).

For example, how to check the result of running `./misc/generate_rosdistro_cache_job.py`

dirk-thomas:
> For example, how to check the result of running `./misc/generate_rosdistro_cache_job.py`

[...]

PR #454, opened 2017-06-20
Add new jobs to display the failing jobs by ROS distro

j-rivero:
The pull request should add a new job per each ROS distro. The new job is pretty simple and only run a System Groovy script. 

The Groovy script detect which jobs from each ROS distro are failing and check two different factors to determinate which jobs to report:
  * `min_number_of_failures` : the minimum number of failing runs to report a job
  *  `min_days_failing` : the minimum days old of the first failing run.

If both conditions are met, the job will display the name and URL of the failing job.

I've tested separately the Groovy script and the XML generation of the job but not the generate_all.py script, please give an extra look into it.


ros_buildfarm/templates/snippet/builder_system-groovy_check-for-failing-jobs.xml.em
```diff
@@ -0,0 +1,82 @@
...
+    command=
+"""// CHECK FOR FAILING JOBS
+import hudson.model.Job
+
+// Control the number of failures in a row to check
+min_number_of_failures = 5
+// Control the minimun amount of time failing
+min_days_failing = 7
+// rosdistro to search
+rosdistro = @ros_distribution
```
j-rivero: This empy expansion of ros_distribution is not working. I'm not sure if the format should be different or it is a problem of embedding a empy variable inside the Groovy Script. If the latest, I can think on workarounds.

dirk-thomas: You called the variable `ros_distro` (https://github.com/ros-infrastructure/ros_buildfarm/pull/454/files#diff-f331a1152ef526adbb9d081d5e96a1b8R52) not `ros_distribution`. So you have to use the same name here. And since it is a string you will need quotes around it.

dirk-thomas:
Since the added jobs won't affect any of the existing ones please feel free to deploy them to Jenkins to actually run them and check the result.

j-rivero:
> Since the added jobs won't affect any of the existing ones please feel free to deploy them to Jenkins to actually run them and check the result.

I've generated the Lunar job for testing, [here is the result](http://build.ros.org/job/lunar_failing-jobs/1/console). Note: I had to manually modify the ros_distro expansion in the groovy script.

ros_buildfarm/templates/snippet/builder_system-groovy_check-for-failing-jobs.xml.em
j-rivero: arg! I changed the name just before the commit. I set it to  `ros_distro` but I can not make it to expand playing with the quotes and the @. Any idea?

ros_buildfarm/templates/snippet/builder_system-groovy_check-for-failing-jobs.xml.em
```diff
@@ -0,0 +1,82 @@
...
+failed_jobs.each{ job -> 
+   
+    if (! is_failing_from_n_days(min_days_failing, job))
+      return true
+
+    if (! are_previous_builds_failing(min_number_of_failures, job))
+      return true
+  
+    // Conditions checked, reporting failing job                                  
+    println "Failing job: " + job.name + " " +  hudson.model.Hudson.instance.getRootUrl() + job.getUrl()
```
dirk-thomas: To reduce the output this could just print the url (no text and no job name). This would make it a bit more readable.

scripts/misc/generate_failing_jobs_job.py
```diff
@@ -0,0 +1,59 @@
...
+from ros_buildfarm.argument import add_argument_rosdistro_name
+from ros_buildfarm.config import get_index
+from ros_buildfarm.jenkins import configure_job
+from ros_buildfarm.jenkins import configure_management_view
+from ros_buildfarm.jenkins import connect
+from ros_buildfarm.templates import expand_template
+
+def main(argv=sys.argv[1:]):
+    parser = argparse.ArgumentParser(
+        description="Generate the 'dashboard' job on Jenkins")
```
dirk-thomas: `dashboard` - copy-n-paste.

ros_buildfarm/templates/snippet/builder_system-groovy_check-for-failing-jobs.xml.em
```diff
@@ -0,0 +1,82 @@
+@(SNIPPET(
+    'builder_system-groovy',
+    command=
+"""// CHECK FOR FAILING JOBS
+import hudson.model.Job
+
+// Control the number of failures in a row to check
+min_number_of_failures = 5
+// Control the minimun amount of time failing
+min_days_failing = 7
```
dirk-thomas: @mikaelarguedas @tfoote @wjwwood  Since you are in charge of at least one ROS distro which thresholds (in line 8 and 10) make sense from your point of view?

@mikaelarguedas Can you please look at the output of the current job (http://build.ros.org/job/lunar_failing-jobs/1/console) and check if the result makes sense.

ros_buildfarm/templates/snippet/builder_system-groovy_check-for-failing-jobs.xml.em
```diff
@@ -0,0 +1,82 @@
...
+// rosdistro to search
+rosdistro = "@ros_distro"
+
+boolean is_failing_from_n_days(int min_days_failing,
+                               Job job)
```
dirk-thomas: Please avoid alignment, use this instead:

```
boolean is_failing_from_n_days(
  int min_days_failing, Job job)
```

Same below.

mikaelarguedas:
Looking at the output of the job, I'm wondering if jobs that are not triggered periodically should be in that list (thinking dev jobs and PR jobs). It's useful information to have,  I'm just not sure if it should be displayed as another list, be processed by another job or not considered at all

ros_buildfarm/templates/misc/check_failing_jobs.xml.em
```diff
@@ -0,0 +1,45 @@
...
+))@
+  <assignedNode>master</assignedNode>
+  <canRoam>false</canRoam>
+  <disabled>false</disabled>
+  <blockBuildWhenDownstreamBuilding>false</blockBuildWhenDownstreamBuilding>
+  <blockBuildWhenUpstreamBuilding>false</blockBuildWhenUpstreamBuilding>
+  <triggers>
+@(SNIPPET(
+    'trigger_timer',
+    spec='@daily',
```
dirk-thomas: Lets use a specific time so that the maintainer "know" that e.g. the job has run in the night. Maybe around 5am?

ros_buildfarm/templates/snippet/builder_system-groovy_check-for-failing-jobs.xml.em
```diff
@@ -0,0 +1,82 @@
...
+{
+  now = new Date()
+  start_of_failing_period = now - min_days_failing
+
+  last_success_build = job.getLastSuccessfulBuild()
+  
+  // if there is a last success build check when it was  
+  if (last_success_build)
+  {
```
dirk-thomas: Please keep the open `{` trailing the previous line within a function (for `if`, `while`, etc.).

Same below.

ros_buildfarm/templates/snippet/builder_system-groovy_check-for-failing-jobs.xml.em
```diff
@@ -0,0 +1,82 @@
...
+all_items = hudson.model.Hudson.instance.items
+
+rosdistro_items = all_items.findAll{ job -> job.name.take(1) == initial_rosdistro }
+active_jobs = rosdistro_items.findAll{job -> job.isBuildable()}
+failed_jobs = active_jobs.findAll{job -> job.lastBuild && job.lastBuild.result == hudson.model.Result.FAILURE}
+
+failed_jobs.each{ job -> 
+   
+    if (! is_failing_from_n_days(min_days_failing, job))
+      return true
```
dirk-thomas: Please use consistent indentation (2 spaces in Groovy?).

ros_buildfarm/templates/snippet/builder_system-groovy_check-for-failing-jobs.xml.em
mikaelarguedas: > @mikaelarguedas Can you please look at the output of the current job (http://build.ros.org/job/lunar_failing-jobs/1/console) and check if the result makes sense.

The results look sensible to me, a bunch of dev jobs failed but don't match the thresholds of this build so were not displayed (and that's totally fine as is).

> Since you are in charge of at least one ROS distro which thresholds (in line 8 and 10) make sense from your point of view?

This reflects well the current state of Lunar. Although, I'll leave it up to @tfoote and @wjwwood to give feedback on this because I think lunar doesn't yet have enough packages (and thus failing packages) to give good estimation of what the thresholds should be

tfoote: I don't think we need to wait a full week. If it's been more than 2 days it shouldn't be an intermittent build farm outage.

dirk-thomas: I thought the goal was to identify "long term" failing jobs in order to remove / blacklist them. Even a week seems to be a short amount of time for maintainers to fix the problem.

tfoote:
I agree with @mikaelarguedas I don't consider the PR and dev jobs to be in the same class. We could keep listing them, but maybe segregate them to a separate list.

The state of the dev and PR jobs is the responsibility of the maintainer to monitor. Versus the binary jobs the rosdistro boss needs to pay a lot closer attention to.

ros_buildfarm/templates/snippet/builder_system-groovy_check-for-failing-jobs.xml.em
tfoote: I'd like to change the paradigm of leaving packages failing on the farm to catching bad releases and rolling them back. Leaving the bad releases on the farm blocks other builds and releases from being tested and can get us into a bad state. A rollback of the new release will restore the ability to build the old version, and just like leaving it broken the old version will be available for users. I can't think of a time when a package was failing for 3+days and then recovered without the maintainer doing a new release. So we might as well roll them back in the mean time. It saves resources and unblocks others. 

What would be really useful would be to know the profile of builds of one package across each arch. 
| Package | amd64 | arm64 | 
| foo | :+1:  |   :-1: | 
| bar | :-1: | :-1: |

And the other valuable thing would be to have grouping by packages in a repo as that's the unit of release. So I can judge the successrate of a release. Often regressions are correlated across multiple packages in a release. And having them grouped by repo would be valuable too.


Possibly 

REPONAME | build type1 | build type 2| ...| 
package_name1 | fail_cout | fail_count | ... |
package_name2 | fail_cout | fail_count | ... |
package_name3 | fail_cout | fail_count | ... |

Thus at a very quick glance it would be easy to say repo X has a problem let me find the last release to review. 

And we can filter it to something like 3 failed builds in a row for any package in the repo. Basically enough to keep out an intermittent failures if a build fails for something like an upstream invalidation etc.

j-rivero:
> I added some inline comment but mostly small stuff.

I think I've fixed all the issues, thanks @dirk-thomas for the review!

> I'd like to change the paradigm of leaving packages failing on the farm to catching bad releases and rolling them back.

I see the point although. @tfoote do you think that we need to make the changes under the scope of this PR?

> I agree with @mikaelarguedas I don't consider the PR and dev jobs to be in the same class. We could keep listing them, but maybe segregate them to a separate list.

I also though it the first time I saw the results. I should be able to implement it under the scope of this PR if you think that it is the way to go.

dirk-thomas:
> I think I've fixed all the issues

The patch looks good to me. It implements what was described in #431.

From my point of view I can see the advantage of grouping the jobs based on their type (`doc`, `devel`, `pr`, `sourcedeb`, `binarydeb`) since each group will probably be handled differently (option A). Keep in mind that this is happening in a Groovy script and the logic doesn't have any access to rosdistro, the buildfarm config files, et al. So for some of the mentioned ideas the necessary information might not be available.

Instead the result could also be grouped by the package name / repo name (depending on the job type) as follows (option B):
* split the failing job names at the first double underscore (`__`)
* order the output by the second part (that keeps jobs for the same package name / repo name (depending on the job type) together, but always print the full url as is

Option A makes sense when handling `devel` jobs differently than `release` jobs. Option B makes sense when considering package/repo specific failures. The both have the downside that they are not supporting the opposite case well.

I will leave it up to @mikaelarguedas and @tfoote to clearly describe what they would like to have (given the information available). Please describe as concretely as possible than @j-rivero can likely update the PR with that. If the request becomes "too fancy" I would suggest to merge the current state and let @mikaelarguedas and/or @tfoote implement the view they would like to have.

tfoote:
Here's the perfect case for it to catch. 

I checked and there's 192 regressions. But after clicking through I found that most builds are aborted and the one recent release is failing and breaking all the downstream. To find the failure I clicked through the dependency tree walking upward. I found http://build.ros.org/view/Ibin_uT64/job/Ibin_uT64__gazebo_ros__ubuntu_trusty_amd64__binary/ and created: https://github.com/ros/rosdistro/pull/15372

It's hard for me to know if that's the only failing build or just the first one that I found. 

A snapshot of the status page. 
image

dirk-thomas:
@j-rivero The existing job has been failing multiple times. The error message indicates that this patch needs to be updated to catch those conditions.

@mikaelarguedas @tfoote Please describe the behavior you would like to see as requested earlier. Otherwise we will stick to what is implemented right now.

j-rivero:
> @j-rivero The existing job has been failing multiple times. The error message indicates that this patch needs to be updated to catch those conditions.

Sorry Dirk, which one you mean by "the existing job"? I got lost in the conversation.

dirk-thomas:
> Sorry Dirk, which one you mean by "the existing job"? I got lost in the conversation.

The job which this PR adds: http://build.ros.org/job/lunar_failing-jobs/

mikaelarguedas:
@j-rivero example of failure of this job recently http://build.ros.org/job/lunar_failing-jobs/15/console

As I mentioned I think that being able to list separately jobs that are the responsibility of the maintainer (dev and PR jobs) and the job requiring action from the ROS boss (binary jobs).

Then as I said I don't have enough data to advise on whether option A or option B makes more sense in general. I guess that option A would be enough for my current use case. But I leave the floor to @tfoote to give feedback on the way to go. I'm happy to get this merged and iterate on it later on if we identify changes according to our use of it.

tfoote:
There are a lot more features that would be great as I outlined here: https://github.com/ros-infrastructure/ros_buildfarm/pull/454#discussion_r123415232 however that's quite a bit of feature scope. I'd suggest merging this now and we can look to add those features in the future.

tfoote:
Sorry wrong button

j-rivero:
> The job which this PR adds: http://build.ros.org/job/lunar_failing-jobs/

arg .. I saw the last builds ok and thought that you mean other job. I've added a check at the only point in the code pointed by the error (getNumber method) in 52c8c63. The job is still [happy](http://build.ros.org/job/lunar_failing-jobs/21/) although the conditions that triggered the errors are probably not present.

> I'd suggest merging this now and we can look to add those features in the future.

Ok! feel free to make the requirements for them and I can take care of coding them.

ros_buildfarm/templates/snippet/builder_system-groovy_check-for-failing-jobs.xml.em
```diff
@@ -0,0 +1,82 @@
...
+  if (last_success_build) {
+    return last_success_build.getTime() < start_of_failing_period
+  }
+  
+  // all builds are failing, check when was the first one
+  return job.getFirstBuild().getTime() < start_of_failing_period       
+}
+
+
+// check if the N previous from referece_build_num failed
```
dirk-thomas: Spelling: `referece_build_num`

Also what is `referece_build_num` trying to reference? I couldn't find `reference_build_num` anywhere.

j-rivero: ouch, outdated comment. should make some sense after 0fe10f7

dirk-thomas:
If there are no objections I will merge this and afterwards make a new release.

mikaelarguedas:
> If there are no objections I will merge this and afterwards make a new release.

:+1:

PR #457, opened 2017-06-27
avoid installing wrapper scripts

### LINKED ISSUE #456: python3-ros-buildfarm breaks python-apt
I am not sure if this is the correct place to report this issue, but installing package `python3-ros-buildfarm` on Ubuntu (for pre-release tests) breaks python-apt.
E.g. after installation I am not able to run `update-manager` (or `ubuntu-drivers`), it quits with `AttributeError: module 'apt' has no attribute 'Cache'`. I presume this is caused by file `/usr/bin/apt.py` that gets installed with the package. Uninstalling the package resolves the error.

dirk-thomas:
Fixes #456.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -5,8 +5,14 @@ from setuptools import setup
 
 # Get a list of scripts to install
 scripts = []
-for root, _, filenames in os.walk('scripts'):
+for root, dirnames, filenames in os.walk('scripts'):
+    # don't install the wrapper scripts
+    # since they would overlay Python packages with the same name
+    if 'wrapper' in dirnames:
+        dirnames.remove('wrapper')
     for filename in filenames:
+        if not filename.endswith('.py'):
+            continue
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -5,8 +5,14 @@
 
 # Get a list of scripts to install
 scripts = []
-for root, _, filenames in os.walk('scripts'):
+for root, dirnames, filenames in os.walk('scripts'):
```
tfoote: From the docs os.walk needs `topdown=True` for the modified directories to work in all cases.

dirk-thomas: That is the default value in all Python versions: https://docs.python.org/2/library/os.html#os.walk https://docs.python.org/3/library/os.html#os.walk

tfoote: Interesting, I wonder if that was different at some time in the past. The other examples I was reading through were very explicit about calling it out.

PR #458, opened 2017-07-01
add 'Failed to stat' to the list of apt known errors

mikaelarguedas:
Not sure if this swill fix all the stretch failures we'e seen in the last two weeks but it seem to solve most of them.
Example of job that this fix allowed to run apt-update successfully after facing this error: http://build.ros.org/view/Lbin_dsv8_dSv8/job/Lbin_dsv8_dSv8__katana__debian_stretch_arm64__binary/11/console


dirk-thomas:
Thank you!

PR #459, opened 2017-07-02
print blank lines around error message

mikaelarguedas:
To match the print behavior of other wrappers (and this wrapper in [other places](https://github.com/ros-infrastructure/ros_buildfarm/compare/blank_lines_around_msg?expand=1#diff-943c6ed2c6f830b08cb10542d040aab1R128)) and make log more readable

### DIFF
--- a/ros_buildfarm/wrapper/apt.py
+++ b/ros_buildfarm/wrapper/apt.py
@@ -92,8 +92,10 @@ def call_apt_update_install_clean(
                 # retry with update command
                 continue
 
+            print('')
             print('Invocation failed due to the following known error '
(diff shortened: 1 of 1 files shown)


dirk-thomas:
The log files are already excessively long. I would rather not add any more lines to it. Due to the length searching for these keywords seems to be the most reliable approach.

mikaelarguedas:
In this case should I modify this PR to remove the extra print in the similar "invocation error" messages? It seems odd to print extra lines only for some of them but not all.
If it's not desired feel free to just close this PR and delete the branch

dirk-thomas:
Since it is symmetric to the similar message in `call_apt_repeatedly` I will merge this.

PR #460, opened 2017-07-02
Fix Debian revision (replace - by .)

jspricke:
Debian policy doesn't allow hyphens in the revision:

https://www.debian.org/doc/debian-policy/ch-controlfields.html#s-f-Version
https://manpages.debian.org/stretch/dpkg-dev/deb-version.5.html

As %z in strftime could convert to a -0800, replace it with UTC time
instead.

### DIFF
--- a/ros_buildfarm/binarydeb_job.py
+++ b/ros_buildfarm/binarydeb_job.py
@@ -108,6 +108,11 @@ def append_build_timestamp(rosdistro_name, package_name, sourcedeb_dir):
         '-m',  # keep maintainer details
         'Append timestamp when binarydeb was built.',
     ]
+    # Backwards compatibility for #460
+    # TODO(tfoote) remove when kinetic EOL
+    if rosdistro_name in ['indigo', 'kinetic', 'lunar']:
+        cmd[2] = '%s-%s' % (version, strftime('%Y%m%d-%H%M%S%z'))
+    # End backwards compatibility removal section
     print("Invoking '%s' in '%s'" % (' '.join(cmd), source_dir))
     subprocess.check_call(cmd, cwd=source_dir)
 
--- a/ros_buildfarm/binarydeb_job.py
+++ b/ros_buildfarm/binarydeb_job.py
@@ -15,7 +15,7 @@
 import os
 import subprocess
 import sys
-from time import strftime
+from time import gmtime, strftime
 import traceback
 
 from ros_buildfarm.common import get_debian_package_name
@@ -101,7 +101,7 @@ def append_build_timestamp(rosdistro_name, package_name, sourcedeb_dir):
         source_dir, ['Source', 'Version', 'Distribution', 'Urgency'])
     cmd = [
         'debchange',
-        '-v', '%s-%s' % (version, strftime('%Y%m%d-%H%M%S%z')),
+        '-v', '%s.%s' % (version, strftime('%Y%m%d.%H%M%S', gmtime())),
         '-p',  # preserve directory name
         '-D', distribution,
         '-u', urgency,
--- a/ros_buildfarm/binarydeb_job.py
+++ b/ros_buildfarm/binarydeb_job.py
@@ -104,7 +104,7 @@ def append_build_timestamp(rosdistro_name, package_name, sourcedeb_dir):
         '-v',
         '%s.%s' % (version, strftime('%Y%m%d.%H%M%S', gmtime()))
         # Backwards compatibility for #460
-        if rosdistro_name not in ('indigo', 'kinetic', 'lunar')
+        if rosdistro_name not in ('indigo', 'jade', 'kinetic', 'lunar')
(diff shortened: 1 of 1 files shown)


jspricke:
Would be great if you could rebuild all packages as this currently breaks our internal build farm.

tfoote:
Overall switching makes sense. Though the challenge of the version comparison is a problem. If we deploy this a lot of packages would no longer update automatically until we had a new upstream version even if we rebuilt.

This is because the version number is actually being parsed as:
`1.12.5-0xenial-20170613-161430` and debian inc `-800` which is higher than any truncated version. Especially the actual upstream value of 1.12.5. So switching to have the only dash at the right place will evaluate as a lower upstream version and the debian increment number is irrelevant. 

There is the epoch element we could resort to: `dpkg --compare-versions 1.12.5-0xenial-20170613-161430-0800 lt 1:1.12.5-1xenial.20170613.161430 && echo true` But it would be really nice to find another migration path.

jspricke:
* Tully Foote <notifications@github.com> [2017-07-03 01:34]:
> This is because the version number is actually being parsed as:
> `1.12.5-0xenial-20170613-161430` and debian inc `-800` which is higher than any truncated version. Especially the actual upstream value of 1.12.5. So switching to have the only dash at the right place will evaluate as a lower upstream version and the debian increment number is irrelevant. 

It's actually more problematic. As you don't comply with the policy,
different tools behave differently. For example I've seen dpkg accepting
some versions but aptitude rejecting them.

> There is the epoch element we could resort to: `dpkg --compare-versions 1.12.5-0xenial-20170613-161430-0800 lt 1:1.12.5-1xenial.20170613.161430 && echo true` But it would be really nice to find another migration path.

The epoch should only be used if upstream breaks the version numbering,
as you would have to carry it for the lifetime of the package (i.e., for
as long as there is ros-kinetic-*, which is not so long). But as we are
basically upstream, we could predict the next upstream version. For this
the usually Debian way is to construct a version like this:

<fake new upstream>+really<old upstream>-<debian revision>

so we could get something like:

1.12.5.0+really1.12.5-1xenial.20170613.161430

(still have to check that bloom doesn't produce the .0 version.)

Something like this should do:

```
from apt import Cache

apt_cache = Cache()
latest_package = apt_cache[package].versions[-1]
latest_version = latest_package.version

if latest_version.count('-') > 1 and package.version == latest_version.split('-')[0]:
    return package.version + '.0+really' + package.version

return package.version
```

This assumes that the system running the script has the broken packages
in it's apt sources. Otherwise we could compile a list of broken
versions and match against it.

dirk-thomas:
I marked the current state of the PR with the label "requires-changes" since it can't be merged as-is. If any of you @jspricke @tfoote could describe a path forward on this and update the patch accordingly that would be great.

jspricke:
As commented above, I would add a function to handle broken versions. Comments welcome.

dirk-thomas:
> I would add a function to handle broken versions

I don't know enough to implement your proposal. Please update the patch in this PR to whatever you think will work. Thanks.

jspricke:
> I don't know enough to implement your proposal.

Can you comment what you are missing, so I can explain it and we can
discuss the implementation? Thanks.

dirk-thomas:
It sounds to me that you propose to change the naming of every ROS Debian package across all distros with a new scheme. While that might be the "right" thing to do it sounds extremely invasive to me - the chances users will get confused which version of a package they are using seems pretty high. But since I don't have as much knowledge about the Debian packaging as others I will leave it up to @tfoote to work with you to decide if this can be resolved for the existing distros (and how, coming up with a working patch) or if that is something we should apply for future distros only.

tfoote:
Interesting. As I can see it clearly for m-turtle we should switch to the cleaned up logic.
And we could apply the new numbering scheme for existing repos.

In looking at how adding the extra field works. It seems anything with the `.` beats the `-` so I found this passes: 
```
$ dpkg --compare-versions 1.12.5-0xenial-20170613-161430-0800 lt 1.12.5.0-1xenial.20170613.161430 && echo true
true
```

So we could potentially even just insert the extra `.0` in the 4th position for existing active distros to get the versioning to be ordered correctly on the existing distros instead of needing to do the prefixed number. @jspricke does that make sense to you?

jspricke:
I would prefer to increase the Debian revision, as commented above, but I guess this could work as well.

tfoote:
I think that since we're already committed to the current scheme for our existing packages and this is mostly a cleanup to follow policy I think that we should not try to migrate our current rosdistros, and start supporting the cleaner format in Melodic. Adding special logic to introspect the current version and potentially do the above changes seems like it would be prone to bugs etc.

To that end I updated this PR to include a specific backwards compatibility section that reverts to the current behavior for our currently supported rosdistros. So for the next release 'melodic', it will get the new behavior.

@dirk-thomas @mikaelarguedas @clalancette

dirk-thomas:
Thank you for the patch.

jspricke:
Awesome, thanks for completing this :).

Issue #455, opened 2017-06-22
Stretch armv8 jobs fail on apt since the official release


mikaelarguedas:
Stretch armv8 jobs have been consistently failing in the past few days. It seemed to have appeared just after Stretch was released. I didn't look into the issue yet and thus I'm opening this for tracking it

```
[0mInvoking 'apt-get update'
Hit:1 http://repositories.ros.org/ubuntu/building stretch InRelease
Ign:2 http://cdn-fastly.deb.debian.org/debian stretch InRelease
Ign:3 http://cdn-fastly.deb.debian.org/debian stretch InRelease
Hit:4 http://cdn-fastly.deb.debian.org/debian stretch Release
Hit:5 http://cdn-fastly.deb.debian.org/debian stretch Release
Err:5 http://cdn-fastly.deb.debian.org/debian stretch Release
  Failed to stat - stat (2: No such file or directory)
Reading package lists...
E: The repository 'http://deb.debian.org/debian stretch Release' does no longer have a Release file.
Invocation failed without any known error condition, printing all lines to debug known error detection:
  1 'Hit:1 http://repositories.ros.org/ubuntu/building stretch InRelease'
  2 'Ign:2 http://cdn-fastly.deb.debian.org/debian stretch InRelease'
  3 'Ign:3 http://cdn-fastly.deb.debian.org/debian stretch InRelease'
  4 'Hit:4 http://cdn-fastly.deb.debian.org/debian stretch Release'
  5 'Hit:5 http://cdn-fastly.deb.debian.org/debian stretch Release'
  6 'Err:5 http://cdn-fastly.deb.debian.org/debian stretch Release'
  7 '  Failed to stat - stat (2: No such file or directory)'
  8 'Reading package lists...'
  9 'E: The repository 'http://deb.debian.org/debian stretch Release' does no longer have a Release file.'
None of the following known errors were detected:
  1 'Failed to fetch'
  2 'Hash Sum mismatch'
  3 'Unable to locate package'
  4 'is not what the server reported'
```
Example of failing job: http://build.ros.org/view/Lbin_dsv8_dSv8/job/Lbin_dsv8_dSv8__marti_data_structures__debian_stretch_arm64__binary/3/console

tfoote:
This looks like an issue with the server/cdn.

For: https://deb.debian.org/debian I get
```
Not Found

The requested URL /debian was not found on this server.

Apache Server at deb.debian.org Port 443
```

[...]

PR #461, opened 2017-07-04
hit cloudfront mirror rather than deb.debian.org

mikaelarguedas:
see https://github.com/ros-infrastructure/ros_buildfarm/issues/455#issuecomment-312924863

likely to address #455


dirk-thomas:
@mikaelarguedas Would you like to get this merged now in order to use it across all jobs or would you like to selectively test the branch for a while?

mikaelarguedas:
I'll wait for new failures to show up to validate the approach before merging. @tfoote also suggested to replace this in all templates and not only for `contrib` and `nonfree` so I may just add that as well

mikaelarguedas:
It looks that this is the only place we inject sources, everything else comes from the upstream image so I'm going to merge it as is and monitor the jobs in the following weeks

PR #463, opened 2017-07-18
make Dockerfile template more flexible

ruffsl:
Adding snippets and logic for maintaining osrf/docker_images, rework of #72.

### DIFF
--- a/ros_buildfarm/templates/snippet/from_base_image.Dockerfile.em
+++ b/ros_buildfarm/templates/snippet/from_base_image.Dockerfile.em
@@ -1,6 +1,13 @@
-@# same logic as in builder_check-docker.xml.em
-@[if arch in ['i386', 'armhf', 'arm64']]@
-FROM osrf/@(os_name)_@arch:@os_code_name
+@[if 'base_image' in locals()]@
+FROM @base_image
 @[else]@
+@# same logic as in builder_check-docker.xml.em
+@[  if arch in ['i386', 'armhf', 'arm64']]@
+FROM osrf/@(os_name)_@arch:@os_code_name
+@[  else]@
 FROM @os_name:@os_code_name
+@[  end if]@
+@[end if]@
+@[if vars().get('maintainer_name')]@
+LABEL maintainer "@maintainer_name"
 @[end if]@
--- a/ros_buildfarm/templates/__init__.py
+++ b/ros_buildfarm/templates/__init__.py
@@ -156,14 +156,15 @@ def _expand_template(template_name, **kwargs):
     interpreter.invoke('afterInclude')
 
 
-def create_dockerfile(template_name, data, dockerfile_dir):
+def create_dockerfile(template_name, data, dockerfile_dir, verbose=True):
     data['template_name'] = template_name
     data['wrapper_scripts'] = get_wrapper_scripts()
     content = expand_template(template_name, data)
     dockerfile = os.path.join(dockerfile_dir, 'Dockerfile')
     print("Generating Dockerfile '%s':" % dockerfile)
-    for line in content.splitlines():
-        print(' ', line)
+    if verbose:
+        for line in content.splitlines():
+            print(' ', line)
     with open(dockerfile, 'w') as h:
         h.write(content)
 


dirk-thomas:
In my opinion it would be good to only integrate functionality into this repo which is also being used by it. Or add functionality which is "generic" and otherwise difficult to provide externally. If specific pieces are only needed by the `osrf/docker_templates` repository it should be possible to keep them there. My rational for that is if they such docker template specific things need to change they don't require a change in this repo and can be updated / maintained independently.

Looking through the current patch quite a bit good be achieved "outside" of this repository and wouldn't strictly need to be in `ros_buildfarm`. I would argue they could be kept in e.g. the `osrf/docker_templates` repository where they are actually being used:

* `template_prefix_path` is already being exposed in order to allow the caller to customize the lookup paths. The added logic in `expand_template_prefix_path` passing a `data` dict and requiring the packages to have a `templates` subfolder seem to be very application specific.
* `create_entrypoint` could easily be called after `create_dockerfile` and could therefore live externally.
* The two snippets `add_generated_comment.Dockerfile.em` and `vcs_import.Dockerfile.em` aren't being used in the `ros_buildfarm` package. They could stay in e.g. the `osrf/docker_templates` repository where they are being used.

On the other hand some other changes could be integrated in this repo:

* The changes in `from_base_image.Dockerfile.em` make sense to make the template more flexible.
* The `verbose` option for `create_dockerfile` could be easily added from the outside (by just printing the content of the generated file) but adding that optional keyword argument would be fine as well (even though it is not being used).

I am not saying the first group of changes has no chance of being added here but currently I don't see a strong enough reason why they should be in this repo rather than where they are actually being used. Maybe you can elaborate why you think those should be added here.

ruffsl:
Ok, I've moved dockerfile/entrypoint/path logic to docker_templates, as well as the two other snippets.
See https://github.com/ruffsl/docker_templates/pull/3

I'd like to avoid the print out of dockerfiles from polluting travis logs that will use this templating functions, so I added an optional  verbose argument to avoid breaking any function calls or expected output, but if the print statement are not needed, I'd suggest omitting them.

ros_buildfarm/templates/snippet/from_base_image.Dockerfile.em
```diff
@@ -1,6 +1,15 @@
+@[if 'base_image' in locals()]@
+FROM @base_image
+@[else]@
```
dirk-thomas: With this additional nesting please add some indentation to the line 5, 7, and 9 to keep the template more readable:

```
@[  if arch in ['i386', 'armhf', 'arm64']]@
@[  else]@
@[  end if]@
```

ros_buildfarm/templates/snippet/from_base_image.Dockerfile.em
```diff
@@ -1,6 +1,15 @@
...
 @# same logic as in builder_check-docker.xml.em
 @[if arch in ['i386', 'armhf', 'arm64']]@
 FROM osrf/@(os_name)_@arch:@os_code_name
 @[else]@
 FROM @os_name:@os_code_name
 @[end if]@
+@[end if]@
+@[if 'maintainer_name' in locals()]@
+@[if maintainer_name]@
```
dirk-thomas: Please these two conditions into a single one, e.g.:

```
@[if vars().get('maintainer_name')]@
```

dirk-thomas:
> Ok, I've moved dockerfile/entrypoint/path logic to docker_templates, as well as the two other snippets.
> See ruffsl/docker_templates#3

:+1:

> I'd like to avoid the print out of dockerfiles from polluting travis logs that will use this templating functions, so I added an optional verbose argument to avoid breaking any function calls or expected output, but if the print statement are not needed, I'd suggest omitting them.

The current behavior on the buildfarm does print the content of the Dockerfile to embed that information into the job output. We shouldn't change that without a reason so the new `verbose` option looks good to me.

I will add a couple of nitpick comment to the changes to the `from_base_image.Dockerfile.em` template.

ros_buildfarm/templates/snippet/from_base_image.Dockerfile.em
ruffsl: Done

ros_buildfarm/templates/snippet/from_base_image.Dockerfile.em
ruffsl: and Done

dirk-thomas:
And merged... Thanks.

PR #471, opened 2017-08-17
increase limit of age and/or count for kept build logs for some jobs

dirk-thomas:
For these job types some more history would be good.


mikaelarguedas:
Not sure about the need for dev / pr jobs. Is it because people need to introspect PRs or commits more than 100 days old?

dirk-thomas:
For some repos with low activity 100 days is sometimes not enough. You might not see any job.

mikaelarguedas:
Sounds good

PR #472, opened 2017-08-18
Update language and job names regarding Jenkins build agents.

nuclearsandwich:
This uses the Jenkins standard term for nodes running executors across all of the buildfarm job names and node tags.

The changes require parallel changes in the buildfarm deployment since the unix username of the jenkins agent is [baked into the config](https://github.com/ros-infrastructure/ros_buildfarm/compare/update-terminology?expand=1#diff-a2efa14094821b24c4ef6f5dc244a820R140). I think ideally, we should find a way to avoid hard-coding the username since we're going to be looking soon at situations where some executors may be running as different users on the same machine.

This also changes the expected tags for the agent on master and regular build agents to... you guessed it, `agent_on_master` and `buildagent` respectively.

These changes are both made in the xenial branch of the buildfarm_deployment and buildfarm_deployment_config. Because the Jenkins master can't be reconfigured, there isn't a good automated way to backport the changes to an existing farm. Doing so would involve manually changing the username and home directory of the existing agent on master user and then manually editing the /etc/default/jenkins-slave config to use the tag `agent_on_master`.


doc/configure_customized_buildfarm.rst
```diff
@@ -70,7 +70,7 @@ Change the priority and timeouts of the Jenkins jobs
 ----------------------------------------------------
 
 You can customize the priorities of the generated jobs to fit your needs.
-Depending on the performance of your build slave you might also want yo adjust
+Depending on the performance of your build agent you might also want to adjust
```
dirk-thomas: The spelling error has been fixed on the master branch: e98c37e64ae336b8694650ed2f09152c0ad5a01f Please rebase.

nuclearsandwich:
@dirk-thomas I think we'll want to to deploy the configuration changes in this PR to the ROS buildfarm during the Xenial migration.

We've previously talked about this change requiring a major version increment. I would be fine making this change and cutting a 2.0, even if in a month we cut a 3.0 that introduces more changes still. But we could try and either avoid that or use the major version increment to make other major changes, not that I know what other breaking changes we would want.

If instead of just changing the hard-coded values we made them configurable and kept the previous values as defaults, would that keep the changes within the 1.x window and allow us to be more deliberate about a major release in the future?

dirk-thomas:
I just made a new release of this package (likely the last before the transition).

I would suggest that we use a branch (e.g. a to-be-created one named `xenial`) to deploy the new farm. That allows us to easily apply patches in the next days / weeks and once we are confident that everything "works" we can merge that into `master` and release a new version. Whenever that has happened the buildfarm can be updated to switch to the `master` branch again.

nuclearsandwich:
> I would suggest that we use a branch (e.g. a to-be-created one named xenial) to deploy the new farm. That allows us to easily apply patches in the next days / weeks and once we are confident that everything "works" we can merge that into master and release a new version. Whenever that has happened the buildfarm can be updated to switch to the master branch again.

This sounds great to me. It also brings up a question I had about whether the production buildfarm is pinned to the latest release or master. The individual job types periodically reconfigure on a timer, is there a periodic job that re-configures everything or does that only happen manually.

More precisely: what steps for changing the production buildfarm branch to xenial (and then back again after a suitable time period) should I add to my migration playbook?

dirk-thomas:
> It also brings up a question I had about whether the production buildfarm is pinned to the latest release or master.

The official ROS buildfarm is tracking `master`. I would suggest the new farm to track `xenial`. While we recommend others to use a tagged version this allows us to apply changes quickly.

> The individual job types periodically reconfigure on a timer, is there a periodic job that re-configures everything or does that only happen manually.

The management jobs are only touched by invoking the `generate_all_jobs.py` script. All other jobs are automatically been regenerated by these management jobs.

> what steps for changing the production buildfarm branch to xenial (and then back again after a suitable time period) should I add to my migration playbook?

I modified the version field in 3613f0c45cd5a73f7e61e14392cbfea46bb5790c. When you checkout the `xenial` branch and invoke the `generate_all_jobs.py` script from there this will make sure that all management jobs use the `xenial` branch and also generate / update all other jobs using that branch.

So once all changes (except this commit) have been merged from the `xenial` branch into `master` it is sufficient to call the `generate_all_jobs.py` script to update the management jobs. During the next nightly regeneration all other jobs will switch over to use `master`.

PR #465, opened 2017-07-24
Improve docs to remove obsolete jobs

dirk-thomas:
Fixes #464.

Ready for review.

PR #473, opened 2017-08-29
improve ongoing operations docs

dirk-thomas:
The fist commit only fixes style problems from #465.

The second commit mentions the management jobs which need to be disabled / deleted.

### DIFF
--- a/doc/ongoing_operations.rst
+++ b/doc/ongoing_operations.rst
@@ -112,30 +112,30 @@ Usually you want to disable the jobs first, wait a little bit in case you need t
 Disable all jobs related to a specific target
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-Assuming that the ROS distribution is called `lunar` and the platform is `Ubuntu Yakkety` you can disable the jobs with the following prefixes:
-* `Lsrc_uY__` which matches the Lunar source jobs for Ubuntu Yakkety.
-* `Lbin_uY64__` which matches the Lunar binary jobs for Ubuntu Yakkety for the `amd64` architecture.
-* ... add additional prefixes for other architectures.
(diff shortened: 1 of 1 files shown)


doc/ongoing_operations.rst
```diff
@@ -112,30 +112,31 @@ Usually you want to disable the jobs first, wait a little bit in case you need t
...
 
-Assuming that the ROS distribution is called `lunar` and the platform is `Ubuntu Yakkety` you can disable the jobs with the following prefixes:
-* `Lsrc_uY__` which matches the Lunar source jobs for Ubuntu Yakkety.
-* `Lbin_uY64__` which matches the Lunar binary jobs for Ubuntu Yakkety for the `amd64` architecture.
-* ... add additional prefixes for other architectures.
+Assuming that the ROS distribution is called ``lunar`` and the platform is ``Ubuntu Yakkety`` you can disable the jobs with the following prefixes:
+ * ``Lsrc_uY__`` which matches the Lunar source jobs for Ubuntu Yakkety.
+ * ``Lbin_uY64__`` which matches the Lunar binary jobs for Ubuntu Yakkety for the ``amd64`` architecture.
+ * ... add additional prefixes for other architectures.
+ * ``Lrel_sync-packages-to-testing_yakkety_`` which matches the management jobs to sync package for Ubuntu Yakkety.
```
mikaelarguedas: Is seems inconsistent to pin down architectures for the binary jobs but not the management jobs. Would it be better to either:
- target all architectures:
```
``Lbin_uY``
``Lrel_sync-packages-to-testing_yakkety_``
```

- or specific ones:
```
``Lbin_uY64__``
``Lrel_sync-packages-to-testing_yakkety_amd64``
... add additional prefixes for other architectures
```

dirk-thomas: `Lbin_uY` seems dangerous since it doesn't enforce the double underscore separator afterwards. In the future a job like `Lbin_uYa` would be matched too.

I updated the patch to use the full management job name instead as in your second option.

mikaelarguedas: :+1:

PR #474, opened 2017-09-05
Add initial version of upload trigger job generators.

nuclearsandwich:
This adds a template and generating script for the upload trigger jobs.
On the current farm as `upload_main` and `upload_testing`.

The first commit only creates the generating script. It doesn't add the script to `generate_all_jobs.py` or check any configuration keys for whether or not the jobs should be created. I also need to figure out the right way to check all combinations of the sync-testing jobs since they're distro+arch specific.

The scripts themselves are part of a puppet branch that is waiting to be published. I've asked @tfoote to be the second set of eyes verifying there's nothing sensitive in the script contents itself before publishing the puppet changes that will place these scripts on repo hosts.

There's some formatting of the template still to iron out but better to get the overall setup reviewed early than spend time scrutinizing something that needs fundamental changes.

### DIFF
--- a/ros_buildfarm/templates/release/trigger_upload_repo_job.xml.em
+++ b/ros_buildfarm/templates/release/trigger_upload_repo_job.xml.em
@@ -34,7 +34,7 @@
     <hudson.tasks.Shell>
       <command>#!/bin/bash
 $HOME/upload_triggers/upload_repo.bash @repo
-    </command>
+</command>
     </hudson.tasks.Shell>
   </builders>
   <publishers/>
--- a/ros_buildfarm/templates/release/trigger_upload_repo_job.xml.em
+++ b/ros_buildfarm/templates/release/trigger_upload_repo_job.xml.em
@@ -16,7 +16,7 @@
   <canRoam>false</canRoam>
   <disabled>false</disabled>
   <blockBuildWhenDownstreamBuilding>false</blockBuildWhenDownstreamBuilding>
-  <blockBuildWhenUpstreamBuilding>true</blockBuildWhenUpstreamBuilding>
+  <blockBuildWhenUpstreamBuilding>@block_when_upstream_building</blockBuildWhenUpstreamBuilding>
   <triggers>
     <jenkins.triggers.ReverseBuildTrigger>
       <spec/>
--- a/ros_buildfarm/templates/release/trigger_upload_repo_job.xml.em
+++ b/ros_buildfarm/templates/release/trigger_upload_repo_job.xml.em
@@ -1,0 +1,42 @@
+<project>
+  <actions/>
+  <description/>
+  <keepDependencies>false</keepDependencies>
+  <properties>
+    <jenkins.advancedqueue.priority.strategy.PriorityJobProperty plugin="PrioritySorter@@3.5.0">
+      <useJobPriority>false</useJobPriority>
+      <priority>-1</priority>
+    </jenkins.advancedqueue.priority.strategy.PriorityJobProperty>
+    <org.jenkinsci.plugins.requeuejob.RequeueJobProperty plugin="jobrequeue@@1.1">
+      <requeueJob>false</requeueJob>
+    </org.jenkinsci.plugins.requeuejob.RequeueJobProperty>
+  </properties>
+  <scm class="hudson.scm.NullSCM"/>
+  <assignedNode>building_repository</assignedNode>
+  <canRoam>false</canRoam>
+  <disabled>false</disabled>
+  <blockBuildWhenDownstreamBuilding>false</blockBuildWhenDownstreamBuilding>
+  <blockBuildWhenUpstreamBuilding>true</blockBuildWhenUpstreamBuilding>
+  <triggers>
+    <jenkins.triggers.ReverseBuildTrigger>
+      <spec/>
+      <upstreamProjects>@upstream_job_names</upstreamProjects>
+      <threshold>
+        <name>SUCCESS</name>
+        <ordinal>0</ordinal>
+        <color>BLUE</color>
+        <completeBuild>true</completeBuild>
+      </threshold>
+    </jenkins.triggers.ReverseBuildTrigger>
+  </triggers>
+  <concurrentBuild>false</concurrentBuild>
+  <builders>
+    <hudson.tasks.Shell>
+      <command>#!/bin/bash
+$HOME/upload_triggers/upload_repo.bash @repo
+    </command>
+    </hudson.tasks.Shell>
+  </builders>
+  <publishers/>
+  <buildWrappers/>
+</project>
--- a/scripts/release/generate_release_trigger_upload_jobs.py
+++ b/scripts/release/generate_release_trigger_upload_jobs.py
@@ -43,7 +43,7 @@ def main(argv=sys.argv[1:]):
         job_name = 'upload_%s' % repo
         upstream_job_names = ['{0}_sync-packages-to-{1}'.format(
             get_release_job_prefix(rosdistro), repo) for rosdistro in distributions]
-        upstream_job_names = ', '.join(upstream_job_names)
+        upstream_job_names = ', '.join(sorted(upstream_job_names))
         job_config = expand_template(template_name, {
             'repo': repo,
             'upstream_job_names': upstream_job_names})
--- a/scripts/release/generate_release_trigger_upload_jobs.py
+++ b/scripts/release/generate_release_trigger_upload_jobs.py
@@ -19,8 +19,10 @@ import sys
 
 from ros_buildfarm.argument import add_argument_config_url
 from ros_buildfarm.argument import add_argument_dry_run
+from ros_buildfarm.common import JobValidationError
 from ros_buildfarm.common import get_release_job_prefix
 from ros_buildfarm.config import get_index
+from ros_buildfarm.config import get_release_build_files
 from ros_buildfarm.jenkins import configure_job
 from ros_buildfarm.jenkins import connect
 from ros_buildfarm.templates import expand_template
@@ -37,22 +39,49 @@ def main(argv=sys.argv[1:]):
 
     config = get_index(args.config_url)
     jenkins = connect(config.jenkins_url)
-    distributions = config.distributions.keys()
 
     for repo in ['main', 'testing']:
         job_name = 'upload_%s' % repo
         block_when_upstream_building = 'true'
         if repo == 'testing':
             block_when_upstream_building = 'false'
-        upstream_job_names = ['{0}_sync-packages-to-{1}'.format(
-            get_release_job_prefix(rosdistro), repo) for rosdistro in distributions]
-        upstream_job_names = ', '.join(sorted(upstream_job_names))
         job_config = expand_template(template_name, {
             'block_when_upstream_building': block_when_upstream_building,
             'repo': repo,
-            'upstream_job_names': upstream_job_names})
+            'upstream_job_names': get_upstream_job_names(config, repo)})
 
         configure_job(jenkins, job_name, job_config, dry_run=args.dry_run)
+
+
+def get_upstream_job_names(config, repo):
+    distributions = config.distributions.keys()
+    if repo == 'main':
+        upstream_job_names = ['{0}_sync-packages-to-{1}'.format(
+            get_release_job_prefix(rosdistro), repo) for rosdistro in distributions]
+    elif repo == 'testing':
+        upstream_job_names = []
+        for rosdistro in distributions:
+            architectures_by_code_name = {}
+            build_files = get_release_build_files(config, rosdistro)
+            for k, build_file in build_files.items():
+                for os_name in build_file.targets.keys():
+                    for code_name, architectures in build_file.targets[os_name].items():
+                        architectures_by_code_name[code_name] = \
+                            set(architectures_by_code_name.get(code_name, set()).union(
+                                set(architectures.keys())))
+
+            for code_name, archs in architectures_by_code_name.items():
+                for arch in archs:
+                    upstream_job_names.append(
+                        '{prefix}_sync-packages-to-{repo}_{codename}_{arch}'.format(
+                            prefix=get_release_job_prefix(rosdistro),
+                            repo=repo,
+                            codename=code_name,
+                            arch=arch))
+    else:
+        raise JobValidationError("Unknown upstream jobs for job 'upload_{}'." % repo)
+    upstream_job_names = ','.join(sorted(upstream_job_names))
+    return upstream_job_names
 
 
 if __name__ == '__main__':
--- a/scripts/release/generate_release_trigger_upload_jobs.py
+++ b/scripts/release/generate_release_trigger_upload_jobs.py
@@ -67,8 +67,8 @@ def get_upstream_job_names(config, repo):
                 for os_name in build_file.targets.keys():
                     for code_name, architectures in build_file.targets[os_name].items():
                         architectures_by_code_name[code_name] = \
-                            architectures_by_code_name.get(code_name, set()) \
-                            | set(architectures.keys())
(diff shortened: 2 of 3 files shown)


ros_buildfarm/templates/release/trigger_upload_repo_job.xml.em
```diff
@@ -0,0 +1,42 @@
...
+    <org.jenkinsci.plugins.requeuejob.RequeueJobProperty plugin="jobrequeue@@1.1">
+      <requeueJob>false</requeueJob>
+    </org.jenkinsci.plugins.requeuejob.RequeueJobProperty>
+  </properties>
+  <scm class="hudson.scm.NullSCM"/>
+  <assignedNode>building_repository</assignedNode>
+  <canRoam>false</canRoam>
+  <disabled>false</disabled>
+  <blockBuildWhenDownstreamBuilding>false</blockBuildWhenDownstreamBuilding>
+  <blockBuildWhenUpstreamBuilding>true</blockBuildWhenUpstreamBuilding>
```
dirk-thomas: This seems to be `false` for the existing `upload_testing` job deployed on Jenkins. The question is is there reason for that on Jenkins?

scripts/release/generate_release_trigger_upload_jobs.py
```diff
@@ -0,0 +1,55 @@
...
+
+    config = get_index(args.config_url)
+    jenkins = connect(config.jenkins_url)
+    distributions = config.distributions.keys()
+
+    for repo in ['main', 'testing']:
+        job_name = 'upload_%s' % repo
+        upstream_job_names = ['{0}_sync-packages-to-{1}'.format(
+            get_release_job_prefix(rosdistro), repo) for rosdistro in distributions]
+        upstream_job_names = ', '.join(upstream_job_names)
```
dirk-thomas: These should be sorted in order to make the generated result deterministic.

nuclearsandwich:
> Once the upload_testing job has the right upstream projects this looks ok to me.

Do you think this will be worth merging as-is, which would have us running the script standalone rather than integrating it with generate_all_jobs.py and a ros_buildfarm_config key or do we want to make those other changes before merging as well?

We do want those changes soon, it's a matter of what to prioritizing before tomorrow and the end of the week.

dirk-thomas:
> > Once the upload_testing job has the right upstream projects this looks ok to me.

> Do you think this will be worth merging as-is, which would have us running the script standalone rather than integrating it with generate_all_jobs.py and a ros_buildfarm_config key or do we want to make those other changes before merging as well?

I am fine with running the new script "standalone" for now. It doesn't need to be integrated with `generate_all_jobs.py`.

But as far as I can see it currently generates the wrong upstream jobs for the `upload_testing` job which is not ok. The job wouldn't have the right upstream jobs configured and therefore wouldn't work correctly.

Also the pending question / comment should be addressed before merging (both currently show up in a dry-run compared to the existing jobs).

nuclearsandwich:
> But as far as I can see it currently generates the wrong upstream jobs for the upload_testing job which is not ok. The job wouldn't have the right upstream jobs configured and therefore wouldn't work correctly.

Yep, I'm addressing that now. They are indeed incorrect as is.



> Also the pending question / comment should be addressed before merging (both currently show up in a dry-run compared to the existing jobs)

Yes. This definitely isn't ready to merge as-is. I was making sure that I knew what needed to be done in order to do all the work at once rather than context switch.

ros_buildfarm/templates/release/trigger_upload_repo_job.xml.em
tfoote: Because the testing upload is triggered by the sync which is triggered by the builds automatically if this waited for upstream it could only upload to testing when the farm had gone completely idle.

For the upload_main it's ok to have it block since it's only one level of dependency on the manually triggered sync-to-main jobs. So it will stop it from running if there's any sync-to-main job queued. This makes it a little safer if two sync-to-main-jobs get triggered close together, you'll get the aggregate uploaded instead of potentially 2 uploads. It should be safe to do it either way as they're all locked to the same executor. I was mostly looking to avoid the number of times that the public mirror needed to sync since we have seen issues with that and have little to no introspection into it's state.

nuclearsandwich: Addressing this as well.

nuclearsandwich:
With fcdbe13 this PR now blocks on upstream for main only and correctly generates the upstream jobs for testing with a pretty egregious enumeration. One thing to note is that not all upstream jobs on the current farm appear in the new job list because they're jobs for Ubuntu versions that are no longer part of the production config like saucy.

It's hard to test this right now while the buildfarm is offline.

scripts/release/generate_release_trigger_upload_jobs.py
```diff
@@ -0,0 +1,88 @@
...
+                for arch in archs:
+                    upstream_job_names.append(
+                        '{prefix}_sync-packages-to-{repo}_{codename}_{arch}'.format(
+                            prefix=get_release_job_prefix(rosdistro),
+                            repo=repo,
+                            codename=code_name,
+                            arch=arch))
+    else:
+        raise JobValidationError("Unknown upstream jobs for job 'upload_{}'." % repo)
+    upstream_job_names = ','.join(sorted(upstream_job_names))
```
dirk-thomas: No need for the variable `upstream_job_names`. This can `return ', '.join(...)` directly.

scripts/release/generate_release_trigger_upload_jobs.py
```diff
@@ -0,0 +1,88 @@
...
+
+
+def get_upstream_job_names(config, repo):
+    distributions = config.distributions.keys()
+    if repo == 'main':
+        upstream_job_names = ['{0}_sync-packages-to-{1}'.format(
+            get_release_job_prefix(rosdistro), repo) for rosdistro in distributions]
+    elif repo == 'testing':
+        upstream_job_names = []
+        for rosdistro in distributions:
```
dirk-thomas: Instead of creating the variable `distributions` this should use `config.distributions.keys()` directly. Otherwise it is not obvious that the `if` block as well as the `elif` block use the same generator.

Also replace the usage a few lines above and then remove the variable `distributions`.

scripts/release/generate_release_trigger_upload_jobs.py
```diff
@@ -0,0 +1,88 @@
...
+                            set(architectures_by_code_name.get(code_name, set()).union(
+                                set(architectures.keys())))
+
+            for code_name, archs in architectures_by_code_name.items():
+                for arch in archs:
+                    upstream_job_names.append(
+                        '{prefix}_sync-packages-to-{repo}_{codename}_{arch}'.format(
+                            prefix=get_release_job_prefix(rosdistro),
+                            repo=repo,
+                            codename=code_name,
```
dirk-thomas: Please use one way of naming consistently. The remaining code use `code_name` instead of `codename`.

scripts/release/generate_release_trigger_upload_jobs.py
```diff
@@ -0,0 +1,88 @@
...
+    distributions = config.distributions.keys()
+    if repo == 'main':
+        upstream_job_names = ['{0}_sync-packages-to-{1}'.format(
+            get_release_job_prefix(rosdistro), repo) for rosdistro in distributions]
+    elif repo == 'testing':
+        upstream_job_names = []
+        for rosdistro in distributions:
+            architectures_by_code_name = {}
+            build_files = get_release_build_files(config, rosdistro)
+            for k, build_file in build_files.items():
```
dirk-thomas: Is `k` being used anywhere?

scripts/release/generate_release_trigger_upload_jobs.py
```diff
@@ -0,0 +1,88 @@
...
+                for os_name in build_file.targets.keys():
+                    for code_name, architectures in build_file.targets[os_name].items():
+                        architectures_by_code_name[code_name] = \
+                            set(architectures_by_code_name.get(code_name, set()).union(
```
dirk-thomas: Instead of calling `union` this should use `set(...) | set(...)` which makes it clearer that a new set with the union of both is desired.

scripts/release/generate_release_trigger_upload_jobs.py
nuclearsandwich: Nope, is there a throwaway binding like `_` that should be used?

dirk-thomas: > is there a throwaway binding like `_` that should be used?

Exactly: `_`

scripts/release/generate_release_trigger_upload_jobs.py
nuclearsandwich: To be clear, your preference here is that the two cases do not share work if they are able?

nuclearsandwich:
The latest changes address all feedback but the one I had a question about.

The produced diffs are below. In both cases the changes to the executed command reflect the updated command for the new deployment. and the change to the `<command>` line is a trailing whitespace fix.

```
Updating job 'upload_main' (dry run)
    <<<
    --- remote config
    +++ new config
    @@ -23 +23 @@
    -      <upstreamProjects>Irel_sync-packages-to-main,Jrel_sync-packages-to-main,Krel_sync-packages-to-main,Lrel_sync-packages-to-main, </upstreamProjects>
    +      <upstreamProjects>Irel_sync-packages-to-main,Jrel_sync-packages-to-main,Krel_sync-packages-to-main,Lrel_sync-packages-to-main</upstreamProjects>
    @@ -35,2 +35,2 @@
    -      <command>#!/bin/bash 
    -/home/jenkins-slave/upload_triggers/upload_main.bash
    +      <command>#!/bin/bash
    +$HOME/upload_triggers/upload_repo.bash main
    >>>
```

The change to the dependency line removes a trailing comma+space which I believe to be an artifact of the manual entry rather than an mandate of the web editors output. I'll change it if I turn out wrong.

```
Updating job 'upload_testing' (dry run)
    <<<
    --- remote config
    +++ new config
    @@ -23 +23 @@
    -      <upstreamProjects>Irel_sync-packages-to-testing_saucy_amd64,Irel_sync-packages-to-testing_saucy_i386,Irel_sync-packages-to-testing_trusty_amd64,Irel_sync-packages-to-testing_trusty_armhf,Irel_sync-packages-to-testing_trusty_i386,Jrel_sync-packages-to-testing_trusty_amd64,Jrel_sync-packages-to-testing_trusty_armhf,Jrel_sync-packages-to-testing_trusty_i386,Jrel_sync-packages-to-testing_utopic_amd64,Jrel_sync-packages-to-testing_utopic_i386,Jrel_sync-packages-to-testing_vivid_amd64,Jrel_sync-packages-to-testing_vivid_i386,Krel_sync-packages-to-testing_wily_amd64,Krel_sync-packages-to-testing_wily_i386,Krel_sync-packages-to-testing_xenial_amd64,Krel_sync-packages-to-testing_xenial_armhf,Krel_sync-packages-to-testing_xenial_i386,Krel_sync-packages-to-testing_wily_amd64,Krel_sync-packages-to-testing_wily_i386,Krel_sync-packages-to-testing_xenial_amd64,Krel_sync-packages-to-testing_xenial_armhf,Krel_sync-packages-to-testing_xenial_i386,Krel_sync-packages-to-testing_jessie_amd64,Krel_sync-packages-to-testing_jessie_arm64, Lrel_sync-packages-to-testing_stretch_amd64, Lrel_sync-packages-to-testing_stretch_arm64, Lrel_sync-packages-to-testing_xenial_amd64, Lrel_sync-packages-to-testing_xenial_arm64, Lrel_sync-packages-to-testing_xenial_armhf, Lrel_sync-packages-to-testing_yakkety_amd64, Lrel_sync-packages-to-testing_zesty_amd64,</upstreamProjects>
    +      <upstreamProjects>Irel_sync-packages-to-testing_trusty_amd64,Irel_sync-packages-to-testing_trusty_armhf,Irel_sync-packages-to-testing_trusty_i386,Jrel_sync-packages-to-testing_trusty_amd64,Jrel_sync-packages-to-testing_trusty_armhf,Jrel_sync-packages-to-testing_trusty_i386,Krel_sync-packages-to-testing_jessie_amd64,Krel_sync-packages-to-testing_jessie_arm64,Krel_sync-packages-to-testing_xenial_amd64,Krel_sync-packages-to-testing_xenial_arm64,Krel_sync-packages-to-testing_xenial_armhf,Krel_sync-packages-to-testing_xenial_i386,Lrel_sync-packages-to-testing_stretch_amd64,Lrel_sync-packages-to-testing_stretch_arm64,Lrel_sync-packages-to-testing_xenial_amd64,Lrel_sync-packages-to-testing_xenial_arm64,Lrel_sync-packages-to-testing_xenial_armhf,Lrel_sync-packages-to-testing_zesty_amd64</upstreamProjects>
    @@ -35,2 +35,2 @@
    -      <command>#!/bin/bash 
    -/home/jenkins-slave/upload_triggers/upload_testing.bash
    +      <command>#!/bin/bash
    +$HOME/upload_triggers/upload_repo.bash testing
    >>>
```

There are a different set of sync jobs in the before and after because the generated config relies on what's currently in the ros_buildfarm_config and we no longer have configs in place for saucy, utopic, vivid, wily, and yakkety and other eol distros.

scripts/release/generate_release_trigger_upload_jobs.py
dirk-thomas: The reuse of the same generator seems to be fine (I just had to convince myself in an interactive Python shell).

This can stay as-i. Sorry for the noise.

scripts/release/generate_release_trigger_upload_jobs.py
```diff
@@ -0,0 +1,87 @@
...
+    elif repo == 'testing':
+        upstream_job_names = []
+        for rosdistro in distributions:
+            architectures_by_code_name = {}
+            build_files = get_release_build_files(config, rosdistro)
+            for build_file in build_files.values():
+                for os_name in build_file.targets.keys():
+                    for code_name, architectures in build_file.targets[os_name].items():
+                        architectures_by_code_name[code_name] = \
+                            architectures_by_code_name.get(code_name, set()) \
```
dirk-thomas: Our existing code keeps the operator at the end of this line (rather than at the beginning of the next line).

nuclearsandwich:
With outstanding feedback addressed I'm going to merge this into the xenial branch if CI passes.

The next stage of this is adding it to the `generate_all_jobs.py` script and adding a key in ros_buildfarm_config that only creates these jobs for our production buildfarm.

dirk-thomas:
What is the status of this PR?

nuclearsandwich:
> What is the status of this PR?

It was previously blocked on the CI failures that are now fixed by #476. I've rebased this PR and its target xenial branch to incorporate that fix.

Since we have some time before the xenial branch merges upstream. Where's the best place to add documentation for this new command, mostly to instruct users who aren't maintaining build.ros.org itself that they don't need to use it.

dirk-thomas:
> Where's the best place to add documentation for this new command, mostly to instruct users who aren't maintaining build.ros.org itself that they don't need to use it.

This repo has a `doc` folder which contains the documentation. I would say that is the best place to add the information. So it can just be added to this PR.

nuclearsandwich:
Added an entry to https://github.com/ros-infrastructure/ros_buildfarm/blob/27e8561658858fccc665edb67caabe690a36b558/doc/jobs/release_jobs.rst for this job.

> __generate_release_trigger_upload_jobs.py__ generates a pair of jobs that trigger an upload to packages.ros.org. Only needed for the official buildfarm at build.ros.org.

nuclearsandwich:
@dirk-thomas if you don't have any additional feedback on the documentation this is ready to merge into the xenial branch.

dirk-thomas:
My approval still stands. This can be merged any time.

PR #476, opened 2017-10-02
Fix builds using pip3 alongside Python 2.7

nuclearsandwich:
Two builds on Travis started failing with a recent platform update. To get the build working again I've set it to install the system pip3 and use sudo when installing the python3 version of EmPy.

We could also just install python3-empy from apt. I don't have a preference either way.

I was testing these configuration changes with custom builds. Here's a link to the [most recent one](https://travis-ci.org/ros-infrastructure/ros_buildfarm/builds/282492141) and another will be part of this PR. Hopefully it too succeeds.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -164,6 +164,7 @@ matrix:
         - docker
       env: JOB_TYPE=prerelease UNDERLAY_REPOSITORY_NAMES="roscpp_core" OVERLAY_PACKAGE_NAMES=roscpp
       before_script:
+        - pip3 install EmPy
         # install catkin for test results
         - sudo sh -c 'echo "deb http://packages.ros.org/ros/ubuntu $(lsb_release -sc) main" > /etc/apt/sources.list.d/ros.list'
         - sudo apt-key adv --keyserver hkp://ha.pool.sks-keyservers.net:80 --recv-key 421C365BD9FF1F717815A3895523BAEEB01FA116
@@ -184,6 +185,7 @@ matrix:
         - docker
       env: JOB_TYPE=external_prerelease
       before_script:
+        - pip3 install EmPy
         - python setup.py install
         - mkdir job && cd job
         - ln -s .. ros_buildfarm
--- a/.travis.yml
+++ b/.travis.yml
@@ -164,7 +164,6 @@ matrix:
         - docker
       env: JOB_TYPE=prerelease UNDERLAY_REPOSITORY_NAMES="roscpp_core" OVERLAY_PACKAGE_NAMES=roscpp
       before_script:
-        - pip3 install EmPy
         # install catkin for test results
         - sudo sh -c 'echo "deb http://packages.ros.org/ros/ubuntu $(lsb_release -sc) main" > /etc/apt/sources.list.d/ros.list'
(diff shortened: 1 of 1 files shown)


nuclearsandwich:
> I think this is effectively forcing python3 to be installed in addition to the one that travis is putting into the environment. If we want to be uisng python 3 shouldn't we switch the above instead? Or if we're supposed to be getting coverage using 2.7 we should be installing the python2 version of empy not the python3 version.

I will confess a significant degree of ignorance as to the precise mechanics. My understanding is that some of the buildfarm scripts are hard-coded to run python3 even in python2 contexts and these tests cover those situations. Prior to a recent platform change, a pip3 binary was present even in the python2.7 image. The changes here restore that behavior without visiting whether or not the behavior is sensible or correct.

tfoote:
If that's the case it's fine. Though if we can I'd suggest using the deb `python3-empy`. It will be more reproducible/audit-able when/if there's a change.

nuclearsandwich:
> Though if we can I'd suggest using the deb python3-empy. It will be more reproducible/audit-able when/if there's a change.

Yep, that would work too. I'll wait for @dirk-thomas's input before switching anything up.

.travis.yml
```diff
@@ -164,7 +164,8 @@ matrix:
         - docker
       env: JOB_TYPE=prerelease UNDERLAY_REPOSITORY_NAMES="roscpp_core" OVERLAY_PACKAGE_NAMES=roscpp
       before_script:
-        - pip3 install EmPy
+        - sudo apt-get install -y python3-pip
```
dirk-thomas: This needs to call `apt-get update` before in order to reliably install the package.

dirk-thomas:
> Though if we can I'd suggest using the deb python3-empy. It will be more reproducible/audit-able when/if there's a change.

All dependencies are being installed using `pip` therefore I don't think this single one should be different. Also I don't see how it would make it more audit-able - in both cases the exact version installed is logged in the output.

nuclearsandwich:
@dirk-thomas since this PR only touches the Travis data it doesn't have any effect on the production farm but is there a documented process or checklist for testing changes on the buildfarm before merging PRs in this project?

dirk-thomas:
> is there a documented process or checklist for testing changes on the buildfarm before merging PRs in this project?

There is no checklist for it. The common rules apply: test the patch as well as possible before (either locally or on a test farm), if necessary run some builds to convince yourself that everything still works as desired after merging.

In this case merging will be just fine since it doesn't affect the operation.

dirk-thomas:
Please squash commits before / during the merge. I have force pushed to master to squash the six separate commits into a single one: 631b814358d794d53e08e4bbce7cb2f404ece0d0

nuclearsandwich:
> Please squash commits before / during the merge. I have force pushed to master to squash the six separate commits into a single one: 631b814

This policy isn't documented in the CONTRIBUTING.md. Can I fix that?

dirk-thomas:
> This policy isn't documented in the CONTRIBUTING.md.

We apply this policy in general to each and every repo since it doesn't make sense to have 5+ commits for a small change like this. Basically you will "pay" for every commit in the future when writing changelogs, reviewing changes, tracing changes over time, etc.

> Can I fix that?

Please feel free to do that if you think it is necessary. Imo it doesn't make sense to mention it in every repo though. Instead we could reference an existing tutorial which covers "guidelines for writing good pull requests" which contain many more items beside this.

PR #477, opened 2017-10-23
Update plugin versions for current buildfarm.

nuclearsandwich:
PrioritySorter  -> 3.5.1
analysis-core   -> 1.92
dashboard-view  -> 2.9.11
git             -> 3.5.1
github          -> 1.28.0
mercurial       -> 2.1
script-security -> 1.34
warnings        -> 4.63

Using log output from a sampling of reconfigure jobs. I have the beginnings of some shell scripts to make this easier. They're not quite working yet.

### DIFF
--- a/ros_buildfarm/templates/dashboard_view_all_jobs.xml.em
+++ b/ros_buildfarm/templates/dashboard_view_all_jobs.xml.em
@@ -1,4 +1,4 @@
-<hudson.plugins.view.dashboard.Dashboard plugin="dashboard-view@@2.9.10">
+<hudson.plugins.view.dashboard.Dashboard plugin="dashboard-view@@2.9.11">
   <name>@view_name</name>
   <description>Generated at @ESCAPE(now_str) from template '@ESCAPE(template_name)'</description>
   <filterExecutors>false</filterExecutors>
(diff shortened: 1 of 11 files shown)


nuclearsandwich:
> Beside the changes in the xml it also seems that the CMake warning parser is being removed?

The missing and thus removed CMake warning parser appears to have been due to an over-escaped quote in the configured CMake warning parser when it was first ported from the old buildfarm and this has been fixed.

I don't see any other persistent changes in the warnings structure. I do see different representations using self closing tags `<foobar/>` or `<foobar></foobar>` but it doesn't seem to be consistent and doesn't generate diff noise during reconfiguration.

dirk-thomas:
> I don't see any other persistent changes in the warnings structure.

Please double check each job type again. I just checked [this](http://build.ros.org/view/Ldev/job/Ldev__ackermann_msgs__ubuntu_xenial_amd64/) dev job which was using:

```
<hudson.plugins.warnings.WarningsPublisher plugin="warnings@4.62">
```

After saving the job config using the web interface the config is updated to:

```
<hudson.plugins.warnings.WarningsPublisher plugin="warnings@4.63">
```

but there is an additional tag inserted:

```
<categoriesPattern/>
```

nuclearsandwich:
> but there is an additional tag inserted:
> 
> <categoriesPattern/>

Ah yeah I missed that when scanning thinking it was a self-closing vs tag-pair change like the others. The update to 4.63 is part of this PR already but I'll add the categoriesPattern tag to the templates as well.

As an aside, I dislike that GitHub's new PR status symbol in the favicon takes CI into account but not review as I keep thinking this magically became approved.

dirk-thomas:
Can this branch be removed?

PR #479, opened 2017-10-31
Merge the changes from Xenial into master.

nuclearsandwich:
This updates the master branch to reflect the changes deployed as part of the xenial migrate. Includes

- #477 
- #474 
- #472 

IMO this pull request should either be fast-forwarded or regular merged but not squashed as commit tidying was performed in the xenial-targeted PRs.


dirk-thomas:
The second commit should probably not be part of this PR?

nuclearsandwich:
> The second commit should probably not be part of this PR?

I wanted to ask about that. I could create a temporary branch and drop it, Then the history of master would never reflect the time that the buildfarm spent tracking a non master branch. That only bothers me a little from a forensic perspective but if it doesn't bother you then I can do so.

dirk-thomas:
I would argue that looking at the history of the `ros_buildfarm/__init__.py` / `setup.py` file on *master* it doesn't make much sense to show the "xenial" state anywhere. Therefore I would suggest to create branch without that commit and merge a new PR from there. You could also cherry-pick the three other commits directly to master but that would leave them without a reference to this PR which is probably not as informative.

nuclearsandwich:
Closing in favor of #480.

nuclearsandwich:
Do not delete the `xenial` branch until build.ros.org has been reconfigured. In fact it would probably be nice to push a commit to the `xenial` branch reverting its configuration to master and leaving it until 19 Nov. Which is what I'm doing for the deployment scripts.

PR #481, opened 2017-11-02
increase days_to_keep for some job types

dirk-thomas:
Since some job only run rarely and a new build after e.g. nine months removes all previous entries.

### DIFF
--- a/ros_buildfarm/templates/release/binarydeb_job.xml.em
+++ b/ros_buildfarm/templates/release/binarydeb_job.xml.em
@@ -9,7 +9,7 @@ but disabled since the package is blacklisted (or not whitelisted) in the config
   <properties>
 @(SNIPPET(
     'property_log-rotator',
-    days_to_keep=180,
+    days_to_keep=730,
     num_to_keep=30,
 ))@
 @(SNIPPET(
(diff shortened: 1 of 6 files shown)


ros_buildfarm/templates/release/binarydeb_job.xml.em
```diff
@@ -9,7 +9,7 @@ but disabled since the package is blacklisted (or not whitelisted) in the config
   <properties>
 @(SNIPPET(
     'property_log-rotator',
-    days_to_keep=180,
+    days_to_keep=730,
```
mikaelarguedas: Was there a reason for this to be shorter than the other ones ?

dirk-thomas: Which ones are you referring to with "the other ones"? It had and has the same value as the corresponding sourcedeb job.

mikaelarguedas: That was not clearly formulated: was there a reason for source jobs and binary jobs to have shorter times than the other jobs updated in this PR?

dirk-thomas: Not really. Some job types have even less days. For each type we are just "eyeballing" what might make sense :stuck_out_tongue_winking_eye:

mikaelarguedas: sounds good, thanks

nuclearsandwich:
I didn't end up commenting when this was under review because after examining it there were no issues. This increases the days to keep builds but not the total number of builds kept, so it should not have a long-term impact on the storage capacity required for the buildfarm host. But we must remember to review implications on storage requirements when tuning these parameters in the future. 👍

PR #482, opened 2017-11-07
477 follow up: update plugin versions

mikaelarguedas:
I noticed a bungh of jobs being reconfigured overnight. By looking at the diff I identified some plugin versions that needed to be updated

### DIFF
--- a/ros_buildfarm/templates/release/trigger_upload_repo_job.xml.em
+++ b/ros_buildfarm/templates/release/trigger_upload_repo_job.xml.em
@@ -3,7 +3,7 @@
   <description/>
   <keepDependencies>false</keepDependencies>
   <properties>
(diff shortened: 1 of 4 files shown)


dirk-thomas:
@nuclearsandwich Please double check the configuration of *all job types* (same as in #477).

nuclearsandwich:
> @nuclearsandwich Please double check the configuration of all job types (same as in #477).

These can all get checked again, though the root cause that missed these changes was a failure in the mechanism I used to propagate changes in the config from the web UI back to the templates rather.

Per @mikaelarguedas's request I've put a ✅ on this PR so it can merge and I'll do another sweep before merging the xenial branch to master.

mikaelarguedas:
Merging this as it is an incremental improvement. We still need to double check the configuration of every job to see if there is any other plugin version update we missed

PR #492, opened 2017-12-20
Do not make job unstable if there are skipped tests

### LINKED ISSUE #491: Should skipped tests result in a stable or unstable job ?
This came up in https://github.com/ros/ros_comm/pull/1217#issuecomment-352960510

Skipped tests are making jobs unstable the same way as failed tests: https://github.com/ros-infrastructure/ros_buildfarm/blob/5634941787686f25e25c4b4450626ee3fad669cc/ros_buildfarm/templates/snippet/publisher_xunit.xml.em#L19

This is surprising to me as a user, as if I decide to disable a failing test I would expect my job to go from UNSTABLE to STABLE.

mikaelarguedas:
remove threshold for unstable results for skipped test
fixes #491


dirk-thomas:
[Build Status](http://build.ros.org/job/Lpr__ros_comm__ubuntu_xenial_amd64/674/)

dirk-thomas:
Thanks for the patch. The result looks good.

PR #493, opened 2018-01-02
Add in artful and bionic to the short os names.

clalancette:
While we are here, change the order of the listing so it is Ubuntu chronological, followed by Debian chronological.  This matches what we do in other repositories (like https://github.com/ros-infrastructure/buildfarm_deployment_config/blob/master/hiera/hieradata/buildfarm_role/repo.yaml#L100)

### DIFF
--- a/ros_buildfarm/common.py
+++ b/ros_buildfarm/common.py
@@ -275,9 +275,9 @@ def get_short_os_code_name(os_code_name):
         'xenial': 'X',
         'yakkety': 'Y',
         'zesty': 'Z',
-        'artful', : 'A',
-        'bionic', : 'B',
-        'wheezy', : 'W',
+        'artful': 'A',
+        'bionic': 'B',
+        'wheezy': 'W',
         'jessie': 'J',
         'stretch': 'S',
     }
--- a/ros_buildfarm/common.py
+++ b/ros_buildfarm/common.py
@@ -267,9 +267,7 @@ def get_short_os_name(os_name):
 
 def get_short_os_code_name(os_code_name):
     os_code_name_mappings = {
-        'jessie': 'J',
         'saucy': 'S',
-        'stretch': 'S',
         'trusty': 'T',
         'utopic': 'U',
         'vivid': 'V',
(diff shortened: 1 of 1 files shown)


ros_buildfarm/common.py
```diff
@@ -267,16 +267,19 @@ def get_short_os_name(os_name):
...
         'saucy': 'S',
-        'stretch': 'S',
         'trusty': 'T',
         'utopic': 'U',
         'vivid': 'V',
         'wily': 'W',
         'xenial': 'X',
         'yakkety': 'Y',
         'zesty': 'Z',
+        'artful', : 'A',
```
tfoote: Syntax error of extra comma on this and the next two lines.

ros_buildfarm/common.py
```diff
@@ -267,16 +267,19 @@ def get_short_os_name(os_name):
...
+        'bionic', : 'B',
+        'wheezy', : 'W',
+        'jessie': 'J',
```
tfoote: I think it makes sense to keep things in alphabetical order. It's not enforced here, but is what we do elsewhere.

nuclearsandwich: I don't have overridingly strong preferences but I do prefer chronological rather than alphabetical sorting for distro codenames.

dirk-thomas: The code currently uses an alphabetical order. When adding new distros that should stay that way.

Reordering should be done separately (which I agree with @tfoote is not desired imo).

ros_buildfarm/common.py
clalancette: Ouch, good call.  Fixed in a85cc40179feaec881aacf0d2dd4d32e488f1d0d

ros_buildfarm/common.py
clalancette: While I really don't care all that much, I will point out that other places in ros-infrastructure do use Ubuntu-release-order, Debian-release-order: https://github.com/ros-infrastructure/buildfarm_deployment_config/blob/master/hiera/hieradata/buildfarm_role/repo.yaml#L100, https://github.com/ros-infrastructure/rospkg/blob/master/stdeb.cfg#L7 (and the rest of the packages with stdeb.cfg).

That being said, I will change this back to the original order to get this moving.

PR #499, opened 2018-01-11
readme: point to the Buildfarm Discourse instead of the old SIG.

gavanderhoorn:
As per subject.

Seems more in-line with `buildfarm_deployment` and other related readmes.


dirk-thomas:
Thank you.

PR #500, opened 2018-01-16
Move all jobs that are at priority 40 down to 35.

clalancette:
This is to clear up a bit of space in the priority scheme
for the individual ROS distro builds.

### DIFF
--- a/ros_buildfarm/templates/release/release_trigger-broken-with-non-broken-upstream_job.xml.em
+++ b/ros_buildfarm/templates/release/release_trigger-broken-with-non-broken-upstream_job.xml.em
@@ -10,7 +10,7 @@
 ))@
 @(SNIPPET(
     'property_job-priority',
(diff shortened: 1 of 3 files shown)


clalancette:
I've manually updated all jobs with `trigger` in the name on the buildfarm to priority 35.  Merging this, thanks for the review.

dirk-thomas:
Can you please enumerate which jobs you have touched. I just looked at e.g. the `trigger-broken-with-non-broken-upstream` ones and they don't seem to have been updated.

clalancette:
> Can you please enumerate which jobs you have touched. I just looked at e.g. the trigger-broken-with-non-broken-upstream ones and they don't seem to have been updated.

Yeah, you were right.  I forgot to click "more" on the Jenkins search results, so some of the jobs didn't get updated.  I've now done that, and they should all be 35 now.

PR #501, opened 2018-01-17
resolve catkin instead of assuming current rosdistro.

tfoote:
This is necessary for addon rosdistros.

Re: https://answers.ros.org/question/269152/buildfarm-addon-fork-requires-ros-my_distribution-catkin/

I also considered deferring it to after the dependency checks and looking if 'catkin' was in `pkg_names + build_depends` but wanted to keep the change minimal.

### DIFF
--- a/scripts/devel/create_devel_task_generator.py
+++ b/scripts/devel/create_devel_task_generator.py
@@ -98,8 +98,7 @@ def main(argv=sys.argv[1:]):
         'python3',
     ]
     if 'catkin' not in pkg_names:
-        debian_pkg_names.append(
-            get_debian_package_name(args.rosdistro_name, 'catkin'))
+        debian_pkg_names += resolve_names(['catkin'], **context)
     print('Always install the following generic dependencies:')
     for debian_pkg_name in sorted(debian_pkg_names):
         print('  -', debian_pkg_name)


tfoote:
@tonybaltovski for testing

nuclearsandwich:
So I understand how this change would help the asker in the linked question:

Rather than assuming that the appropriate catkin package matches the current distro name, this looks it up using rosdep. I started following the call to `resolve_for_os` but stopped short of learning whether the logic checks for shadowed entry in the current rosdistro. Will this still find `ros-kinetic-catkin` on build.ros.org and likewise for other whole-rosdistro buildfarm setups?

tfoote:
Yes, the context is rosdistro aware so rosdep resolve will work. As with all rosdep lookups it will take the first resolution. So if you want to overlay a special version of catkin it would be necessary to make sure it comes first in the rosdep rule encountered. I don't know exactly how the context is resolved.

I ran a local test of our standard build to confirm using this branch. Here's the output from that section.

https://github.com/ros-infrastructure/ros_buildfarm/blob/master/doc/jobs/devel_jobs.rst


```
updated cache in /home/buildfarm/.ros/rosdep/sources.cache
Crawling for packages in workspace '/tmp/catkin_workspace/src'
Found the following packages:
  - cpp_common
  - roscpp_core
  - roscpp_serialization
  - roscpp_traits
  - rostime
Package maintainer emails: dthomas@osrfoundation.org
Resolved the dependencies to the following binary packages:
  - ros-indigo-catkin
Always install the following generic dependencies:
  - build-essential
  - python3
  - ros-indigo-catkin
Identified the following build dependencies (ignoring packages available from source):
  - boost
  - catkin
  - libconsole-bridge-dev
Resolved the dependencies to the following binary packages:
  - libboost-all-dev
  - libconsole-bridge-dev
  - ros-indigo-catkin
Identified the following run and test dependencies (ignoring packages available from source):
  - boost
  - libconsole-bridge-dev
Resolved the dependencies to the following binary packages:
  - libboost-all-dev
  - libconsole-bridge-dev
Using the following distribution repositories:
  http://repositories.ros.org/ubuntu/testing (/tmp/keys/0.key)
Generating Dockerfile '/tmp/docker_build_and_install/Dockerfile':
  # generated from devel/devel_task.Dockerfile.em
  
  FROM ubuntu:trusty
  ```

tfoote:
I don't know if it will be complete enough to resolve his use case. But we can go ahead and merge this and look for more feedback from that use case if other issues come up. With it merged it will be easier for him to test it for his use case.

PR #503, opened 2018-01-22
update list of EOL ubuntu distributions

mikaelarguedas:
addresses https://github.com/ros-infrastructure/ros_buildfarm_config/pull/103#issuecomment-359530548

needs https://github.com/osrf/multiarch-docker-image-generation/pull/16 to be deployed to fix zesty builds

### DIFF
--- a/ros_buildfarm/templates/snippet/old_release_set.Dockerfile.em
+++ b/ros_buildfarm/templates/snippet/old_release_set.Dockerfile.em
@@ -1,3 +1,3 @@
-@[if os_name == 'ubuntu' and os_code_name in ['saucy', 'utopic']]@
+@[if os_name == 'ubuntu' and os_code_name in ['saucy', 'utopic', 'vivid', 'wily', 'yakkety', 'zesty']]@
 RUN find /etc/apt/ -name *.list -exec sed -i -e 's/archive.ubuntu.com\|security.ubuntu.com/old-releases.ubuntu.com/g' {} \;
 @[end if]@


dirk-thomas:
> needs osrf/multiarch-docker-image-generation#16 to be deployed to fix zesty builds

Can you clarify why this is being blocked on the other PR? Afaik the `ros_buildfarm` doesn't use these Docker images anywhere except for ARM and we don't have any ARM build for non-LTS distros.

mikaelarguedas:
You're right, I should rephrase. If the zesty image used doesnt have the sources updated to point to the new location the builds will fail. This connects to the PR updating the source in our custom docker images used for building.
As all the jobs try to update / install packages, we either need to modify the sources everywhere by including the snippet modified in this PR or get the ubuntu base docker images to have the sources already updated as they are unusable as is.

mikaelarguedas:
As discussed offline, this will not allow all job types to build on EOL distros but makes the snippet more accurate. As we don't plan on spending time on making sure all job types work on EOL distros we merge this as is. We can reconsider the follow-up work if there is high demand from users.

dirk-thomas:
For context: job types which don't use this snippet won't work. In the past we haven't build doc jobs, devel jobs, or PR jobs on non-LTS distros so those didn't work before and won't work for newer EOL distros with this patch.

mikaelarguedas:
Additional context: source jobs will need the change as they are built on non-LTS distros

dirk-thomas:
> source jobs will need the change as they are built on non-LTS distros

I think it would be valuable to add support for those in a follow up PR.

mikaelarguedas:
> I think it would be valuable to add support for those in a follow up PR.

binary and source jobs have been fixed in #504

PR #505, opened 2018-01-23
use -d option to skip checking for build deps in source jobs on newer Ubuntu

dirk-thomas:
Same as #384 just for newer Ubuntu distros.

### DIFF
--- a/ros_buildfarm/sourcedeb_job.py
+++ b/ros_buildfarm/sourcedeb_job.py
@@ -108,7 +108,14 @@ def build_sourcedeb(sources_dir, os_name=None, os_code_name=None):
         '--git-ignore-branch',
         # dpkg-buildpackage args
         '-S']
-    if os_name == 'debian' and os_code_name == 'stretch':
+    ubuntu_before_artful = (
+        'precise', 'quantal', 'raring', 'saucy',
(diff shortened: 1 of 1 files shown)


dirk-thomas:
Worked: http://build.ros.org/view/Msrc_uA/job/Msrc_uA__catkin__ubuntu_artful__source/5/

ros_buildfarm/sourcedeb_job.py
```diff
@@ -108,7 +108,14 @@ def build_sourcedeb(sources_dir, os_name=None, os_code_name=None):
         '--git-ignore-branch',
         # dpkg-buildpackage args
         '-S']
-    if os_name == 'debian' and os_code_name == 'stretch':
+    ubuntu_before_artful = (
+        'precise', 'quantal', 'raring', 'saucy',
+        'trusty', 'utopic', 'vivid', 'wily',
+        'xenial', 'yakkety', 'zesty')
+    if (
+        os_name == 'debian' and os_code_name == 'stretch' or
```
tfoote: Before modification for debian would make sense so this doesn't bit us in the future on debian.

dirk-thomas: Done in 4363beef3a3c621470fa582fa95ae15f1090824a.

PR #506, opened 2018-01-23
install gnupg on newer Ubuntu

dirk-thomas:
Similar as #372.

### DIFF
--- a/ros_buildfarm/templates/snippet/add_distribution_repositories.Dockerfile.em
+++ b/ros_buildfarm/templates/snippet/add_distribution_repositories.Dockerfile.em
@@ -1,5 +1,12 @@
 RUN mkdir /tmp/keys
-@[if os_name == 'debian' and os_code_name == 'stretch']@
+@{
+debian_before_stretch = ('squeeze', 'wheezy', 'jessie')
+ubuntu_before_bionic = (
+    'precise', 'quantal', 'raring', 'saucy',
+    'trusty', 'utopic', 'vivid', 'wily',
+    'xenial', 'yakkety', 'zesty', 'artful')
+}@
+@[if os_name == 'debian' and os_code_name not in debian_before_stretch or os_name == 'ubuntu' and os_code_name not in ubuntu_before_bionic]@
 @# In Debian Stretch apt doesn't depend on gnupg anymore
 @# https://anonscm.debian.org/cgit/apt/apt.git/commit/?id=87d468fe355c87325c943c40043a0bb236b2407f
 RUN for i in 1 2 3; do apt-get update && apt-get install -q -y gnupg && apt-get clean && break || if [[ $i < 3 ]]; then sleep 5; else false; fi; done


dirk-thomas:
The artful build worked without this patch: http://build.ros.org/view/Msrc_uA/job/Mbin_uA64__catkin__ubuntu_artful_amd64__binary/1/

dirk-thomas:
The original problem works: http://build.ros.org/view/Msrc_uB/job/Mbin_ubhf_uBhf__catkin__ubuntu_bionic_armhf__binary/2/

The remaining failure is due to the fact that `locale-gen` failed in http://build.ros.org/view/Msrc_uB/job/Mbin_ubhf_uBhf__catkin__ubuntu_bionic_armhf__binary/1/ and Docker cached the "broken" image layer.

mikaelarguedas:
It seems more correct to do it for both artful and bionic if we don't know where the dependency comes from (as this may change in the future), but I'd understand if you want to stick to bionic only to avoid the overhead

dirk-thomas:
Even after deleting the docker image and regenerating it the result is the same:

```
Step 6/26 : RUN locale-gen en_US.UTF-8
 ---> Running in 5269f4ffd8ee
Generating locales (this might take a while)...
  en_US.UTF-8...localedef: ../sysdeps/unix/sysv/linux/spawni.c:368: __spawnix: Assertion `ec >= 0' failed.
qemu: uncaught target signal 6 (Aborted) - core dumped
/usr/sbin/locale-gen: line 177:    25 Aborted                 (core dumped) localedef $no_archive -i $input -c -f $charset $locale_alias $locale
 done
Generation complete.
 ---> 1443447abb83
```

tfoote:
I think that's a separate issue from the gpg one this is addressing. It's only on armhf and armv8 with qemu.

dirk-thomas:
> I think that's a separate issue

Indeed: https://bugs.launchpad.net/qemu/+bug/1673976

mikaelarguedas:
Agreed with @tfoote, I don't think it's related to this PR. Other surprise (to me), `locales` is already installed in the our arm images while its not in the base image

dirk-thomas:
Follow up: osrf/multiarch-docker-image-generation#17

gavanderhoorn:
Are these ROS Answers Q&As related (as in: deal with similar issues, they're not reporting issues with the farm)?

 - [GPG error on Debian](https://answers.ros.org/question/264654)
 - [ROS-Lunar Install on Debian Stretch: apt-key deprecation](https://answers.ros.org/question/275824)

@IanTheEngineer mentions in the second one that use of `apt-key` (and `gnupg` and `dirmngr` are deprecated - or at least have been in the Debian community.

tfoote:
It's loosely related in that debian best practices are slowly changing. We don't use the exact recommended public installation instructions because we actually want the full key in our config instead of using the url redirection and adding another resource to our build dependencies.

Issue #507, opened 2018-01-29
release new version targeting hosts based on Xenial


jonazpiazu:
According to the [job generation documentation](https://github.com/ros-infrastructure/ros_buildfarm/blob/master/doc/environment.rst#virtualenv) , the ros_buildfarm is installed using `pip`. 

However the package is not updated. For example, for the `import_upstream` job, the path to the configuration file wrongly points to `/home/jenkins-slave/reprepro_config/ros_bootstrap.yaml`

https://github.com/ros-infrastructure/ros_buildfarm/blob/master/ros_buildfarm/templates/release/import_upstream_job.xml.em#L29

dirk-thomas:
When installing this package via `pip` using the package name the latest release is being installed. Currently that is version 1.4.1 from August 30th 2017. That version was tagged when Ubuntu Trusty was the targeted host platform. The line you referenced has been changed in the meantime to target Xenial as the host system.

Since there are several regressions (ros-infrastructure/buildfarm_deployment#165, ros-infrastructure/buildfarm_deployment#176, the `import_upstream` job is significantly slower) still pending related to the migration to Xenial there hasn't been a new version of `ros_buildfarm` yet. If you still want to use the latest deployment using Xenial you need to install the latest code from the master branch of this repo rather than the latest release (e.g. but invoking `pip` with the URL of this repo).

jonazpiazu:
I see. Then I suggest to update the install instructions, as I would assume that they currently apply to the Xenial version.

dirk-thomas:
> I would assume that they currently apply to the Xenial version.

The current instructions start with:

> The following commands can be used on Ubuntu Trusty::

jonazpiazu:
My understanding from [this](https://discourse.ros.org/t/ros-buildfarm-october-2017-guide-to-new-changes/2840) and [this](https://discourse.ros.org/t/status-of-xenial-migration/3330/9) posts was that the *official* version for the buildfarm is Xenial. But maybe I run too much.

dirk-thomas:
That is correct. Yet there is no new release of `ros_buildfarm` at this point for the reasons mentioned above. Yes, that is not optimal but that is how it is atm.

cclauss:
Would it be worthwhile to take another look at this one?
* The two issues sited were closed in March 2018.
* Trusty will EOL in April 2019
* Python 3.7 is current but we are advocating 3.4.

dirk-thomas:
[...]

PR #508, opened 2018-01-29
release new version targeting hosts based on Xenial

jonazpiazu:
More info in this issue https://github.com/ros-infrastructure/ros_buildfarm/issues/507

### DIFF
--- a/doc/environment.rst
+++ b/doc/environment.rst
@@ -38,7 +38,7 @@ The following commands can be used on Ubuntu Trusty::
   pip3 install empy
   pip3 install jenkinsapi
   pip3 install rosdistro
-  pip3 install ros_buildfarm
+  pip3 install git+https://github.com/ros-infrastructure/ros_buildfarm.git
 
 
 Provide credentials for Jenkins master


dirk-thomas:
This change also implies that the buildfarm you are creating would track `master` which I would rather not recommend. While we do our best to keep things stable some changes on the master branch might require changes to the deployment at the same time. So it is recommended to use the latest release unless you want to live on the bleeding edge...

jonazpiazu:
The problem in my case is that there is no release for buildfarm_deployment and buildfarm_deployment_config, so I am using the master for the deployment.

But without the change for the ros_buildfarm version, I am not able to setup a working scenario, as the last tagged version of ros_buildfarm is not "compatible" with the master version of buildfarm_deployment.

So that is why I consider that this PR is relevant for anyone trying to deploy the buildfarm now.

dirk-thomas:
Yes, this patch will help as a temporary workaround. But for the reason mentioned above I don't think it is sane "default" recommendation and therefore I won't merge it but keep it open for visibility until there is a new release of this package targeting Xenial.

dirk-thomas:
Closing in favor of the referenced issue #507.

PR #512, opened 2018-02-08
use 'old' Debian naming for ardent release

dirk-thomas:
Follow up of #460.

Since the previous packages of Ardent have been built with the "old" naming scheme (containing dashed) the patch releases should continue to do so.

Only for ROS 2 B-turtle we will switch to the "new" naming scheme (containing dots).


dirk-thomas:
@nuclearsandwich FYI, I updated the `ros2` fork accordingly: https://github.com/ros2/ros_buildfarm/commit/93d19cb8003ecb34317c63c0ad0284973ddb04c3

nuclearsandwich:
> @nuclearsandwich FYI, I updated the ros2 fork accordingly: ros2/ros_buildfarm@93d19cb

🙇

PR #515, opened 2018-02-11
git clone with --recurse-submodules

flixr:
Recurse into git submodules when cloning.

Tested and working for the prerelease job.

### DIFF
--- a/ros_buildfarm/templates/snippet/scm_git.xml.em
+++ b/ros_buildfarm/templates/snippet/scm_git.xml.em
@@ -42,5 +42,12 @@
         </options>
       </hudson.plugins.git.extensions.impl.PreBuildMerge>
 @[end if]@
+      <hudson.plugins.git.extensions.impl.SubmoduleOption>
+        <disableSubmodules>false</disableSubmodules>
+        <recursiveSubmodules>true</recursiveSubmodules>
+        <trackingSubmodules>false</trackingSubmodules>
+        <reference></reference>
+        <parentCredentials>false</parentCredentials>
+      </hudson.plugins.git.extensions.impl.SubmoduleOption>
     </extensions>
   </scm>
--- a/ros_buildfarm/templates/devel/devel_script_clone.sh.em
+++ b/ros_buildfarm/templates/devel/devel_script_clone.sh.em
@@ -8,7 +8,7 @@ mkdir -p @workspace_path/src
 @[for repo_spec, path in scms]@
 if [ ! -d "@path" ]; then
 @[if repo_spec.type == 'git']@
-    (set -x; git clone -b @repo_spec.version @repo_spec.url @path)
+    (set -x; git clone --recurse-submodules -b @repo_spec.version @repo_spec.url @path)
     (set -x; git -C @path --no-pager log -n 1)
 @[elif repo_spec.type == 'hg']@
     (set -x; hg clone -b @repo_spec.version @repo_spec.url @path)


dirk-thomas:
Thank you for creating this patch. This will use submodules only for locally run devel / prerelease jobs. I think it would be preferred to keep the behavior the same across job types as well as where the job is run. Therefore this should be applied to all of them to keep the behavior consistent. It would be great if you could update the patch accordingly.

gavanderhoorn:
Just curious @dirk-thomas: was there a specific reason the scripts were not cloning recursively?

I remember some questions on ROS Answers about packages with submodules in their repositories not building properly on the farm. I seem to remember there was some specific reason why `--recursive` was not specified.

dirk-thomas:
I don't recall any specific reason. Do you by any chance have a reference to that discussion?

flixr:
@dirk-thomas am I missing something? To me it looks like the `devel_script_clone.sh.em` is used in all cases for cloning.
At least I didn't find any other occurrences of `git clone`.

dirk-thomas:
The Jenkins jobs don't invoke `git clone` directly. The `git` plugin is performing the cloning of the repository (which uses the Java API of `git` behind the scenes.

flixr:
So for the cloning of the release repos this should not be needed... only for the upstream repos...
But I guess for the doc job.

ros_buildfarm/templates/doc/doc_independent_job.xml.em
```diff
@@ -64,7 +64,7 @@ repo_name = os.path.splitext(os.path.basename(repo_url))[0]
     'builder_shell',
     script='\n'.join([
         'echo "# BEGIN SECTION: Clone %s"' % repo_name,
-        'python3 -u $WORKSPACE/ros_buildfarm/scripts/wrapper/git.py clone --depth 1 %s $WORKSPACE/repositories/%s' % (repo_url, repo_name),
+        'python3 -u $WORKSPACE/ros_buildfarm/scripts/wrapper/git.py clone --recurse-submodules --depth 1 %s $WORKSPACE/repositories/%s' % (repo_url, repo_name),
```
dirk-thomas: This is only used for the folllowing repos and neither of them uses submodules: https://github.com/ros-infrastructure/ros_buildfarm_config/blob/01e30d05362b4763567c459e8e4612f7dc49619b/doc-independent-build.yaml#L5

So this change isn't needed.

dirk-thomas: Local doc jobs are already using the devel template for cloning the repositories, see https://github.com/ros-infrastructure/ros_buildfarm/blob/93d19cb8003ecb34317c63c0ad0284973ddb04c3/ros_buildfarm/templates/doc/doc_script.sh.em#L44

dirk-thomas:
What is not covered by the PR yet are the jobs which are run on Jenkins (not locally). The configuration for the `git` plugin within these jobs is coming from this template: https://github.com/ros-infrastructure/ros_buildfarm/blob/93d19cb8003ecb34317c63c0ad0284973ddb04c3/ros_buildfarm/templates/snippet/scm_git.xml.em

flixr:
Ok, I can remove the commit for that doc job again...

I'm not that familiar with the buildfarm yet, in which cases are the upstream repos (which might contain submodules) cloned?
Is it only for the doc jobs? Because for the release jobs it should not be needed, as the bloom/gbp release repos never contain submodules...

dirk-thomas:
All jobs which operate on the source repository would benefit from supporting submodules: `devel`, `pull request`, `prerelease`, `doc`. The first three are very similar and share most of the code / logic.

gavanderhoorn:
https://answers.ros.org/question/167996/possible-to-use-git-submodules-for-bloom/

dirk-thomas:
I would think it was just never implemented since the need wasn't there. As mentioned in the answer:

> If your upstream repository uses submodules, you'll either have to extend bloom to support submodules

Maybe @wjwwood can pitch in here but I would think it would be possible to make a PR for `bloom` to support this use case if anybody wants to.

wjwwood:
It was fixed in the past via `vcstools`, and @flixr already made a pr to let bloom support it (bypass error message): https://github.com/ros-infrastructure/bloom/pull/461

flixr:
> Maybe @wjwwood can pitch in here but I would think it would be possible to make a PR for bloom to support this use case if anybody wants to.

If https://github.com/vcstools/vcstools/pull/130 is accepted, it works with bloom (see https://github.com/ros-infrastructure/bloom/pull/461)

flixr:
> All jobs which operate on the source repository would benefit from supporting submodules: devel, pull request, prerelease, doc. The first three are very similar and share most of the code / logic.

Makes sense. I'm however unsure how to proceed here as I'm not really familiar with Jenkins.
It seems the [Jenkins git plugin](https://plugins.jenkins.io/git) has an option to recursively update submodules, but I don't know how exactly it should be named in the xml template or where it should be added.

flixr:
I added the `recursiveSubmodules` option to the jenkins template.
However I'm not 100% sure this is correct and have no way to test this.

flixr:
@dirk-thomas IMHO this is done.
The tests also show that Jenkins correctly clones with `--recurse-submodules` as enabled with the `recursiveSubmodules` Jenkins option.

ros_buildfarm/templates/snippet/scm_git.xml.em
```diff
@@ -18,6 +18,7 @@
     </branches>
     <doGenerateSubmoduleConfigurations>false</doGenerateSubmoduleConfigurations>
     <submoduleCfg class="list"/>
+    <recursiveSubmodules>true</recursiveSubmodules>
```
dirk-thomas: As far as I can see this is not a valid configuration option. I think it needs to be the following instead (after the `hudson.plugins.git.extensions.impl.CloneOption` / `hudson.plugins.git.extensions.impl.PreBuildMerge` tag):

```
<hudson.plugins.git.extensions.impl.SubmoduleOption>
  <disableSubmodules>false</disableSubmodules>
  <recursiveSubmodules>true</recursiveSubmodules>
  <trackingSubmodules>false</trackingSubmodules>
  <reference></reference>
  <parentCredentials>false</parentCredentials>
  </hudson.plugins.git.extensions.impl.SubmoduleOption>
</extensions>
```

flixr: Ok, from the test output it just looked like it was already cloning with `--recurse-submodules`, although I have to admit that I don't really know where to look at exactly.
Also I couldn't really find a reference for specifying these options in such an XML config file, only how to set it in the web interface.

flixr: Ah, thx... you already updated it.

dirk-thomas: > Ok, from the test output it just looked like it was already cloning with `--recurse-submodules`.

The Travis CI won't tell anything about Jenkins. Those don't run a Jenkins server but only invoke the "script" versions of all jobs. They only show that your change in the `devel_script_clone.sh.em` file is being used and works.

flixr: Ah, that makes sense.

dirk-thomas:
The patch looks fine now. Thank you for the contribution.

I will wait with merging this until the similar changes are being accepted into the other tools (ros-infrastructure/bloom#461, vcstools/vcstools#130).

flixr:
@dirk-thomas could you merge this already?
This doesn't directly depend on the other PRs (those are "only" needed to actually do a release with bloom) and the devel jobs are failing for the repos with submodules...

dirk-thomas:
I would rather keep the supported features "the same" across the tools. Rather than diverging here we should push the maintainers on the other repos to accept the patches.

flixr:
@dirk-thomas sorry that I'm asking the same thing again and basically repeating myself...
But it would be really great if this could be merged.
With https://github.com/ros-infrastructure/bloom/pull/461 merged using bloom for repos with "regular" submodules definitely works, https://github.com/vcstools/vcstools/pull/130 is only needed for repos that have submodules with relative URLs.

wjwwood:
How is vcstools used here? is it installed from apt or pip?

wjwwood:
Err, I guess it doesn't matter, either way you need it to be released not just merged right?

Even if you're using this repository from source it won't use the others from source.

ros_buildfarm/templates/snippet/scm_git.xml.em
dirk-thomas: I think this added line still needs to be reverted. The newly added block below is performing the intended feature.

dirk-thomas:
With both referenced PRs being merged this can be merged too.

Thank you for adding support for submodules.

flixr:
:smile: Thanks guys for the constructive feedback and I really appreciate all the effort you put in to keeping things working for all cases and keeping it maintainable.

PR #517, opened 2018-02-20
document and use option `canonical_base_url`

### LINKED ISSUE #511: doc jobs: 'api_documentation' URL hard-coded?
I'm not entirely sure, but it appears that the URL template that is used to populate the `api_documentation` key in `manifest.yaml` is hard-coded to always point to `docs.ros.org`:

https://github.com/ros-infrastructure/ros_buildfarm/blob/0d35d14a1e332fc59d2fa4365f37fd586a9cab50/scripts/doc/create_doc_task_generator.py#L399-L400

Is this as it should be, or should this actually be based on the base URL of the farm being deployed? Could the `canonical_base_url` key in the `doc-build.yaml` file be used for that?

dirk-thomas:
Fixes #511.


gavanderhoorn:
Haven't been able to test it, but this seems to address #511. Thanks @dirk-thomas, I hadn't had an opportunity to do something about #511.

PR #520, opened 2018-03-05
Add the mail publisher to the trigger_upload_repo_job.

clalancette:
This makes it match the actual configuration on the build
farm.

I realize that there are 2 hard-coded email addresses in here.  We should probably change that to ros-buildfarm@googlegroups.com, but this configuration matches the one that is currently deployed.  I'm happy to change it, but if we do, we'll need to re-deploy to the build farm.

### DIFF
--- a/ros_buildfarm/templates/release/trigger_upload_repo_job.xml.em
+++ b/ros_buildfarm/templates/release/trigger_upload_repo_job.xml.em
@@ -38,11 +38,12 @@ $HOME/upload_triggers/upload_repo.bash @repo
     </hudson.tasks.Shell>
   </builders>
   <publishers>
-    <hudson.tasks.Mailer plugin="mailer@@1.20">
-      <recipients>@notification_emails</recipients>
-      <dontNotifyEveryUnstableBuild>false</dontNotifyEveryUnstableBuild>
-      <sendToIndividuals>false</sendToIndividuals>
-    </hudson.tasks.Mailer>
+@(SNIPPET(
+    'publisher_mailer',
+    recipients=recipients,
+    dynamic_recipients=[],
+    send_to_individuals=False
+))@
   </publishers>
   <buildWrappers/>
 </project>
--- a/ros_buildfarm/templates/release/trigger_upload_repo_job.xml.em
+++ b/ros_buildfarm/templates/release/trigger_upload_repo_job.xml.em
@@ -37,6 +37,12 @@ $HOME/upload_triggers/upload_repo.bash @repo
 </command>
     </hudson.tasks.Shell>
   </builders>
-  <publishers/>
+  <publishers>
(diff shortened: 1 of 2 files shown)


nuclearsandwich:
> We should probably change that to ros-buildfarm@googlegroups.com

Assuming that we want these messages sent to that group, we should use the existing config item for notification email recipients: https://github.com/ros-infrastructure/ros_buildfarm_config/blob/production/index.yaml#L79-L81

If we want them sent to infra peeps like @tfoote and myself without sending to the overall buildfarm we should create a separate config key or something.

dirk-thomas:
@clalancette Can you please follow up on the comments so that this can be merged.

nuclearsandwich:
I updated this to use the top-level notification_emails values from the ros_buildfarm_config index.

Dry run diff:

```diff
<<<
--- remote config
+++ new config
@@ -42 +42 @@
-      <recipients>steven+build.ros.org@openrobotics.org tfoote+buildfarm@osrfoundation.org</recipients>
+      <recipients>ros-buildfarm@googlegroups.com tfoote+buildfarm@osrfoundation.org</recipients>
>>>
<<<
--- remote config
+++ new config
@@ -42 +42 @@
-      <recipients>steven+build.ros.org@openrobotics.org tfoote+buildfarm@osrfoundation.org</recipients>
+      <recipients>ros-buildfarm@googlegroups.com tfoote+buildfarm@osrfoundation.org</recipients>
>>>
```

ros_buildfarm/templates/release/trigger_upload_repo_job.xml.em
```diff
@@ -37,6 +37,12 @@ $HOME/upload_triggers/upload_repo.bash @repo
...
     </hudson.tasks.Shell>
   </builders>
-  <publishers/>
+  <publishers>
+    <hudson.tasks.Mailer plugin="mailer@@1.20">
+      <recipients>@notification_emails</recipients>
+      <dontNotifyEveryUnstableBuild>false</dontNotifyEveryUnstableBuild>
+      <sendToIndividuals>false</sendToIndividuals>
+    </hudson.tasks.Mailer>
+  </publishers>
```
dirk-thomas: Please use the `publisher_mailer.xml.em` snippet for this.

nuclearsandwich:
Same diff is produced after switching to the mailer snippet.

nuclearsandwich:
> Same diff is produced after switching to the mailer snippet.

Sorry, I meant to say that I checked again after switching to the snippet and _confirmed_ the diff is the same.

dirk-thomas:
> I meant to say that I checked again after switching to the snippet and confirmed the diff is the same.

I wasn't suggesting otherwise. Thanks for double checking! Better than having to re-release after a day :wink:

PR #521, opened 2018-03-06
Generate YAML of build information

DLu:
In lieu of [the csv that used to be generated](http://www.ros.org/debbuild/hydro.csv), this PR now generates a neat YAML file with the build information. 

I have not tested with the main build farm, but a private build farm. With the main build farm, I cannot access http://54.183.65.232. 

Resulting yaml entry:
```
navigation:
  build_status:
    ubuntu:
      xenial:
        amd64: 
            build: 1.15.1-0xenial-20180215-152207-0800
            main: 1.15.1-0xenial-20171213-034534-0800
            test: 1.15.1-0xenial-20180215-152207-0800
        source:
            build: 1.15.1-0xenial
            main: 1.15.1-0xenial
            test: 1.15.1-0xenial}
  maintainers:
  - {email: davidvlu@gmail.com, name: David V. Lu!!}
  - {email: mferguson@fetchrobotics.com, name: Michael Ferguson}
  status: developed
  url: https://github.com/ros-planning/navigation.git
  version: 1.15.1-0
```

### DIFF
--- a/ros_buildfarm/status_page.py
+++ b/ros_buildfarm/status_page.py
@@ -1044,20 +1044,7 @@ def write_yaml(yaml_filename, ordered_pkgs, repos_data):
         pkg_d = {'version': pkg.version, 'url': pkg.repository_url, 'status': pkg.status}
         if pkg.status_description:
             pkg_d['status_description'] = pkg.status_description
-        pkg_d['maintainers'] = []
-        for maintainer in pkg.maintainers:
-            maintainer_dict = {'email': maintainer.email}
-            # In order to write the simplest data to the yaml, we try to encode the name to ascii.
-            ascii_name = maintainer.name.decode('utf-8').encode('ascii', 'ignore')
-            if maintainer.name == ascii_name:
-                # If it is the same, we write the ascii name to the dict
-                # Otherwise it will appear in the file as
-                # name: !!python/unicode 'David V. Lu!!'
-                maintainer_dict['name'] = ascii_name
-            else:
-                # Otherwise, we write in all its unicode glory
-                maintainer_dict['name'] = maintainer.name
-            pkg_d['maintainers'].append(maintainer_dict)
+        pkg_d['maintainers'] = [{'email': m.email, 'name': m.name} for m in pkg.maintainers]
 
         pkg_d['build_status'] = {}
 
@@ -1076,4 +1063,4 @@ def write_yaml(yaml_filename, ordered_pkgs, repos_data):
         summary[pkg.name] = pkg_d
 
     with open(yaml_filename, 'w') as f:
-        yaml.dump(summary, f, allow_unicode=True)
+        yaml.safe_dump(summary, f, allow_unicode=True)
--- a/ros_buildfarm/status_page.py
+++ b/ros_buildfarm/status_page.py
@@ -23,6 +23,7 @@ import re
 import shutil
 import sys
 import time
+import yaml
 
 from .common import get_debian_package_name
 from .common import get_release_view_name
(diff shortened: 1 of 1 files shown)


ros_buildfarm/status_page.py
```diff
@@ -1030,3 +1034,39 @@ def _compare_package_version(distros, pkg_name):
...
+        pkg_d = {'version': pkg.version, 'url': pkg.repository_url, 'status': pkg.status}
+        if pkg.status_description:
+            pkg_d['status_description'] = pkg.status_description
+        pkg_d['maintainers'] = []
+        for maintainer in pkg.maintainers:
+            maintainer_dict = {'email': maintainer.email}
+            try:
+                maintainer_dict['name'] = str(maintainer.name)
+            except:
+                maintainer_dict['name'] = maintainer.name
```
dirk-thomas: Can you describe why the `try` / `except` is needed (potentially adding a comment to the code to clarify)?

ros_buildfarm/status_page.py
```diff
@@ -1030,3 +1034,39 @@ def _compare_package_version(distros, pkg_name):
...
+                # Dynamically Create the Nested Dictionary
+                d = pkg_d['build_status']
+                for field in [key.os_name, key.os_code_name, key.arch]:
+                    if field not in d:
+                        d[field] = {}
+                    d = d[field]
+                d[name] = str(build_data[pkg.debian_name])
+        summary[pkg.name] = pkg_d
+
+    yaml.dump(summary, open(summary_filename, 'w'))
```
dirk-thomas: Please use a context manager to take care of closing the file

```
with open(...) as h:
    yaml.dump(summary, h)
```

ros_buildfarm/status_page.py
```diff
@@ -1030,3 +1034,39 @@ def _compare_package_version(distros, pkg_name):
         data[i] = '<div>%s</div>' % value
 
     return data
+
+
+REPOS_DATA_NAMES = ['build', 'test', 'main']
+def write_summary(summary_filename, ordered_pkgs, repos_data):
```
dirk-thomas: Maybe rename this to `write_yaml` to clarify what the function is doing. Otherwise the reader might expect it printing a line with some statistics.

ros_buildfarm/status_page.py
```diff
@@ -1047,9 +1047,15 @@ def write_summary(summary_filename, ordered_pkgs, repos_data):
...
-            except:
+            # In order to write the simplest data to the yaml, we try to encode the name to ascii.
+            ascii_name = maintainer.name.decode('utf-8').encode('ascii', 'ignore')
+            if maintainer.name == ascii_name:
+                # If it is the same, we write the ascii name to the dict
+                # Otherwise it will appear in the file as
+                # name: !!python/unicode 'David V. Lu!!'
+                maintainer_dict['name'] = ascii_name
+            else:
+                # Otherwise, we write in all its unicode glory
```
dirk-thomas: Instead of this approach I would suggest to "suppress/avoid" the `!!python/unicode` when dumping the data (e.g. using a custom dumper class).

But why do you actually care if `!!python/unicode` is in the generated yaml? Are there any cases where you can't read that file?

DLu: * Readability of the file
 * Portability to other languages
 * File size

dirk-thomas:
Thank you for adding this feature.

DLu:
Can I request that someone ping me when this feature/a new release gets deployed to the build farm?

dirk-thomas:
Ping :wink: The change is immediately available on the ROS buildfarm after merging (it uses the master branch rather than a release). You will find the current `yaml` files in http://repositories.ros.org/status_page/

tfoote:
Can we push these files into a subdirectory. They are making the index a lot harder to read now. Or maybe we should just create an index page instead of relying on the directory listing.

PR #522, opened 2018-03-09
Yaml folder

DLu:
As per https://github.com/ros-infrastructure/ros_buildfarm/pull/521#issuecomment-371645922


dirk-thomas:
Thanks.

PR #523, opened 2018-03-14
Use the distribution cache for the package names instead of the distf…

clalancette:
…ile.

This will ensure that we always get an authoritative copy, as opposed to a possibly cached, outdated one.

I've tested this a bit locally, and this seems to give me the same results as using the `dist_file`.  If we switch to this, it will increase the load on repositories.ros.org, but the files are small enough that I don't think that should be a huge issue.  There may be other gotchas that I'm not aware of with this change; I want to see what everyone thinks of doing this.

### DIFF
--- a/ros_buildfarm/release_job.py
+++ b/ros_buildfarm/release_job.py
@@ -70,7 +70,14 @@ def configure_release_jobs(
         print('  - %s %s: %s' % (os_name, os_code_name, ', '.join(
             build_file.targets[os_name][os_code_name])))
 
-    dist_file = get_distribution_file(index, rosdistro_name, build_file)
+    # We've seen situations in the past where fetching the distfile from
+    # GitHub gives us an out-of-date copy (probably because of some CDN in
+    # the middle).  The distribution cache, on the other hand, comes directly
+    # from the ROS infrastructure, so should always be up-to-date.  Get the
+    # dist_file from the dist_cache to ensure it is always the latest.
+    dist_cache = get_distribution_cache(index, rosdistro_name)
+    dist_file = dist_cache.distribution_file
+
     if not dist_file:
         print('No distribution file matches the build file')
         return
@@ -84,8 +91,6 @@ def configure_release_jobs(
               ('ignored' if build_file.skip_ignored_packages else 'disabled'))
         for pkg_name in sorted(explicitly_ignored_pkg_names):
             print('  -', pkg_name)
-
-    dist_cache = get_distribution_cache(index, rosdistro_name)
 
     if explicitly_ignored_pkg_names:
         # get direct dependencies from distro cache for each package
@@ -174,16 +179,6 @@ def configure_release_jobs(
         pkg_xml = dist_cache.release_package_xmls[pkg_name]
         pkg = parse_package_string(pkg_xml)
         pkgs[pkg_name] = pkg
-
-    # We have seen situations in the past where the distribution
-    # file (dist_file) version we get is a cached, out-of-date version from a
-    # CDN, but where the distribution cache (dist_cache) is up-to-date.  Check
-    # for and warn about this situation so that nobody has to spend time
-    # debugging when this is the case.
-    for pkg_name in dist_cache.release_package_xmls.keys():
-        if pkg_name not in pkgs:
-            print("Skipping package '%s': in cache but not in distribution file (can happen if a CDN returns an outdated distfile)" % (pkg_name), file=sys.stderr)
-
     ordered_pkg_tuples = topological_order_packages(pkgs)
 
     other_build_files = [v for k, v in build_files.items() if k != release_build_name]
--- a/ros_buildfarm/release_job.py
+++ b/ros_buildfarm/release_job.py
@@ -75,16 +75,7 @@ def configure_release_jobs(
         print('No distribution file matches the build file')
         return
 
-    dist_cache = get_distribution_cache(index, rosdistro_name)
-
-    # We can get the list of packages either from dist_file.release_packages,
-    # or from dist_cache.release_package_xmls.  The problem with the former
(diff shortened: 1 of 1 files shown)


dirk-thomas:
This sounds good to me. How about the other job types?

nuclearsandwich:
I took a look at what it would take to use the rosdistro cache for harvesting package names authoritatively for other jobs and it has raised some questions:

Left in the release_job.py source is a guard against configuring packages that are not yet in the rosdistro cache.

https://github.com/ros-infrastructure/ros_buildfarm/blob/3b98057cee81bbb96cd95cd05b02ccfda9dd10c7/ros_buildfarm/release_job.py#L170-L171

But I do not see any guard for the reverse condition, where a package is in the rosdistro_cache but not part of the distribution file. This wasn't a problem when our set of package names was coming from the distribution file, but do we need to reverse that guard now since it's possible for that condition to arise?

And if we need info from the distribution file to complete the task at hand, which appears to be the case:

https://github.com/ros-infrastructure/ros_buildfarm/blob/3b98057cee81bbb96cd95cd05b02ccfda9dd10c7/ros_buildfarm/release_job.py#L190-L192

Then does this change still offer an improvement?

nuclearsandwich:
I haven't done much digging to see if other jobs pull package names but nothing else from the distribution file, which would allow them to benefit from this change without addressing my question above, but so I don't lose the work here are the other jobs that pull a full list of package names using the same mechanic:

```
ros_buildfarm/trigger_job.py:    pkg_names = dist_file.release_packages.keys()
ros_buildfarm/release_job.py:    pkg_names = dist_file.release_packages.keys()
scripts/release/check_sync_criteria.py:    all_pkg_names = dist_file.release_packages.keys()
ros_buildfarm/status_page_input.py:    all_pkg_names = dist.release_packages.keys()
ros_buildfarm/status_page.py:    pkg_names = [d.release_packages.keys() for d in distros]
scripts/doc/create_doc_task_generator.py:        set(pkg_names) | set(dist_file.release_packages.keys())
```

dirk-thomas:
> Left in the release_job.py source is a guard against configuring packages that are not yet in the rosdistro cache.

This is necessary since we need the package dependencies for the job generation. Without the cache we can't get them (reliably, up-to-date).

> But I do not see any guard for the reverse condition, where a package is in the rosdistro_cache but not part of the distribution file.

This would be for the time windows when a package has been removed from the `distribution.yaml` but the cache hasn't been updated yet.

> This wasn't a problem when our set of package names was coming from the distribution file, but do we need to reverse that guard now since it's possible for that condition to arise?

I don't expect it to happen often and it would fix itself shortly after in the next run by removing the job in question. But it still be good to address and not generate the job in the first place - good to catch.

clalancette:
> And if we need info from the distribution file to complete the task at hand, which appears to be the case:

This turns out to be exactly the case.  I ended up doing some additional testing here, and as it stands, we definitely need the package to be in the distribution file to successfully complete these checks and create the jobs.

Since that is the case, I think we should turn our attention to pointing out when a package is in the cache, but not in the distribution file.  As @nuclearsandwich points out above, we have the reverse check (when a package is in the distfile, but not in the distcache), so I think I'll try and add a new check for the current situation.  At least the next time this happens someone won't have to spend time debugging it again.

dirk-thomas:
I am not sure I understand the updated patch. It doesn't use the cache anymore and therefore doesn't address the problem we are facing: when a new package is added the reconfigure job fails to configure the job since it get an outdated answer from the GitHub cdn.

nuclearsandwich:
The updated PR, which fetches the distribution file from the cache rather than just fetching the package names to iterate seems like a much more complete solution. As far as I can tell from reading source for [get_distribution_cache](https://github.com/ros-infrastructure/rosdistro/blob/d794b2e9a7e699c00a5fbd3b2bf6869426158a19/src/rosdistro/__init__.py#L149-L171) and the [DistributionCache constructor](https://github.com/ros-infrastructure/rosdistro/blob/d794b2e9a7e699c00a5fbd3b2bf6869426158a19/src/rosdistro/distribution_cache.py#L65-L66) this change treats the rosdistro cache as the canonical source of truth for the current state of the rosdistro. While the distribution cache may yet lag behind the distribution in the Git repository, applying this change to other job types that make use of the distribution file will mean that updates are more atomic in that they happen only when the rosdistro cache updates.

dirk-thomas:
Now the patch completely bypasses the logic of [get_distribution_file](https://github.com/ros-infrastructure/ros_buildfarm/blob/e6af3c74fe6b61d0f80574723f0498c59c31907d/ros_buildfarm/config/__init__.py#L39) which performs build file specific filtering.

dirk-thomas:
> I will not have time to look at this again until next week, so I might suggest that we do the release without this patch and come back to it.

Ok, then we will move forward without this for now.

> So it seems like the thing to do would be to add API surface to retrieving the distribution file from the distribution cache with filtering, as opposed to adding filtering logic to each time we need to filter the distribution file.

Filtering is not a concept in the Python package `rosdistro` but purely exists within the `ros_buildfarm` package. The existing function `filter_distribution_files_by_tags` on the build file should be sufficient as is (and if not it can be refactored). But I don't see a need to introduce that concept into `rosdistro`.

dirk-thomas:
> I will not have time to look at this again until next week

@clalancette Any update on this?

dirk-thomas:
@clalancette We are considering a new release of this package. What is the status / estimated time line for this patch?

clalancette:
I haven't had time to come back to this.  I'd say that we should again punt on it, because I'm not sure when I'll really have time to address it.

dirk-thomas:
Closing due to inactivity. Please feel free to comment on the closed issue and the ticket can be reopened if necessary.

PR #524, opened 2018-03-15
Call super in JobValidationError to correcly print the error

Rayman:
Without this call, you get a meaningless JobValidationError. With this change it'll properly print the message when an exception occurs.


dirk-thomas:
Thank you for the improvement.

PR #527, opened 2018-03-19
Ignore the seccomp profile warning in `docker info`.

nuclearsandwich:
When using a custom seccomp profile a warning is printed in docker info.

Connects to https://github.com/ros-infrastructure/buildfarm_deployment/issues/176.


dirk-thomas:
@nuclearsandwich This change required the modified Groovy script to be approved. Not sure if this needs to be done in the deployment repo too somewhere.

PR #529, opened 2018-03-23
Create all_distro yaml

DLu:
Following up on #521, this PR creates an additional YAML file that makes it easier to aggregate info across distros/releases. The goal is to make the separate distros easier to merge recursively and reduce file size.

Previously we had this as the entry.
```
navigation:
  build_status:
    ubuntu:
      xenial:
        amd64: 
            build: 1.15.1-0xenial-20180215-152207-0800
            main: 1.15.1-0xenial-20171213-034534-0800
            test: 1.15.1-0xenial-20180215-152207-0800
        source:
            build: 1.15.1-0xenial
            main: 1.15.1-0xenial
            test: 1.15.1-0xenial}
  maintainers:
  - {email: davidvlu@gmail.com, name: David V. Lu!!}
  - {email: mferguson@fetchrobotics.com, name: Michael Ferguson}
  status: developed
  url: https://github.com/ros-planning/navigation.git
  version: 1.15.1-0
```

Now we change it to
```
navigation:
  kinetic:
    build_status:
      ubuntu:
        xenial:
          amd64: {build: 1.15.1-0, main: 1.15.1-0, test: 1.15.1-0}
          source: {build: 1.15.1-0, main: 1.15.1-0, test: 1.15.1-0}
    status: developed
    url: https://github.com/ros-planning/navigation.git
    version: 1.15.1-0
  maintainers: {davidvlu@gmail.com: David V. Lu!!, mferguson@fetchrobotics.com: Michael Ferguson}
```
Of note:
 * `build_status`, `status` `url` and `version` are dropped down a level underneath the rosdistro name.
 * The `build_status` only marks the version number, not the full debian name.
 * The maintainers list becomes a dictionary mapping email address to name.

~~After the distro/release specific yaml is made, it is recursively merged with `all_distros.yaml`~~

~~I acknowledge that the merging might end up being slow, given the size of the individual yamls, but it was a little fuzzy to me how to spin it off into a separate job.~~

### DIFF
--- a/ros_buildfarm/status_page.py
+++ b/ros_buildfarm/status_page.py
@@ -24,7 +24,6 @@ import shutil
 import sys
 import time
 import yaml
-import collections
 
 from .common import get_debian_package_name
 from .common import get_release_view_name
@@ -157,10 +156,7 @@ def build_release_status_page(
         os.mkdir(yaml_folder)
 
     yaml_filename = os.path.join(yaml_folder, 'ros_%s_%s.yaml' % (rosdistro_name, release_build_name))
-    summary = write_yaml(yaml_filename, ordered_pkgs, repos_data, rosdistro_name)
-
-    super_yaml_filename = os.path.join(yaml_folder, 'all_builds.yaml')
-    merge_yaml(super_yaml_filename, summary)
+    write_yaml(yaml_filename, ordered_pkgs, repos_data, rosdistro_name)
 
 
 def build_debian_repos_status_page(
@@ -1082,21 +1078,3 @@ def write_yaml(yaml_filename, ordered_pkgs, repos_data, rosdistro_name):
 
     with open(yaml_filename, 'w') as f:
         yaml.safe_dump(summary, f, allow_unicode=True)
-    return summary
-
-def dict_merge(dct, merge_dct):
-    for k, v in merge_dct.iteritems():
-        if (k in dct and isinstance(dct[k], dict) and isinstance(merge_dct[k], collections.Mapping)):
-            dict_merge(dct[k], merge_dct[k])
-        else:
-            dct[k] = merge_dct[k]
-
-def merge_yaml(super_yaml_filename, summary):
-    print("Updating aggregated yaml '%s':" % super_yaml_filename)
-    if os.path.exists(super_yaml_filename):
-        main_dict = yaml.load(open(super_yaml_filename))
-    else:
-        main_dict = {}
-    dict_merge(main_dict, summary)
-    with open(super_yaml_filename, 'w') as f:
-        yaml.safe_dump(main_dict, f, allow_unicode=True)
--- a/ros_buildfarm/status_page.py
+++ b/ros_buildfarm/status_page.py
@@ -24,6 +24,7 @@ import shutil
 import sys
 import time
 import yaml
+import collections
 
 from .common import get_debian_package_name
 from .common import get_release_view_name
@@ -156,7 +157,10 @@ def build_release_status_page(
         os.mkdir(yaml_folder)
 
     yaml_filename = os.path.join(yaml_folder, 'ros_%s_%s.yaml' % (rosdistro_name, release_build_name))
-    write_yaml(yaml_filename, ordered_pkgs, repos_data)
+    summary = write_yaml(yaml_filename, ordered_pkgs, repos_data, rosdistro_name)
+
+    super_yaml_filename = os.path.join(yaml_folder, 'all_builds.yaml')
+    merge_yaml(super_yaml_filename, summary)
 
 
 def build_debian_repos_status_page(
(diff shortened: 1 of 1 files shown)


dirk-thomas:
I am not sure I understand the approach of the patch. Each status jobs runs independently. How would the `all_builds.yaml` ever contain more than the currently processed distribution? It will always be empty when the job starts.

Even without that problem I don't think this approach is a good idea. While status page jobs are currently only being processed on a single executor they should still be parallelizable. Writing the same file from multiple potentially concurrent jobs looks problematic. Also the duplication of information seems unnecessary to me.

You can already automate the fetching of the existing `yaml` files. based on the information in the index you can construct the URLs to all yaml files and download the ones you want.

DLu:
I forgot that the script writes into a local folder and then recursively copies it to http://repositories.ros.org/status_page/yaml/ afterward. That would make it hard to aggregate. 

I pushed a change that removes the merging and just outputs the slightly cleaner format.

dirk-thomas:
You would rather not embed the ROS distro name into every item of the dict.

Instead I would suggest to use the following yaml structure:

```
packages:
   ... current content ...
release_build_name: '...'
rosdistro_name: '...'
```

That make the information machine readable. Any merging logic can then be applied to build a combined data structure.

DLu:
I'm trying to get to a structure like this:
```
move_base:
  indigo:
    build_status:
      ubuntu:
        trusty:
          amd64: {build: 1.12.14-0, main: 1.12.14-0, test: 1.12.14-0}
          armhf: {build: 1.12.14-0, main: 1.12.14-0, test: 1.12.14-0}
          i386: {build: 1.12.14-0, main: 1.12.14-0, test: 1.12.14-0}
          source: {build: 1.12.14-0, main: 1.12.14-0, test: 1.12.14-0}
    status: maintained
    url: https://github.com/ros-planning/navigation/tree/indigo-devel
    version: 1.12.14-0
  kinetic:
    build_status:
      debian:
        jessie:
          amd64: {build: 1.14.2-0, main: 1.14.2-0, test: 1.14.2-0}
          arm64: {build: 1.14.2-0, main: 1.14.2-0, test: 1.14.2-0}
          source: {build: 1.14.2-0, main: 1.14.2-0, test: 1.14.2-0}
      ubuntu:
        xenial:
          amd64: {build: 1.14.2-0, main: 1.14.2-0, test: 1.14.2-0}
          arm64: {build: 1.14.2-0, main: 1.14.2-0, test: 1.14.2-0}
          armhf: {build: 1.14.2-0, main: 1.14.2-0, test: 1.14.2-0}
          i386: {build: 1.14.2-0, main: 1.14.2-0, test: 1.14.2-0}
          source: {build: 1.14.2-0, main: 1.14.2-0, test: 1.14.2-0}
    status: maintained
    url: https://github.com/ros-planning/navigation/tree/kinetic-devel
    version: 1.14.2-0
  lunar:
    build_status:
      debian:
        stretch:
          amd64: {build: 1.15.1-0, main: 1.15.1-0, test: 1.15.1-0}
          arm64: {build: 1.15.1-0, main: 1.15.1-0, test: 1.15.1-0}
          source: {build: 1.15.1-0, main: 1.15.1-0, test: 1.15.1-0}
      ubuntu:
        xenial:
          amd64: {build: 1.15.1-0, main: 1.15.1-0, test: 1.15.1-0}
          arm64: {build: 1.15.1-0, main: 1.15.1-0, test: 1.15.1-0}
          armhf: {build: 1.15.1-0, main: 1.15.1-0, test: 1.15.1-0}
          source: {build: 1.15.1-0, main: 1.15.1-0, test: 1.15.1-0}
    status: maintained
    url: https://github.com/ros-planning/navigation/tree/lunar
    version: 1.15.1-0
  maintainers: {davidvlu@gmail.com: David V. Lu!!, mferguson@fetchrobotics.com: Michael Ferguson}
```
which is a simple recursive merge of all the individual files, rather than special logic for the structure you suggested.

mikaelarguedas:
I have a use case where the previous values (with the full version name) would be valuable. When creating docker images for ROS, we currently use the variants to define the scope of the images. These packages are almost never released (because not changing) so the version number (MAJOR.MINOR.PATCH-BLOOMDEBINC) doesnt change. As a consequence the images don't get rebuilt on syncs. If we were to provide the full version (including the timestamp) we could easily detect changes and generate dockerfiles accordingly.
If a machine readable file provides this for an entire distro, we could move away from having to poke the repository itself to get the version of the packages in it.

dirk-thomas:
> which is a simple recursive merge of all the individual files, rather than special logic for the structure you suggested.

I understand your use case. But I think the structure I mentioned is more generic. There might be use case beyond yours and in your yaml file the question "what packages exist in Kinetic" can't be answered easily.

Therefore I think a structure following the actual topology of the entities (a ROS distro contain multiple ROS packages) makes more sense here. The merging logic for your case would then only be a few lines of Python code.

> If a machine readable file provides this for an entire distro, we could move away from having to poke the repository itself to get the version of the packages in it.

The index.yaml already lets you know what release build files exist. And for each of them you can get the yaml file.

mikaelarguedas:
> The index.yaml already lets you know what release build files exist. And for each of them you can get the yaml file.

I guess I missed something... if this gets merged, the yaml file will not contain the full version number anymore but only `MAJOR.MINOR.PATCH-BLOOMDEBINC`, right?

DLu:
Fair points.

mikaelarguedas:
> The index.yaml already lets you know what release build files exist. And for each of them you can get the yaml file.

@dirk-thomas can you clarify which index.yaml you are referring to ? the one from [ros_buildfarm_config](https://github.com/ros-infrastructure/ros_buildfarm_config/blob/90e681d3472b41ffc726cc2d7ede803b78ae57aa/index.yaml) ?

dirk-thomas:
> can you clarify which index.yaml you are referring to ? the one from ros_buildfarm_config ?

Yes.

mikaelarguedas:
ok that's what I thought. The part I was missing was `'ros_%s_%s.yaml' % (rosdistro_name, release_build_name)`

:+1:

PR #530, opened 2018-03-23
Fix duplicate apt repos

dirk-thomas:
Fixes #509. Connect to #509.

Only add the entries when they are not already there (not considering commented entries for that comparison).

PR #532, opened 2018-03-26
Use egrep to find repository components in arbitrary positions.

nuclearsandwich:
This is a proposed enhancement of #530 

The existing implementation assumes that the repositories in question are part of the sources.list on a line of their own when they may exist as one component among several on the same line. This trades the fixed string check for regular expression that will find the desired component regardless of its position among multiple components or if it is on a
line by itself.

@dirk-thomas did you have a Debian test case that you were using for #530? I've only tested this using generate_devel_script.py for indigo fanuc on ubuntu trusty.

### DIFF
--- a/ros_buildfarm/templates/snippet/add_additional_repositories.Dockerfile.em
+++ b/ros_buildfarm/templates/snippet/add_additional_repositories.Dockerfile.em
@@ -8,9 +8,9 @@ archive_commands = []
 old_releases_commands = []
 for distribution, archive_type in product((os_code_name, os_code_name + '-updates', os_code_name + '-security'), ('deb', 'deb-src')):
     archive_entry = '%s http://archive.ubuntu.com/ubuntu/ %s multiverse' % (archive_type, distribution)
-    archive_pattern = '%s http://archive.ubuntu.com/ubuntu/ %s ([-a-z]+ )*multiverse( [-a-z])*' % (archive_type, distribution)
+    archive_pattern = '%s http://archive\.ubuntu\.com/ubuntu/ %s ([-a-z]+ )*multiverse( [-a-z])*' % (archive_type, distribution)
     old_releases_entry = '%s http://old-releases.ubuntu.com/ubuntu/ %s multiverse' % (archive_type, distribution)
-    old_releases_pattern = '%s http://old-releases.ubuntu.com/ubuntu/ %s ([-a-z]+ )*multiverse( [-a-z]+)*' % (archive_type, distribution)
+    old_releases_pattern = '%s http://old-releases\.ubuntu\.com/ubuntu/ %s ([-a-z]+ )*multiverse( [-a-z]+)*' % (archive_type, distribution)
     archive_commands.append('(grep -q -E -x -e "%s" /etc/apt/sources.list || echo "%s" >> /etc/apt/sources.list)' % (archive_pattern, archive_entry))
     old_releases_commands.append('(grep -q -E -x -e "%s" /etc/apt/sources.list || echo "%s" >> /etc/apt/sources.list)' % (old_releases_pattern, old_releases_entry))
 }@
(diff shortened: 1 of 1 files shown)


dirk-thomas:
The problem with interpreting the previous fixed string as a regex is that the fixed part does contain characters which have special meaning in a regex. Without escaping them the result is not necessarily correct. Therefore I don't think we should use a regex here.

> did you have a Debian test case that you were using for #530?

http://build.ros.org/job/Kbin_djv8_dJv8__cob_substitute__debian_jessie_arm64__binary/

nuclearsandwich:
> The problem with interpreting the previous fixed string as a regex is that the fixed part does contain characters which have special meaning in a regex.

That's a fair point, although because I've split the pattern lines into a separate variable we could apply that escaping. I think the assumption that these repositories would always be added on unique lines is an unnecessarily brittle one even if we have the ability to enforce it for our own images at the moment.

nuclearsandwich:
I've updated this PR to quote literal dots in patterns. I also noticed that a sed substitution was using wildcard dots so I quoted those for good measure.

While reading the patterns closely, I noticed that there was a doubled trailing slash on one line, first introduced in https://github.com/ros-infrastructure/ros_buildfarm/commit/c95cbb3a41e80c413b671f1fea5f7167a48ef997 with no explanation. I figured I'd remove it and thought to check the current text of those lines in our live images.

```
% docker run -ti osrf/ubuntu_armhf:trusty cat /etc/apt/sources.list
deb http://ports.ubuntu.com trusty main restricted universe multiverse
deb http://ports.ubuntu.com trusty-updates main restricted universe multiverse
deb http://ports.ubuntu.com trusty-backports main restricted universe multiverse
```
shows no trailing slash at all. Rather than try and tightly couple the string literal to the particular upstream image. I've just opted to make trailing slashes optional for all patterns. A survey of upstream ubuntu and debian docker images shows that ubuntu defaults to a trailing slash while debian does not.

ros_buildfarm/templates/snippet/add_additional_repositories.Dockerfile.em
```diff
@@ -30,10 +33,11 @@ RUN @(' && '.join(commands))
...
     entry = 'deb http://httpredir.debian.org/debian %s %s' % (os_code_name, component)
-    commands.append('(grep -q -F -x -e "%s" /etc/apt/sources.list || echo "%s" >> /etc/apt/sources.list)' % (entry, entry))
+    pattern = 'deb http://httpredir\.debian\.org/debian/? %s ([-a-z] )*%s( [-a-z])*' % (os_code_name, component)
+    commands.append('(grep -q -E -x -e "%s" /etc/apt/sources.list || echo "%s" >> /etc/apt/sources.list)' % (pattern, entry))
 }@
 RUN @(' && '.join(commands))
 # Hit cloudfront mirror because of corrupted packages on fastly mirrors (https://github.com/ros-infrastructure/ros_buildfarm/issues/455)
 # You can remove this line to target the default mirror or replace this to use the mirror of your preference
-RUN sed -i 's/httpredir.debian.org/cloudfront.debian.net/' /etc/apt/sources.list
+RUN sed -i 's/httpredir\.debian\.org/cloudfront.debian.net/' /etc/apt/sources.list
```
dirk-thomas: Is there a reason to only escape two of the dots?

nuclearsandwich: > Is there a reason to only escape two of the dots?

This is a sed substitution string so only the portion within the first // is treated as a regular expression. The text  after the second `/`, `cloudfront.debian.net` is the replacement string and is already a literal (if it included back-references like `\1` it wouldn't be truly a literal, but the '.' characters would still be literal).

dirk-thomas: Oh, I see. Too much different tools going on :wink:

PR #534, opened 2018-03-28
remove using the test_depend from binary jobs

dirk-thomas:
When releasing a new version of the `ros2cli` repo into Ardent the binarydeb job of e.g. `ros1_bridge` was being retriggered by the rebuild of `ros2run` since it has a [test_depend](https://github.com/ros2/ros1_bridge/blob/f8a6b6dce79937380b341d943d667ccde5287e5e/package.xml#L33) on it.

This patch removes considering `test_depend` for dependencies between binarydeb jobs. While this looks like the "right" thing to be done I would like to get feedback what others think or if they see any problems with this change.

### DIFF
--- a/ros_buildfarm/release_job.py
+++ b/ros_buildfarm/release_job.py
@@ -547,14 +547,15 @@ def _get_direct_dependencies(pkg_name, dist_cache, pkg_names):
         return None
     pkg_xml = dist_cache.release_package_xmls[pkg_name]
     pkg = parse_package_string(pkg_xml)
+    # downstream binary packages don't change if an upstream package is rebuild
+    # if it only has a test dependency therefore don't use test dependencies
     depends = set([
         d.name for d in (
             pkg.buildtool_depends +
             pkg.build_depends +
             pkg.buildtool_export_depends +
             pkg.build_export_depends +
-            pkg.exec_depends +
-            pkg.test_depends)
+            pkg.exec_depends)
         if d.name in pkg_names])
     return depends
 


mikaelarguedas:
I think thats fine, if a package passed the tests when the binary was built (via dh_auto_test), it shouldn't be rebuilt on test dependency change.

dirk-thomas:
> if a package passed the tests when the binary was built

Binarydeb jobs do not any kind of testing?

mikaelarguedas:
In ROS1 tests are explicitely ignored in binaries. In ROS2 (because we use standard cmake variables for enabling / disabling testing) tests are ran on binary jobs, below an extract of the [rclcpp binary job](http://build.ros2.org/view/Abin_uX64/job/Abin_uX64__rclcpp__ubuntu_xenial_amd64__binary/11).
We currently ignore the results of the tests as in ROS1 and ROS2 we run `dh_auto_test || true`
<details>

```
83% tests passed, 4 tests failed out of 24

Label Time Summary:
copyright     =   0.61 sec (1 test)
cppcheck      =   1.74 sec (1 test)
cpplint       =  11.16 sec (1 test)
gmock         =   0.13 sec (1 test)
gtest         =   4.65 sec (18 tests)
lint_cmake    =   1.37 sec (1 test)
linter        =  15.71 sec (5 tests)
uncrustify    =   0.82 sec (1 test)

Total Test time (real) =  20.52 sec

The following tests FAILED:
	 20 - copyright (Failed)
	 22 - cpplint (Failed)
	 23 - lint_cmake (Failed)
	 24 - uncrustify (Failed)
Errors while running CTest
Makefile:85: recipe for target 'test' failed
make[2]: *** [test] Error 8
make[2]: Leaving directory '/tmp/binarydeb/ros-ardent-rclcpp-0.4.0/obj-x86_64-linux-gnu'
dh_auto_test: make -j1 test ARGS+=-j1 returned exit code 2```

dirk-thomas:
> In ROS2 (because we use standard cmake variables for enabling / disabling testing) tests are ran on binary jobs, below an extract of the rclcpp binary job.

@nuclearsandwich @wjwwood Can you comment if that is intentional? That seems to waste quite some resource on every binarydeb build. Should we "get rid" of that if we don't care about the result anyway?

dirk-thomas:
> I might suggest adding a comment that this is specifically the dependencies for the jenkins chaining then.

Since the change was in the release module I didn't think that would be necessary. I added a comment to it anyway to clarify: 36bd2e3e084cd88e0090c8c483ed7aa5a471e6df

mikaelarguedas:
> Can you comment if that is intentional? That seems to waste quite some resources on every binarydeb build. Should we "get rid" of that if we don't care about the result anyway?

Currently in ROS 1 we have the inconsistency in that the pure cmake packages run their tests but not the catkin ones.

As running the tests of a package at packaging time is standard debian policy, I think that we should rather aim to run the tests and not ignore the results rather than disabling the tests. This will get us closer to comply with debian rules.

nuclearsandwich:
> @nuclearsandwich @wjwwood Can you comment if that is intentional?

I don't have any notes about intent one way or the other. It's likely that I had tests running for my own sanity while I was bootstrapping and it never occurred to me to disable them.

> That seems to waste quite some resources on every binarydeb build. Should we "get rid" of that if we don't care about the result anyway?

I think that probably merits a separate discussion. A PR to bloom would be needed change the behavior and it would only take effect for newly bloomed packages. But I would consider running tests on binaries a potential improvement and I have very imprecise meeting notes from r2b3 release window about wanting to move in the direction of observing test results during binary builds even if we're not currently doing so.

dirk-thomas:
> I think that we should rather aim to run the tests and not ignore the results rather than disabling the tests.

If that is the goal we can't merge this PR.

Instead the test results should not be ignored - otherwise the resource usage isn't justified.

Waiting for a consensus on what people think we should do...

nuclearsandwich:
> If that is the goal we can't merge this PR.

I'm not sure that's strictly accurate. Merging this PR and building tests in binary jobs would introduce the possibility that a change in a test dependency, which does not trigger a new binary job could cause a break in a future binary job that would need to be triaged back to the change in the test dependency. If we had dev/PR jobs for the ROS 2 buildfarm, it would be easier to catch such issues earlier in the release cycle.

 > Instead the test results should not be ignored - otherwise the resource usage isn't justified.

Is that really the case? It would certainly be better to use information that we paid a computational cost for but even if the buildfarm doesn't fail or mark a build unstable having test output in the logs if needed still has some value.

dirk-thomas:
The remaining question is still: do we want to build packages based on test dependencies. Conceptionally it is not necessary for the result of the Debian package (aka the Debian package must not change if a test dependency is being rebuild - otherwise the package would have specified a build or run dependency). It could of course influence the test result though - I am just not convinced that this is the right location to check for that.

> A simple way to do that would be using `|| echo "uniquestring_postbuild_test_failed"` which will give an easy console output for a jenkins postbuild step to detect and mark the build as unstable.

The part which makes this not "that simple" is that the command line is part of the Debian rules file generated by `bloom` and the "catching" would need to happen in the `ros_buildfarm` repo. Currently those two don't share knowledge like that and I am hesitant to add such a dependency.

dirk-thomas:
> I would be fine not rebuilding packages automatically based on test dependency updates. It will provide a window where a change could cause a delayed/deferred regression, but I think the risk of that is relatively low and that the benefit of faster rebuilds outweights the lack of completely traceable causes of regressions.

This sounds perfectly reasonable to me. This will mean this patch gets merged as-is. If I don't hear any opposing comments I will merge this by the end of the day (since it already has two approvals).

---

Imo warning users (or even failing the build) sounds "disruptive" to me. Since I don't think we should just "enable" that I think the resources spend on testing and unnecessary since the result is not visible until anyone looks into the build output. But since the opinions seem to not align on this question I will leave the possibility of changing the rule (in whatever way) to someone else to follow up on.

PR #536, opened 2018-04-03
update changelog for upcoming major version

dirk-thomas:


nuclearsandwich:
The nested bullets don't appear to render correctly with GitHub's rst renderer. Link: https://github.com/ros-infrastructure/ros_buildfarm/blob/34897d3bb010fedcd2f2fb432702a5297a866db2/CHANGELOG.rst

dirk-thomas:
Added the missing empty line in between.

Issue #539, opened 2018-04-26
sometimes could not find cached deb package for `test_depend`?


k-okada:
I have following error on buildfirm (always on arm and sometimes on amd64/i386), It seems they can not find deb package within `test_depends` 

- 51 agent-d0968e8a NG
- 52 agent-6b80438 OK
- 53 agent-9da19ae8 OK
- 54  agent-0b3a0b30 NG
- 55 agent-e26dbb61  NG
- 56 agent-e26dbb61 OK

http://build.ros.org/job/Kbin_uX64__nextage_calibration__ubuntu_xenial_amd64__binary/57/console
```
# BEGIN SUBSECTION: append build timestamp
23:23:25 dpkg-parsechangelog: warning:     debian/changelog(l5): found trailer where expected start of change data
23:23:25 LINE:  -- TORK <dev@opensource-robotics.tokyo.jp>  Tue, 16 Jan 2018 00:00:00 -0000
23:23:25 Invoking 'debchange -v 0.8.4-0xenial-20180416-222325-0800 -p -D xenial -u high -m Append timestamp when binarydeb was built.' in '/tmp/binarydeb/ros-kinetic-nextage-calibration-0.8.4'
23:23:25 debchange: warning:     debian/changelog(l5): found trailer where expected start of change data
23:23:25 LINE:  -- TORK <dev@opensource-robotics.tokyo.jp>  Tue, 16 Jan 2018 00:00:00 -0000
23:23:25 # END SUBSECTION
23:23:28 Looking for the '.dsc' file of package 'ros-kinetic-nextage-calibration' with version '0.8.4-0'
23:23:29 Traceback (most recent call last):
23:23:29   File "/usr/lib/python3/dist-packages/apt/cache.py", line 194, in __getitem__
23:23:29     return self._weakref[key]
23:23:29   File "/usr/lib/python3.5/weakref.py", line 131, in __getitem__
23:23:29     o = self.data[key]()
23:23:29 KeyError: 'ros-kinetic-nextage-gazebo'
23:23:29 
23:23:29 During handling of the above exception, another exception occurred:
23:23:29 
23:23:29 Traceback (most recent call last):
23:23:29   File "/usr/lib/python3/dist-packages/apt/cache.py", line 198, in __getitem__
23:23:29     rawpkg = self._cache[key]
23:23:29 KeyError: 'ros-kinetic-nextage-gazebo'
23:23:29 
[...]

PR #540, opened 2018-04-26
Revert "remove using the test_depend from binary jobs (#534)"

mikaelarguedas:
This reverts commit 0ab65b0f0e7a51d04c24503a152b40423aecc347.

See https://github.com/ros-infrastructure/ros_buildfarm/issues/539 for details


mikaelarguedas:
As this was released in 2.0.0, we may want to release a new version with this reverted.
Unless we want to wait to have a better resolution to the issue?

dirk-thomas:
I am fine with reverting the change. We should add a comment though why we need these test dependencies for future understanding.

PR #541, opened 2018-04-28
Add devel job test statistics collation

gavanderhoorn:
This PR adds the backend for ros-infrastructure/roswiki#232. ros-infrastructure/roswiki#246 adds the frontend to `roswiki`.

This extends all devel jobs (and only those) with two additional post-build actions:

 1. a Groovy script that collates test results from the current build and a configurable nr of previous builds (currently: 5). The results are then serialised to a yaml file in a `collated_test_stats` sub directory of the `$workspace`.
 1. a "Send build artefacts over SSH" step that scps the resulting yaml to a sub dir of the `docs` dir on the repo host (`docs/$distro/devel_jobs/$repo`).

An example of the output of the Groovy script:

```yaml
dev_job_data:
  history:
  - job_id: 13
    uri: job/Idev__actionlib__ubuntu_trusty_amd64/13/
    stamp: 1524852610.998
    result: success
    tests:
      skipped: 0
      failed: 0
      total: 52
  - job_id: 12
    uri: job/Idev__actionlib__ubuntu_trusty_amd64/12/
    stamp: 1523977812.967
    result: success
    tests:
[...]
  latest_build:
    skipped: 0
    failed: 0
    total: 52
  base_url: http://buildfarm/
  total_builds: 12
  job_health: 100
  job_health_icon: icon-health-80plus
```

Some things to discuss / some comments:

 1. these steps are added as post-build steps as they need to be run *after* the `publisher_xunit` step, which is already a post-build activity.
 1. the Groovy script is currently embedded in the `devel_job` template instead of stored in a separate snippet. It can be made a snippet, but seeing as there is very little re-use value here, I opted not to.
 1. to keep this patch small, files are currently copied to `docs` on the repo host instead of a separate location. For this minor extension we felt this would be ok for now, but if/when more extensive scanning/testing stats are to be made available to the wiki infrastructure, a location outside of the `docs` hierarchy might be more appropriate. If that is not acceptable it can be changed.
 1. the `publisher_groovy-postbuild` snippet seems to set `behavior` to `2` ([here](https://github.com/ros-infrastructure/ros_buildfarm/blob/2e3afa9a01ab6f99b9de13480023ffef3f4dc5b8/ros_buildfarm/templates/snippet/publisher_groovy-postbuild.xml.em#L6)): this corresponds to *Mark build as failed* for the "If the script fails" setting. The Groovy script should probably not fail the build, so the snippet would probably need to be updated to accept an argument for the `behavior` element (I can do that if desired).
 1. we opted to add these steps to the devel jobs. An alternative could be to create separate jobs for this, which would only run the script and scp / rsync the results. That alternative was not implemented to avoid adding many hundreds of new (small) jobs to the farm.

Finally; this has been tested on a local instance of both the ROS wiki and `ros_buildfarm`, but only on a small test set of repositories. I'm fairly certain most of the failure modes have been covered, and no error or warning should fail the script (only messages are printed to the console). However, yaml serialisation (`dump(..)`) has not been tested with pkgs not conforming to naming guidelines (ie: containing diacritics or other 'strange' characters). I'm willing to do that if examples of such pkgs are available.

---

This is the first of (hopefully) many contributions to `ros_buildfarm` to come out of the [Software Quality Working Group](https://discourse.ros.org/c/quality) of the ROSIN EU H2020 project.


ros_buildfarm/templates/devel/devel_job.xml.em
```diff
@@ -245,6 +256,106 @@ if pull_request:
...
+        '',
+        'try {',
+        '  def data = [',
+        '    "history" : []',
+        '  ]',
+        '',
+        '  // gather info on tests of current build',
+        '  def tresult = manager.build.getAction(hudson.tasks.junit.TestResultAction.class)?.result',
+        '  if (tresult)',
+        '  {',
```
dirk-thomas: Nitpick: please use "cuddled" open / close curly braces (as some of the code already does).

ros_buildfarm/templates/devel/devel_job.xml.em
```diff
@@ -245,6 +256,106 @@ if pull_request:
...
+        '      "total" : tresult.totalCount',
+        '    ]',
+        '  }',
+        '  else',
+        '  {',
+        '    manager.listener.logger.println("No test result action for last build, skipping gathering statistics for it.")',
+        '  }',
+        '',
+        '',
+        '  // get acces to the job of the running build',
```
dirk-thomas: Spelling: accesS

ros_buildfarm/templates/devel/devel_job.xml.em
```diff
@@ -245,6 +256,106 @@ if pull_request:
...
+        '  def job_name = manager.build.getEnvironment(manager.listener).get(\'JOB_NAME\')',
+        '  manager.listener.logger.println("Collating test statistics for \'${job_name}\'.")',
+        '  def job = Jenkins.instance.getItem(job_name)',
+        '  if (job == null)',
+        '  {',
+        '    manager.listener.logger.println("No such job: \'${job_name}\'.")',
+        '    return',
```
dirk-thomas: Interesting, just out of curiosity: did you ever run into this being the case during testing?

ros_buildfarm/templates/devel/devel_job.xml.em
```diff
@@ -245,6 +256,106 @@ if pull_request:
...
+        '  data.job_health = job.getBuildHealth().getScore()',
+        '  data.job_health_icon = job.getBuildHealth().getIconClassName()',
+        '',
+        '  // retrieve info on last N builds of this job',
+        '  job.builds.take(num_build_hist).each { b ->',
+        '    tresult = b.getAction(hudson.tasks.junit.TestResultAction.class)?.result',
+        '    if (tresult)',
+        '    {',
+        '      data.history << [',
+        '        "job_id" : b.id as Integer,',
```
dirk-thomas: `job_id` doesn't sounds right to. I would propose either `build_id` or `build_number`.

ros_buildfarm/templates/devel/devel_job.xml.em
```diff
@@ -245,6 +256,106 @@ if pull_request:
...
+        'try {',
+        '  def data = [',
+        '    "history" : []',
+        '  ]',
+        '',
+        '  // gather info on tests of current build',
+        '  def tresult = manager.build.getAction(hudson.tasks.junit.TestResultAction.class)?.result',
+        '  if (tresult)',
+        '  {',
+        '    data.latest_build = [',
```
dirk-thomas: Do we need to store `latest_build` or is the same information also available under the highest build number?

dirk-thomas:
> the Groovy script is currently embedded in the `devel_job` template instead of stored in a separate snippet.

> to keep this patch small, files are currently copied to docs on the repo host instead of a separate location.

> the `publisher_groovy-postbuild` snippet seems to set `behavior` to `2`: this corresponds to *Mark build as failed* for the "If the script fails" setting.

> we opted to add these steps to the devel jobs.

Sounds all good to me as it is.

Regarding the third item: I think it is good to fail visibly (which shouldn't happen) rather than silently being skipped. If that turns out to be happening we can still switch it later.

ros_buildfarm/templates/devel/devel_job.xml.em
gavanderhoorn: No, I haven't. But I'm a rather defensive programmer most of the time. I can take it out if preferred.

ros_buildfarm/templates/devel/devel_job.xml.em
gavanderhoorn: An earlier version of this script did just refer to the highest build id in the `history` list.

But the history is collected later in the script, and could fail. If it does, we would not have any test data, not even for the current job. The test results of the current job are retrieved directly via the `TestResultAction`.

This is just a choice though: if we feel failure is failure (and we don't want to separate these two things) then having `latest_build` point to `data.history[0]` (or even removing `latest_build` entirely) is an option.

gavanderhoorn:
@dirk-thomas wrote:
> Regarding the third item: I think it is good to fail visibly (which shouldn't happen) rather than silently being skipped. If that turns out to be happening we can still switch it later.

Ok. I'll leave this as-is for now then.

I didn't want to be the cause of (potentially) hundreds of builds failing due to some programming error on my/our side.

ros_buildfarm/templates/devel/devel_job.xml.em
dirk-thomas: No need, better be safe than sorry :+1:

ros_buildfarm/templates/devel/devel_job.xml.em
dirk-thomas: Sounds good to me.

dirk-thomas:
Looking at this again I think there is one conflict left. There can be multiple devel jobs for the same repo in a ROS distribution. E.g. `ros_comm` has devel jobs for Ubuntu as well as Debian.

Therefore I would suggest to introduce a configuration option (which we can enable for `source` build file targeting Ubuntu) and only create and upload the information when the flag is being set.

gavanderhoorn:
You're suggesting to parameterise the build, right?

So all dev jobs (but not PR jobs) get the same configuration, but only for specific jobs the `collate_test_stats` parameter is set to `true`, which would make the bash snippet, the groovy script and the scp steps run?

dirk-thomas:
Sorry, I should have been more specific. The idea is not to add a parameter to the job. Instead a configuration being added to the source build file (https://github.com/ros-infrastructure/ros_buildfarm_config/blob/f79c0d3616ddec6a34d4740e6e3f95d827b7f433/melodic/source-build.yaml).

The generated job would (if the flag is set in the config file) contain the logic of this patch. And if the flag is not set not contain the additional (post) build steps.

gavanderhoorn:
Ah. Ok. Like how `pull_request` is used/handled.

Do I understand things correctly in that adding such a parameter needs changes to the various Python scripts involved (ie: `devel_job.py` et al.)?

gavanderhoorn:
Would something like this suffice?

```diff
diff --git a/ros_buildfarm/devel_job.py b/ros_buildfarm/devel_job.py
index 0c613a2..48331a3 100644
--- a/ros_buildfarm/devel_job.py
+++ b/ros_buildfarm/devel_job.py
@@ -391,6 +391,8 @@ def _get_devel_job_config(
         'timeout_minutes': build_file.jenkins_job_timeout,

         'git_ssh_credential_id': config.git_ssh_credential_id,
+
+        'collate_test_stats': build_file.collate_test_stats,
     }
     job_config = expand_template(template_name, job_data)
     return job_config
diff --git a/ros_buildfarm/templates/devel/devel_job.xml.em b/ros_buildfarm/templates/devel/devel_job.xml.em
index 1fa40a2..6e4507d 100644
--- a/ros_buildfarm/templates/devel/devel_job.xml.em
+++ b/ros_buildfarm/templates/devel/devel_job.xml.em
@@ -235,7 +235,7 @@ if pull_request:
         'fi',
     ]),
 ))@
-@[if not pull_request]@
+@[if (not pull_request) and collate_test_stats]@
 @(SNIPPET(
     'builder_shell',
     script='\n'.join([
@@ -256,7 +256,7 @@ if pull_request:
     'publisher_xunit',
     pattern='catkin_workspace/test_results/**/*.xml',
 ))@
-@[if not pull_request]@
+@[if (not pull_request) and collate_test_stats]@
 @(SNIPPET(
     'publisher_groovy-postbuild',
     script='\n'.join([
```

IIUC, each OS target get its own build file, so the new `key:val` pair could just be added without any nesting/scoping.

dirk-thomas:
> Like how `pull_request` is used/handled.

Yeah, exactly.

> Do I understand things correctly in that adding such a parameter needs changes to the various Python scripts involved (ie: `devel_job.py` et al.)?

After a brief look I think this [function](https://github.com/ros-infrastructure/ros_buildfarm/blob/940d8a418a25674fb96d33e1c04a1b1a1530a3be/ros_buildfarm/devel_job.py#L323) needs to get the flag from the `build_file` and pass it to the template. Then the template can use the flag as a condition wrapping the new code segments. And the [build file itself](https://github.com/ros-infrastructure/ros_buildfarm/blob/940d8a418a25674fb96d33e1c04a1b1a1530a3be/ros_buildfarm/config/build_file.py) will need to be updated to read the yaml value and store it in the class (plus adding the new option to the docs).

gavanderhoorn:
I assumed/presumed the `build_file` was just a dict that got initialised directly from the yaml. That would have been convenient :)

I'll take a closer look and see if I can figure it out with your pointers.

dirk-thomas:
> I assumed/presumed the `build_file` was just a dict that got initialised directly from the yaml. That would have been convenient :)

Hope dies last :wink:

gavanderhoorn:
This does seem a bit like a work-around though: ideally we'd collate test statistics for all builds, over all jobs for a particular package.

An alternative to what you suggest might be to include some discriminator in the filename of the collated yaml file, or use it in the path where the yaml file gets stored.

The wiki macro would then check all files and combine the results in some meaningful way.

gavanderhoorn:
> Also almost no packages have additional devel jobs. There are only very few package doing that. Not sure if the extra effort is worth it.

You're referring to the 'extra effort' to support collating stats for all dev jobs of a repository here, right?

I can add the 'disable globally from a buildfile' for this PR, but per-repository disabling should be added later.

dirk-thomas:
> You're referring to the 'extra effort' to support collating stats for all dev jobs of a repository here, right?

Yes.

> I can add the 'disable globally from a buildfile' for this PR, but per-repository disabling should be added later.

Per build file is ok. (Per repo would require a REP update so maybe not necessary?) Please make the default "disabled". I am happy to enable the feature for all ROS distros but then we can do it one by one while watching for changes.

gavanderhoorn:
So summarising:

 1. add a key to the buildfiles in `ros_buildfarm_config` called `collate_test_stats` that is a boolean where:
    - `true`: results in the additional steps proposed in this PR being added to the devel job
    - `false`: does not add the steps
 1. extend `build_file.py` to load this key from the buildfile
 1. extend `_get_devel_job_config(..)` to pass the flag so it can be added to the `job_data` dict
 1. extend `doc/configuration_options.rst` with info on this new setting

re: 2: is it acceptable to use `data.get('<key name>', False)` to get a default 'disabled', or should the same structure as is used for `test_commits` ([here](https://github.com/ros-infrastructure/ros_buildfarm_config/blob/f79c0d3616ddec6a34d4740e6e3f95d827b7f433/melodic/source-build.yaml#L53-L54)) be created?

re: 3: is directly using `build_file.<key name>` not acceptable (similar to how that is done for `notify_maintainers` ([here](https://github.com/ros-infrastructure/ros_buildfarm/blob/940d8a418a25674fb96d33e1c04a1b1a1530a3be/ros_buildfarm/devel_job.py#L387)))?

Just want to make sure I'm understanding things correctly before implementing it.

dirk-thomas:
> So summarising

:+1: 

> re: 2
> re: 3

For `test_commits` a dictionary with `default` is being used to clarify that the build file does not necessarily make the final decision (since the flag can be overridden on a per-repo base).

So the questions comes down to if you want a per repo choice in the future. If not (which I think is sufficient) you would just define the property in the `BuildFile` like `notify_maintainers`.

gavanderhoorn:
> So the questions comes down to if you want a per repo choice in the future. If not (which I think is sufficient) you would just define the property in the `BuildFile` like `notify_maintainers`.

For this PR I'll keep it a per-buildfile property.

I'll stop commenting and push what I have now.

gavanderhoorn:
Added the option to just the source build file, documented it and used `build_file` directly to populate the item in the `job_data` dict.

dirk-thomas:
@tfoote @nuclearsandwich Do you agree that it is fine to leave the source build file version at 2 even though this patch introduces a new flag?

@nuclearsandwich Do you prefer to deploy this change to a test farm first or are we confident enough to merge this and then enable the flag for e.g. `melodic`. And if that goes well also enable for the other ROS distros?

tfoote:
I'm find with leaving the version the same as long as the old parser isn't strict about rejecting it if there's unknown flags.

dirk-thomas:
@nuclearsandwich Ping

nuclearsandwich:
> Do you agree that it is fine to leave the source build file version at 2 even though this patch introduces a new flag?

I think it's fine to add this without a format bump. If we revise the format in the future we should consider whether or not we want or need to make the value required (I haven't formed an opinion either way, I just think it should be considered when the time comes.)


> Do you prefer to deploy this change to a test farm first or are we confident enough to merge this and then enable the flag for e.g. melodic. And if that goes well also enable for the other ROS distros?

If @gavanderhoorn has been testing this change on a local farm I think we can start with a staircase rollout (start with melodic or lunar and expand as we gain confidence) on build.ros.org.

dirk-thomas:
Sounds good to me.

merging this one and enabling it for Melodic on Bionic: ros-infrastructure/ros_buildfarm_config#113

gavanderhoorn:
Thanks :+1:

dirk-thomas:
> Who should I prod to review ros-infrastructure/roswiki#246 ?

I requested a review from @dhood which is nowadays maintaining that repo.It might be until the flag has been enabled and a few jobs have generated the necessary data until the PR will be merged.

dirk-thomas:
@gavanderhoorn This is the first build with the new feature enabled: http://build.ros.org/view/Mdev/job/Mdev__catkin__ubuntu_bionic_amd64/17/console Any idea why it is failing?

gavanderhoorn:
https://github.com/JFrogDev/artifactory-scripts/issues/20#issuecomment-282308241 seems to indicate this is a "common Jenkins issue". What I don't understand though is why I've not had to do any of that.

gavanderhoorn:
Perhaps setting `behavior` to `0` for now would be good? That would at least keep builds from being marked as `failed`.

gavanderhoorn:
Most posts I read about this mention that the System Groovy steps use a different runtime then other Groovy steps, and that the former doesn't have `ivy` on it's classpath. I just checked, but [templates/snippet/publisher_groovy-postbuild.xml](https://github.com/ros-infrastructure/ros_buildfarm/blob/c960ab17a7c5bd0b3a8176e1896d5764616ecdc2/ros_buildfarm/templates/snippet/publisher_groovy-postbuild.xml.em#L1) doesn't seem to be running the script in the System Groovy (I also have the exact same versions of the two plugins mentioned in that snippet here on the test deployment).

[this](https://groups.google.com/d/msg/job-dsl-plugin/EG6eqQYYI7M/2TKKysNw4QEJ) `job-dsl-plugin` google group posting is informative, but is from 2015 and states that `ivy.jar` needs to be added to the classpath manually (using the `Additional classpath` setting on the Groovy Postbuild step). It's also mentioned that the `job-dsl-plugin` brings its own ivy to work around precisely this problem.

Perhaps there are some differences in configuration from the 'default' `ros_buildfarm_deployment` deployment / `ros_buildfarm` job config and what `build.ros.org` is running? @nuclearsandwich?

dirk-thomas:
@gavanderhoorn Does your instance of Jenkins have the `job-dsl` plugin installed?

gavanderhoorn:
No. Already checked.

gavanderhoorn:
Hm. Only place I find any `ivy*.jar` on the build master is in `/var/lib/jenkins/plugins/workflow-cps-global-lib/WEB-INF/lib`. That is apparently a directory of the *Pipeline: Shared Groovy Libraries*. That plugin *is* installed on the master.

Does `build.ros.org` have that as well? I do not remember installing that manually, and I've not updated any plugins.

dirk-thomas:
No, the plugin `workflow-cps-global-lib` is currently not installed on `build.ros.org`. @nuclearsandwich Do you recall if that plugin was removed in the past?

gavanderhoorn:
ros-infrastructure/buildfarm_deployment#195 mentions `workflow-cps`.

Could all be a red herring of course.

nuclearsandwich:
That plugin was removed from our config in https://github.com/ros-infrastructure/buildfarm_deployment/pull/195 which is not to say that we removed it then. Just that it was not present on the buildfarm but was present in the puppet code previously.

We can re-install that plugin as a stopgap but we should figure out what we actually need and install that instead in the long term.

gavanderhoorn:
`ivy.jar` needs to be present on the classpath when the groovy post-build step is run.

I'm not sure the plugin adds it to the classpath, but that could be one way to do this.

Alternatives: add it to `WEB-INF/lib` of the main Jenkins dir (in the `war` folder IIRC), place it somewhere on the master and add it as an item to the `Additional classpath` setting under the Groovy postbuild step.

dirk-thomas:
I installed the Jenkins plugin `workflow-cps-global-lib` and the latest build passed and uploaded the artifact as expected: http://docs.ros.org/melodic/devel_jobs/catkin/

@nuclearsandwich Can you make sure to re-add `workflow-cps-global-lib` to the deployment logic. Thanks.

Over night all melodic devel jobs will be reconfigured (I manually reconfigured the `catkin` one for now).

gavanderhoorn:
Apologies for the disruption.

dirk-thomas:
No problem - all good!

dirk-thomas:
No problem - all good! Thanks for debugging so quickly.

gavanderhoorn:
Unrelated, but perhaps something to check: just noticed this in the catkin dev job output:

```
SSH: Failed to get hostname [ip-172-30-1-35: ip-172-30-1-35: Name or service not known]
```

gavanderhoorn:
Would there be some way to keep track these dependencies or make them explicit? The sort of runtime testing we just did doesn't seem very scalable and/or desirable.

nuclearsandwich:
> Apologies for the disruption.

No worries. It was a calculated risk to deploy first to production.

> @nuclearsandwich Can you make sure to re-add workflow-cps-global-lib to the deployment logic. Thanks.

I can, although I would like to know if we actually use that plugin for anything besides this dependency. There's a plugin which adds just ivy that might be preferable? https://plugins.jenkins.io/ivy

nuclearsandwich:
> Would there be some way to keep track these dependencies or make them explicit? The sort of runtime testing we just did doesn't seem very scalable and/or desirable

As far as I know there's not a way to track what plugins an individual jenkins job needs beyond spotting mentions of it in its build configuration. And in this case it seems like the missing classes are not part of the plugin but rather just a dependency of it.

The buildfarm_deployment repository is meant to encapsulate what we need in Jenkins to run a ROS buildfarm. This would have been caught if we'd tested this change on a testfarm provisioned with the current default config which we opted not to do this time.

gavanderhoorn:
@nuclearsandwich wrote:
> This would have been caught if we'd tested this change on a testfarm provisioned with the current default config which we opted not to do this time.

True.



> > @nuclearsandwich Can you make sure to re-add workflow-cps-global-lib to the deployment logic. Thanks.
> 
> I can, although I would like to know if we actually use that plugin for anything besides this dependency. There's a plugin which adds just ivy that might be preferable? https://plugins.jenkins.io/ivy

That would seem like a more minimalistic way to get this jar in there, yes.

ros_buildfarm/templates/devel/devel_job.xml.em
gavanderhoorn: Seems this does actually happen, see [Mdev__orocos_kinematics_dynamics__ubuntu_bionic_amd64](http://build.ros.org/job/Mdev__orocos_kinematics_dynamics__ubuntu_bionic_amd64/4/console) fi:

> No test result action for last build, skipping gathering statistics for it.

This seems to be caused by the `xUnit` plugin not finding any test reports:

```
[INFO] - [GoogleTest-1.6] - No test report file(s) were found with the pattern 'catkin_workspace/test_results/**/*.xml' relative to '/home/jenkins-agent/workspace/Mdev__orocos_kinematics_dynamics__ubuntu_bionic_amd64' for the testing framework 'GoogleTest-1.6'.  Did you enter a pattern relative to the correct directory?  Did you generate the result report(s) for 'GoogleTest-1.6'?
[WARNING] - No test reports found for the metric 'GoogleTest' with the resolved pattern 'catkin_workspace/test_results/**/*.xml'
[INFO] - Skipping the metric tool processing.
[INFO] - There are errors when processing test results.
[INFO] - Skipping tests recording.
```

PR #549, opened 2018-06-06
[Debian] don't overwrite default sources

mikaelarguedas:
We're facing apt failures on the farm for all debian platforms. Opening this for visibility yand testing. If this proves stable after some time we can consider merging it.

http://build.ros.org/job/Mbin_dsv8_dSv8__catkin__debian_stretch_arm64__binary/6/console

Context for why we overwrote the sources in the first place: https://github.com/ros-infrastructure/ros_buildfarm/issues/455

PR #551, opened 2018-06-07
Install apt transport https

mikaelarguedas:
**WIP do not merge** opening for visibility and testing.
Alternative to https://github.com/ros-infrastructure/ros_buildfarm/pull/549

### DIFF
--- a/ros_buildfarm/__init__.py
+++ b/ros_buildfarm/__init__.py
@@ -14,4 +14,4 @@
 
 # the version must be either a released version number which is also a tag name
 # or an upcoming version number followed by a branch name separated by a dash
-__version__ = '2.0.2-master'  # same version as in setup.py
+__version__ = '2.0.2-install-apt-transport-https'  # same version as in setup.py
--- a/ros_buildfarm/__init__.py
+++ b/ros_buildfarm/__init__.py
@@ -14,4 +14,4 @@
 
 # the version must be either a released version number which is also a tag name
 # or an upcoming version number followed by a branch name separated by a dash
-__version__ = '2.0.2-install-apt-transport-https'  # same version as in setup.py
+__version__ = '2.0.2-master'  # same version as in setup.py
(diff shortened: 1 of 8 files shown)


dirk-thomas:
I updated the comment but isn't there a good chance that the installation will already fail when we hit a https mirror?

mikaelarguedas:
> I updated the comment but isn't there a good chance that the installation will already fail when we hit a https mirror?

Yeah it should be done *before* the `sed` IMO.
We faced mirror issues (requiring the use of cloudfront) mostly for the additional components (nonfree contrib etc). So I would hope that the install before changing the mirror will work (but I dont have any number to back it ^^)

clalancette:
I've deployed the updated version of this PR to 20 of the failing packages on the buildfarm using this command:

```
TZ=PST+08 generate_release_jobs.py https://raw.githubusercontent.com/ros-infrastructure/ros_buildfarm_config/production/index.yaml melodic ds --package-names xacro actionlib_tutorials audio_play diagnostic_analysis ecl_linear_algebra executive_smach filters hardware_interface message_filters qb_chain qb_device_utils robot_activity rosdiagnostic rosmon roswtf rqt_bag rqt_graph rqt_logger_level rqt_py_common rqt_robot_steering
```

All 20 of those that I reconfigured are now passing.  As an additional piece of data, many of the downstream packages (which were not reconfigured) are still failing.  So this definitely seems to be helping so far.

clalancette:
All right, most of Melodic on Debian Stretch has rebuilt with this change, and seems much happier (Moveit is still building, but it will take a while).  Travis keeps failing; I haven't investigated it yet, though it seems unrelated.  Reviews are certainly welcome, then we can decide whether to merge with the failing CI or investigate it.

dirk-thomas:
To be considered for merging the change to the version needs to be reverted.

With that commit CI will get retriggered anyway. If it stil fails please look into it since it will only get merged with a green CI. For spurious error you can always retrigger the Travis builds if necessary.

mikaelarguedas:
Travis has been having connexion issues today. I experienced it on several repositories.

The last 5 travis run all failed at different times but all because of connection issue. Let's :crossed_fingers: for the new round build but if we don't get travis to pass I'd recommend merging before the nightlies anyway to get this deployed on all jobs.

Change looks good to me with the version reverted :+1:

PR #553, opened 2018-06-13
add dh-python explicitly as python3 doesnt ship with it since bionic

mikaelarguedas:
I decided to not add a layer to install it only on the platforms that don't have it so that this layer has a very similar effect on all platforms.

This is currently needed for the Bouncy source jobs to build

@nuclearsandwich FYI

### DIFF
--- a/ros_buildfarm/templates/release/binarydeb_create_task.Dockerfile.em
+++ b/ros_buildfarm/templates/release/binarydeb_create_task.Dockerfile.em
@@ -56,11 +56,6 @@ RUN echo "@today_str"
     os_code_name=os_code_name,
 ))@
 @(TEMPLATE(
-    'snippet/install_dh-python.Dockerfile.em',
-    os_name=os_name,
-    os_code_name=os_code_name,
-))@
-@(TEMPLATE(
     'snippet/install_apt-src.Dockerfile.em',
     os_name=os_name,
     os_code_name=os_code_name,
--- a/ros_buildfarm/templates/release/binarydeb_create_task.Dockerfile.em
+++ b/ros_buildfarm/templates/release/binarydeb_create_task.Dockerfile.em
@@ -56,6 +56,11 @@ RUN echo "@today_str"
     os_code_name=os_code_name,
 ))@
 @(TEMPLATE(
+    'snippet/install_dh-python.Dockerfile.em',
+    os_name=os_name,
+    os_code_name=os_code_name,
+))@
+@(TEMPLATE(
     'snippet/install_apt-src.Dockerfile.em',
     os_name=os_name,
     os_code_name=os_code_name,
--- a/ros_buildfarm/templates/release/binarydeb_task.Dockerfile.em
+++ b/ros_buildfarm/templates/release/binarydeb_task.Dockerfile.em
@@ -56,6 +56,12 @@ RUN echo "@today_str"
 ))@
 
 @(TEMPLATE(
+    'snippet/install_dh-python.Dockerfile.em',
+    os_name=os_name,
+    os_code_name=os_code_name,
+))@
+
+@(TEMPLATE(
     'snippet/install_ccache.Dockerfile.em',
     os_name=os_name,
     os_code_name=os_code_name,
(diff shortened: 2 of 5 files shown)


dirk-thomas:
Since this is only necessary for binary jobs it should only be installed in that case.

mikaelarguedas:
> Since this is only necessary for binary jobs it should only be installed in that case.

Yeah I raised the same question when preparing this. I'm opening this mostly for visibility and to use this branch for some first bouncy jobs. I agree we should narrow the scope before considering merging :+1:

nuclearsandwich:
> Since this is only necessary for binary jobs it should only be installed in that case.

This is needed in sourcedeb jobs as well since debhelper scripts run there. Which might be included in what you mean by binary jobs in which case this is for my own edification.

ros_buildfarm/templates/snippet/install_dh-python.Dockerfile.em
```diff
@@ -0,0 +1,4 @@
+@[if os_name == 'ubuntu' and os_code_name not in ['trusty', 'xenial', 'artful']]@
+@# Ubuntu Bionic doesn't ship dh-python with python3 anymore
+RUN for i in 1 2 3; do apt-get update && apt-get install -q -y dh-python && apt-get clean && break || if [[ $i < 3 ]]; then sleep 5; else false; fi; done
+@[end if]@
```
dirk-thomas: In all the used cases the apt wrapper script seems to be available. There it should be used here instead of the handcrafted loop.

ros_buildfarm/templates/release/binarydeb_task.Dockerfile.em
```diff
@@ -55,6 +55,12 @@ RUN echo "@today_str"
     os_code_name=os_code_name,
 ))@
 
+@(TEMPLATE(
+    'snippet/install_dh-python.Dockerfile.em',
+    os_name=os_name,
+    os_code_name=os_code_name,
+))@
+
```
dirk-thomas: Can you clarify if this is only necessary for sourcedeb job or also for binarydeb jobs?

ros_buildfarm/templates/release/binarydeb_create_task.Dockerfile.em
```diff
@@ -55,6 +55,11 @@ RUN echo "@today_str"
     os_name=os_name,
     os_code_name=os_code_name,
 ))@
+@(TEMPLATE(
+    'snippet/install_dh-python.Dockerfile.em',
+    os_name=os_name,
+    os_code_name=os_code_name,
+))@
```
dirk-thomas: Can you elaborate why the package is needed during the create task docker?

nuclearsandwich: I haven't tested it, but I don't think it's needed here as nothing during this task is actually invoking debhelper tasks.

ros_buildfarm/templates/release/binarydeb_task.Dockerfile.em
nuclearsandwich: It is needed for both sourcedeb and binarydeb jobs for any ament_python packages on bionic. As both the sourcedeb and binarydeb jobs invoke tasks from the debian/rules file. (`clean` is invoked in the sourcedeb jobs)

nuclearsandwich:
I've made some changes in response to feedback. I did some checking and it looks like debian buster also doesn't pull dh-python when installing python3. Is that something we should address now or put off until buster is out and in use?

dirk-thomas:
In the future please try to avoid having branches like this being used in production. For a short testing period that is fine (and sometimes necessary). But this should be the case for 9 days like in this case.

mikaelarguedas:
> In the future please try to avoid having branches like this being used in production

My bad, I was planning on revisiting this after the rest of the packaging :fire: was over.

nuclearsandwich:
I've merged as-is. I wanted to call back to the fact that this will affect debian buster as well. We don't have to do anything about it at the moment, but it might be easier to do now than remember to do it when we need it.

tfoote:
added buster in #556 so we don't get caught by it

nuclearsandwich:
> @tfoote tfoote deleted the install-dh-python branch 26 minutes ago

I should have specified in the thread, but the branch was being preserved as it was still deployed on build.ros2.org. Nothing is running at the moment and it's being reconfigured back to master now. So no harm done.

tfoote:
Ahh sorry. If there's any risk the "Restore branch" button is still available.

PR #554, opened 2018-06-21
Build environment variables from build files.

nuclearsandwich:
Allows release build files to specify environment variables which will be made available in the binarydeb build containers.

### DIFF
--- a/ros_buildfarm/argument.py
+++ b/ros_buildfarm/argument.py
@@ -63,6 +63,13 @@ def add_argument_build_name(parser, build_file_type):
              build_file_type)
 
 
+def add_argument_env_vars(parser):
+    parser.add_argument(
+        '--env-vars',
+        nargs='*',
+        help="Environment variables as 'key=value' for Dockerfile ENV directives")
+
+
 def add_argument_repository_name(parser):
     parser.add_argument(
         'repository_name',
--- a/ros_buildfarm/argument.py
+++ b/ros_buildfarm/argument.py
@@ -67,7 +67,8 @@ def add_argument_env_vars(parser):
     parser.add_argument(
         '--env-vars',
         nargs='*',
-        help="Environment variables as 'key=value' for Dockerfile ENV directives")
+        help="Environment variables as 'key=value' for Dockerfile "
+        'ENV directives')
 
 
 def add_argument_repository_name(parser):
--- a/ros_buildfarm/argument.py
+++ b/ros_buildfarm/argument.py
@@ -68,7 +68,7 @@ def add_argument_env_vars(parser):
         '--env-vars',
         nargs='*',
         help="Environment variables as 'key=value' for Dockerfile "
-        'ENV directives')
+             'ENV directives')
 
 
 def add_argument_repository_name(parser):
--- a/ros_buildfarm/config/release_build_file.py
+++ b/ros_buildfarm/config/release_build_file.py
@@ -43,6 +43,12 @@ class ReleaseBuildFile(BuildFile):
             self.abi_incompatibility_assumed = \
                 bool(data['abi_incompatibility_assumed'])
 
+        self.build_environment_variables = {}
+        if 'build_environment_variables' in data:
+            self.build_environment_variables = \
+                    data['build_environment_variables']
+            assert(isinstance(self.build_environment_variables, dict))
+
         self.jenkins_binary_job_label = None
         if 'jenkins_binary_job_label' in data:
             self.jenkins_binary_job_label = data['jenkins_binary_job_label']
--- a/ros_buildfarm/config/release_build_file.py
+++ b/ros_buildfarm/config/release_build_file.py
@@ -43,12 +43,6 @@ class ReleaseBuildFile(BuildFile):
             self.abi_incompatibility_assumed = \
                 bool(data['abi_incompatibility_assumed'])
 
-        self.build_environment_variables = None
-        if 'build_environment_variables' in data:
-            self.build_environment_variables = \
-                    data['build_environment_variables']
-            assert(isinstance(self.build_environment_variables, dict))
-
         self.jenkins_binary_job_label = None
         if 'jenkins_binary_job_label' in data:
             self.jenkins_binary_job_label = data['jenkins_binary_job_label']
--- a/ros_buildfarm/config/release_build_file.py
+++ b/ros_buildfarm/config/release_build_file.py
@@ -43,7 +43,7 @@ class ReleaseBuildFile(BuildFile):
             self.abi_incompatibility_assumed = \
                 bool(data['abi_incompatibility_assumed'])
 
-        self.build_environment_variables = {}
+        self.build_environment_variables = None
         if 'build_environment_variables' in data:
             self.build_environment_variables = \
                     data['build_environment_variables']
--- a/ros_buildfarm/release_job.py
+++ b/ros_buildfarm/release_job.py
@@ -699,6 +699,7 @@ def _get_binarydeb_job_config(
         'append_timestamp': build_file.abi_incompatibility_assumed,
 
         'binarydeb_files': binarydeb_files,
+        'build_environment_variables': ['%s=%s' % (var, value) for var, value in build_file.build_environment_variables.items()],
 
         'import_package_job_name': get_import_package_job_name(rosdistro_name),
         'debian_package_name': get_debian_package_name(
--- a/ros_buildfarm/release_job.py
+++ b/ros_buildfarm/release_job.py
@@ -663,6 +663,12 @@ def _get_binarydeb_job_config(
         'binarydeb/*.ddeb',
     ]
 
+    build_environment_variables = None
+    if build_file.build_environment_variables:
+        build_environment_variables = [
+            '%s=%s' % (var, value)
+            for var, value in build_file.build_environment_variables.items()]
+
     sync_to_testing_job_name = [get_sync_packages_to_testing_job_name(
         rosdistro_name, os_code_name, arch)]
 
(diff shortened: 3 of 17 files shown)


ros_buildfarm/argument.py
```diff
@@ -63,6 +63,13 @@ def add_argument_build_name(parser, build_file_type):
              build_file_type)
 
 
+def add_argument_env_vars(parser):
+    parser.add_argument(
+        '--env-vars',
+        nargs='*',
+        help="Environment variables as 'key=value' for Dockerfile ENV directives")
```
nuclearsandwich: This line exceeds the limit. I'm not sure how best to break up strings like this.

ros_buildfarm/config/release_build_file.py
```diff
@@ -43,6 +43,12 @@ def __init__(self, name, data):
             self.abi_incompatibility_assumed = \
                 bool(data['abi_incompatibility_assumed'])
 
+        self.build_environment_variables = {}
```
nuclearsandwich: Initializing this to an empty dict made sense initially, but it creates some similar but disparate constructs later as this is ultimately passed through via argparse nargs, which initializes omitted arguments to None rather than an empty list. So we end up with two lines

https://github.com/ros-infrastructure/ros_buildfarm/pull/554/files#diff-041d2fd48de98faa6a1db2d571bea72cR103 and https://github.com/ros-infrastructure/ros_buildfarm/pull/554/files#diff-b7ddb841a26f20ca5e4325409de758b3R107 which seem like they should be the same but are slightly different (one checks for presence / absence and the other for a non-zero length.

Initializing to None would allow us to keep that consistent but I'm not sure what is preferred.

ros_buildfarm/release_job.py
```diff
@@ -699,6 +699,7 @@ def _get_binarydeb_job_config(
         'append_timestamp': build_file.abi_incompatibility_assumed,
 
         'binarydeb_files': binarydeb_files,
+        'build_environment_variables': ['%s=%s' % (var, value) for var, value in build_file.build_environment_variables.items()],
```
nuclearsandwich: Again I'm not sure what the preferred style for breaking up this line would be.

nuclearsandwich:
@dirk-thomas I put up some comments calling attention to things I know I'd like feedback on in addition to any other review points.

nuclearsandwich:
When we deploy this on the ROS 2 buildfarm we'll want to either have merged #553 or create an omnibus branch to deploy that includes both PRs.

ros_buildfarm/argument.py
dirk-thomas: The same file contains this kind of case (e.g. https://github.com/ros-infrastructure/ros_buildfarm/blob/728cfe8e3c40ecd4e0128030d1d66864816626c8/ros_buildfarm/argument.py#L167-L168). So just use the same style here.

ros_buildfarm/config/release_build_file.py
dirk-thomas: The other options in this file initialize with `None` and only set the value to something else when the config file contains the specific key. So I would use the same here.

Both conditions can be written in the same way since an empty dictionary as well as `None` are consider `False` when interpreted as a `bool`.

ros_buildfarm/release_job.py
dirk-thomas: Commonly we wrap after the open parenthesis / bracket and then where necessary / there is a logical separation, e.g.:

```
        'build_environment_variables': [
            '%s=%s' % (var, value)
            for var, value in build_file.build_environment_variables.items()],
```

ros_buildfarm/templates/release/binarydeb_task.Dockerfile.em
```diff
@@ -61,6 +61,13 @@ RUN echo "@today_str"
     os_code_name=os_code_name,
 ))@
 
+@[if build_environment_variables]@
+@(TEMPLATE(
+    'snippet/set_environment_variables.Dockerfile.em',
+    environment_variables=build_environment_variables,
+))@
+@[end if]@
```
dirk-thomas: Two avoid potentially two consecutive empty lines in the output the block should include either the leading empty line or the trailing empty line, e.g.:

```
))@
@[if build_environment_variables]@

@(TEMPLATE(
    'snippet/set_environment_variables.Dockerfile.em',
    environment_variables=build_environment_variables,
))@
@[end if]@

@(TEMPLATE(
```

scripts/release/run_binarydeb_job.py
```diff
@@ -69,6 +71,7 @@ def main(argv=sys.argv[1:]):
 
         'binarydeb_dir': '/tmp/binarydeb',
         'dockerfile_dir': '/tmp/docker_build_binarydeb',
+        'build_environment_variables': args.env_vars
```
dirk-thomas: Nitpick: move this line one up so that it is passed in the order it is being used in the template.

scripts/release/create_binarydeb_task_generator.py
```diff
@@ -100,6 +102,8 @@ def main(argv=sys.argv[1:]):
         'dependencies': debian_pkg_names,
         'dependency_versions': debian_pkg_versions,
 
+        'build_environment_variables': args.env_vars,
```
dirk-thomas: Nitpick: move this line three up so that it is passed in the order it is being used in the template.

nuclearsandwich:
Working on CI failures and nits now.

ros_buildfarm/templates/release/binarydeb_task.Dockerfile.em
nuclearsandwich: This is addressed in 41c9884. It's not automatically collapsed because the change was on a line above.

ros_buildfarm/argument.py
```diff
@@ -63,6 +63,14 @@ def add_argument_build_name(parser, build_file_type):
              build_file_type)
 
 
+def add_argument_env_vars(parser):
+    parser.add_argument(
+        '--env-vars',
+        nargs='*',
+        help="Environment variables as 'key=value' for Dockerfile "
+        'ENV directives')
```
dirk-thomas: This line will need to be indented so that the single quote matches the column of the double quote in the line above.

ros_buildfarm/release_job.py
```diff
@@ -663,6 +663,12 @@ def _get_binarydeb_job_config(
         'binarydeb/*.ddeb',
     ]
 
+    build_environment_variables = None
+    if build_file.build_environment_variables:
+        build_environment_variables = [
+            '%s=%s' % (var, value)
+            for var, value in build_file.build_environment_variables.items()]
```
dirk-thomas: If this code block is changed to make `build_environment_variables` always a list (potentially empty) the "downstream code can be simplified:
* `@[if build_environment_variables]@` / `@[end if]@` could be removed from the template

nuclearsandwich: That's a great point! I'll do that.

nuclearsandwich: On second thought.. I'm not sure that will work. How does an empty list get handled by argparse's nargs="*" behavior. If we invoke commands `--env-vars` or `--env-vars --other-argument`. Will that be an empty list rather than None?

dirk-thomas:
```
import argparse
p = argparse.ArgumentParser()
p.add_argument('--foo', nargs='*')
p.parse_args(['--foo'])
```

ros_buildfarm/config/release_build_file.py
```diff
@@ -43,6 +43,12 @@ def __init__(self, name, data):
             self.abi_incompatibility_assumed = \
                 bool(data['abi_incompatibility_assumed'])
 
+        self.build_environment_variables = None
+        if 'build_environment_variables' in data:
+            self.build_environment_variables = \
+                    data['build_environment_variables']
+            assert(isinstance(self.build_environment_variables, dict))
```
dirk-thomas: Please add the new option to the documentation.

nuclearsandwich: First pass at documentation in 045cf5e.

nuclearsandwich:
@dirk-thomas Thanks for the docs fixups.

I've just realized that if we want to be able to use the rti package in doc and/or devel jobs we should make the build_environment_variables option part of the superclass build file rather than just for release build files.

ros_buildfarm/templates/devel/devel_create_tasks.Dockerfile.em
```diff
@@ -73,7 +73,8 @@ cmd = \
     ' --os-code-name ' + os_code_name + \
     ' --arch ' + arch + \
     ' --distribution-repository-urls ' + ' '.join(distribution_repository_urls) + \
-    ' --distribution-repository-key-files ' + ' ' .join(['/tmp/keys/%d.key' % i for i in range(len(distribution_repository_keys))])
+    ' --distribution-repository-key-files ' + ' ' .join(['/tmp/keys/%d.key' % i for i in range(len(distribution_repository_keys))]) + \
+    ' --env-vars ' + ' ' .join(env_vars)
```
nuclearsandwich: This breaks from the form with the others because the argument dict is copied directly to the data passed to the template in https://github.com/ros-infrastructure/ros_buildfarm/blob/ba90d748bae49252bcd6edb97eed11951017ec3f/scripts/devel/run_devel_job.py#L60.

I'd be tempted to add it to the dict using the same name as the others for better grep-ability. If I did so, would it be preferable to also remove the env_vars key from the dict?

nuclearsandwich:
I've moved the option to the BuildFile class and added support for devel jobs (they're the only other job type I can easily test in ROS 2). Note that to test this with ROS 2 it needs to be used in tandem with the branch from #548

PR #555, opened 2018-06-22
Super status

DLu:
Awhile ago (2016), [I complained about the relative unreadability of the build farm status pages](http://lists.ros.org/lurker/message/20160115.033604.cac2d7e5.it.html) and I demoed the first version of [ROS Crawl](http://www.metrorobots.com/ros_crawl/).

Unfortunately, that page is out of date because it relied on the csv that the build farm used to generate. Since then, I added the ability to generate yaml (#521) and I've learned some more of the basics about how the build farm actually runs jobs. So here is my replacement for ROS Crawl, but generated from the buildfarm. 

The key module here is `super_status.py` which has the `build_super_status_page` function in it, which 
 * Reads all the [yaml files generated](http://repositories.ros.org/status_page/yaml/)
 * Merges them into one data structure and writes that yaml file to `multi_distro_status.yaml` (which I initially tried to do in #529)
 * Determines the "aggregate status" of each package (more on that below)
 * Merges all of the aggregate statuses for each package into a new dictionary, organized by Git Organization (i.e. `ros-planning`) (or similar structure for other hosts) and repository. 
 * Writes that data structure to `super_status.yaml`
 * Generates `super_status.html`

The aggregate status is my way to collect the build status across different CPU architectures (e.g. [ros_lunar_default](http://repositories.ros.org/status_page/ros_lunar_default.html), [ros_lunar_ds](http://repositories.ros.org/status_page/ros_lunar_ds.html)) and translate it to a per-ROS-distro English status. Current statuses include
 * released (i.e. all green)
 * waiting for new release (all green in build/test, nothing in main)
 * waiting for re-release (all green in build/test, old version in main)
 * source builds, binary doesn't
 * does not build on X (where X could be a OS, OS code name, CPU, or some combination)
 * More than 99% of the current build farm fall into one of those categories. *The rest are marked as complicated*
I even added some tests for this portion of the logic. 

You can see the (*old*) generated files here:
[super_status.tar.gz](https://github.com/ros-infrastructure/ros_buildfarm/files/2128667/super_status.tar.gz)

The (*new*) generated HTML looks like this:
screenshot from 2018-07-16 16-27-59
*You can expand each row to get the English status.* ~~Each organization/repo is expandable/collapsible.~~
Caveats:
 * ~~The dynamic filtering and URL parameters that the other status pages have are not yet implemented.~~
 * ~~The super_status.yaml contains information about the maintainers, but I haven't added to the HTML yet.~~ 
 * I haven't tried all the job generating stuff yet on a real build farm. 
 * The folder where to download the yaml from is hard coded, as I haven't found a place where that is saved in the buildfarm configuration.

### DIFF
--- a/ros_buildfarm/aggregate_status.py
+++ b/ros_buildfarm/aggregate_status.py
@@ -1,0 +1,182 @@
+import re
+import collections
+
+CANDIDATES = ['build', 'test', 'main']
+VERSION_PATTERN = re.compile('([\d\-\.]+)\w+.*')
+
+
+def map_value_matches(M, key, value_list):
+    """
+       Convenience function for repeated operation to check if the
+       dictionary M's value for the given key (which is a set)
+       matches the set version of the list of values passed in.
+    """
+    return M[key] == set(value_list)
+
+
+def some_map_value_matches(M, value_list):
+    """
+       Returns true if some value in M
+       matches the set version of the list of values passed in.
+    """
+    values = set(value_list)
+    for key in M.keys():
+        if M[key] == values:
+            return True
+    return False
+
+
+def some_other_value_matches(M, value_list, exclude):
+    """
+       Returns true if some value in M except the one with key=exclude
+       matches the set version of the list of values passed in.
+    """
+    values = set(value_list)
+    for key in M.keys():
+        if key != exclude and M[key] == values:
+            return True
+    return False
+
+def no_overlap_in_values_and_none(M):
+    if None not in M:
+        return False
+    # Assume M has two key/value pairs
+    values0, values1 = M.values()
+    return len(values0.intersection(values1)) == 0
+
+def get_distro_status(D, expected, blacklist, candidates=CANDIDATES, skip_source=False, debug=False):
+    """
+     D is a recursive structure that map a specific build i.e.
+      (os_name, os_flavor, cpu, candidate) to a version number
+      If the version number is not present, then it is an error, i.e. version_number = None
+
+     expected is the values of (os_name, os_flavor, cpu) that we expect to see
+       (this is constant across all packages)
+
+     blacklist is a set of (os_name, os_flavor, cpu) that we don't expect to see
+       (specific to this package)
+    """
+
+    # The first thing we do is reverse the mapping
+
+    # version_map maps version to a list of (os_name, os_flavor, cpu, candidate)
+    version_map = collections.defaultdict(list)
+
+    # the following four structures map the version to either the os_name, os_flavor, etc
+    os_map = collections.defaultdict(set)
+    flavor_map = collections.defaultdict(set)
+    cpu_map = collections.defaultdict(set)
+    candidate_map = collections.defaultdict(set)
+
+    # this maps the version to the os_flavor + cpu
+    combo_map = collections.defaultdict(set)
+
+    # build the reversed maps
+    for os_name in expected:
+        os_d = D.get(os_name, {})
+        for os_flavor in expected[os_name]:
+            fl_d = os_d.get(os_flavor, {})
+            for cpu in expected[os_name][os_flavor]:
+                if skip_source and cpu == 'source':
+                    continue
+                cpu_d = fl_d.get(cpu, {})
+                for candidate in candidates:
+                    version = None
+                    if candidate in cpu_d:
+                        version = VERSION_PATTERN.match(cpu_d[candidate]).group(1)
+                    elif (os_name, os_flavor, cpu) in blacklist:
+                        continue
+                    os_map[version].add(os_name)
+                    flavor_map[version].add(os_flavor)
+                    cpu_map[version].add(cpu)
+                    candidate_map[version].add(candidate)
+                    combo_map[version].add(os_flavor + '/' + cpu)
+                    version_map[version].append((os_name, os_flavor, cpu, candidate))
+
+    # If there's only one version across all builds (and nothing is missing), then
+    # this package is completely synced and released
+    if len(version_map) == 1:
+        return 'released'
+
+    # If there are two different versions available
+    if len(version_map) == 2:
+        if map_value_matches(candidate_map, None, ['main']):
+            # one version for build/test, and all the main builds are None
+            # this package hasn't been released yet, but it builds fine otherwise
+            return 'waiting for new release'
+        elif some_map_value_matches(candidate_map, ['main']):
+            # still one version for build/test, and some other version for main
+            return 'waiting for re-release'
+
+        if some_map_value_matches(cpu_map, ['source']):
+            cpu_version0, cpu_version1 = sorted(cpu_map.keys())
+            # these versions could be (None, version) or (old_version, new_version)
+
+            if cpu_version0 is None:
+                return 'source builds, binary doesn\'t'
+            elif map_value_matches(cpu_map, cpu_version0, ['source']):
+                # the source is the only thing that builds for the new version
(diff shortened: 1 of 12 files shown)


gavanderhoorn:
Can this also be ordered by package? If the view cannot be filtered (yet), rendering by package seems like it would be a nicer way to show this data. Users might not always know from which repository packages come (I don't).

gbiggs:
With the increasing use of CI (and its addition to the wiki), I wonder if it is possible to add that information to this page where available. Packages that don't have it could have a "?" symbol instead.

Maybe too much for this script? Perhaps another script that takes the results of this and CI result crawls and merges them?

DLu:
@gavanderhoorn Fair point.

@gbiggs Does the buildfarm have any awareness of the CI tools? I know you can find it on Github, but don't know where it is w.r.t. these scripts.

davetcoleman:
This looks super! Thanks :)

DLu:
Who is the best person to review this?

dirk-thomas:
Thank you for working on improving the status pages. Your version looks certainly much nicer than the existing ones. I will try to comment on the various aspects and apologize ahead of time that my comment is going to be pretty lengthy.

* The name "super status" doesn't allow any guess what the page is supposed to show. Considering the content it shows I think it is closest to the current compare page. 

* Assuming that readers will want to check if specific packages are up-to-date, the same across distros, or available across all targeted platforms the grouping by org unit is not really helpful. Maybe the reader doesn't even know where to look if only the package name is known. As mentioned in the cons already implementing the search functionality would help a lot for this use case.

* In the case where the reader is a maintainer the lack of including the maintainer information as well as not being able to search for it makes the page much less useful for this group.

Considering these I am not sure if a new status page is the best step forward. Maybe updating the existing compare pages would achieve the same goals.

* The current grouping comes with several drawbacks mentioned. If a flat list would be used instead (as the compare page currently does) the orgunit could be added to the columns as well. That would allow to use client-side sorting of the table by columns as well as enable the search field to filter e.g. all repos in a specific org unit.

* While the icons provide a better overview sometimes the more detailed information about the exact version is really helpful. Maybe a "combined" compare page could offer both: show the icons by default and provide the version number on hover and have toggle to show the version number instead of icons (which allows e.g. to full text search on the page for a specific version).

A few smaller comments / mostly nitpicks:

* The hover text "waiting for new release" is not very clear imo. Maybe something more like "Waiting for a sync for the first release to become available in main" (doesn't have to be that verbose but I wanted to clarify what information it should try to include).

* Same for "waiting for re-release": "Waiting for a sync for the updated release to become available in main".

* The other status pages offer links to repositories, builds, etc. Those are pretty helpful to avoid that users have to find these URLs on their own.

* Not building on ARM has the same "error status" as a failure on amd64. I think it would be helpful to distinguish those for a better overview about severe and less severe problems.

* What does the status "complicated" mean?

* The alternate row colors don't work correctly in combination with folded lines.

* The page should mention that it focused around the status of the main apt repo (rather then building / testing). There should also be a legend what the various icons mean (similar as on the existing [package status pages](http://repositories.ros.org/status_page/melodic_default.html)).

DLu:
I've added a commit and edited the description/screenshot above. 
 * This PR restores the search/sorting functionality
 * Also adds maintainer information.
 * Adds quick links and searchbar
 * Adds legend
 * Removes the organization/repo nesting (for now)
 * Adds an expandable/collapsible info section with the "mouseover" English description.

Open questions:
 * I created a new page because I didn't want to destroy pages that other people were using. Do you want me to replace the existing compare pages?
 * If you want to keep generating this new page, do you have a different name suggestion?
 * What error text would you suggest for differentiating between `arm` and `amd64`? If it fails on arm (or similar) I'd say something like "fails on wacky processors" but that's probably not helpful.

There's a larger conversation about what the right English descriptions would be. It being centered on the main apt repo reflects how I've used the status pages in the past.

dirk-thomas:
> * This PR restores the search/sorting functionality
> * Also adds maintainer information.
> * Adds quick links and searchbar
> * Adds legend
> * Removes the organization/repo nesting (for now)
> * Adds an expandable/collapsible info section with the "mouseover" English description.

:+1:

> * I created a new page because I didn't want to destroy pages that other people were using. Do you want me to replace the existing compare pages?
> * If you want to keep generating this new page, do you have a different name suggestion?

(Correct me if I am wrong) the only information not available in this new page compared to the existing "compare" pages is the version number. If you could add those (maybe in the expandable section?) I think it would be great to merge the two pages again. So I think editing the existing compare page would be desired.

> * What error text would you suggest for differentiating between `arm` and `amd64`? If it fails on arm (or similar) I'd say something like "fails on wacky processors" but that's probably not helpful.

I don't have a good proposal for it. So not distinguishing them should be ok.

DLu:
I've added version numbers and org/repo links. 
screenshot from 2018-07-19 11-12-04c

In regards to replacing the compare pages, currently the buildfarm generates [one compare page for each distro, comparing it to all previous distros](https://github.com/ros-infrastructure/ros_buildfarm/blob/master/scripts/generate_all_jobs.py#L133) which has led to a [combinatoric number of compare pages](http://repositories.ros.org/status_page/) (some of which are stale). Do you think making just one "super_status" compare page would work instead of all the variations on the compare pages?

DLu:
@dirk-thomas What do you think? Replace all the previous compares with one page?

dirk-thomas:
> Do you think making just one "super_status" compare page would work instead of all the variations on the compare pages?

I don't think that is feasible. The advantage of having separate pages allows us to keep "frozen" states for distros which have been EOLed. With a single page that information would be lost once the distro is EOLed / removed. I would suggest to keep the combinatoric number of pages for now. In a second step we could consider to provide some configuration / heuristic to reduce the number of pages.

> Replace all the previous compares with one page?

I would suggest to modify the existing compare pages with the new features from this PR but keep the number of compare pages (see above).

DLu:
After starting to work on this, I realize that there's two separate tasks here.
 * One is for each distro (`indigo`), look at all the generated yaml across different architectures (`indigo-default`, `indigo-arm`) and generate a high level status.
 * The other is to generate the corresponding HTML

If I directly adapted the existing `release_compare` (combinatoric) pages, it would duplicate the work for the first task across all the pages. Would it make more sense to create another job for the first task, or just accept the duplicated work across pages?

dirk-thomas:
Currently all the status page jobs run in << than 10s. No problem to do the same work in them.

DLu:
@dirk-thomas I've got a number of revisions in the pipe for this. Would you like them all in one PR, or should I divide them up into more incremental pieces?

dirk-thomas:
Depends :wink:

Can you describe what else you are planning to do? Depending on what it is it might be good to do in this PR or better in a separate one.

DLu:
There's three-ish pieces. 
 1. The code to generate the aggregate status (independent of jenkins)
 2. A new job that automatically generates the aggregate status (mostly jenkins wrapper stuff)
 3. The changes to the compare page that incorporate the above aggregate status. 

I could make a PR of each, 1 and 2 combined followed by 3, or just do it in one fell swoop. Your choice.

DLu:
PR 1/3 is #567

dirk-thomas:
@DLu Friendly ping. Are you still interested to work on this or should this ticket be closed?

DLu:
This one is superseded by #567, so we can close this one.

ros-discourse:

https://discourse.ros.org/t/ros-rocks-multidistro-buildfarm-status/27836/1

PR #556, opened 2018-06-23
Add buster compatibility for dh-python

tfoote:
Following up #553 to avoid running into this when we spin up buster.


dirk-thomas:
Needs fixing.

tfoote:
Removed wrapping that ended up with incorrect indention.

PR #561, opened 2018-07-16
Allow searching by email on status pages

DLu:
[This link](http://repositories.ros.org/status_page/ros_lunar_default.html?q=dthomas) does not currently return any results. This PR enables searching by email address by adding all the email addresses to the hidden text field.

### DIFF
--- a/ros_buildfarm/templates/status/js/setup.js
+++ b/ros_buildfarm/templates/status/js/setup.js
@@ -216,15 +216,14 @@ function scan_rows() {
   // go in chunks, with timeouts in between.
   window.rows = [];
   $('table tbody tr').each(function() {
-    // Add lowercased version of name (which is the last meta column) for faster case-insensitive search.
     var name_td = $('td:nth-child(' + window.META_COLUMNS + ')', this);
     var hidden_text = '';
-    if (name_td.text() && name_td.text().length > 0) {
-      hidden_text += name_td.text().toLowerCase()
-    }
-    // Add a copy of the email to the hidden text for additional searching
+    // Add lowercased version of name (which is the last meta column) for faster case-insensitive search.
+    // Also, add a copy of the email to the hidden text for additional searching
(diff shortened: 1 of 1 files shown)


ros_buildfarm/templates/status/js/setup.js
```diff
@@ -218,8 +218,16 @@ function scan_rows() {
...
     // Add lowercased version of name (which is the last meta column) for faster case-insensitive search.
     var name_td = $('td:nth-child(' + window.META_COLUMNS + ')', this);
+    var hidden_text = '';
     if (name_td.text() && name_td.text().length > 0) {
-      name_td.append(' <span class="ht">' + name_td.text().toLowerCase() + '</span>');
+      hidden_text += name_td.text().toLowerCase()
+    }
+    // Add a copy of the email to the hidden text for additional searching
+    name_td.find('a').each(function() {
+      hidden_text += this['href'];
```
dirk-thomas: Should the `mailto:` prefix be stripped from this?

DLu: The text isn't visually seen. The existing code doesn't even insert spaces between names. However, if you're requesting that it strip mailto, I can do that.

dirk-thomas: > The text isn't visually seen.
> However, if you're requesting that it strip mailto, I can do that.

Sure, but then you could search for "mailto" which doesn't seem to be desired. Please remove the prefix.

> The existing code doesn't even insert spaces between names.

Maybe adding at least a space or even a separator would be good to avoid weird search results.

DLu:
I've updated this to have clearer hidden text with spaces, but the CI is failing now. I thiiiiiiink its a momentary hiccup. Can someone confirm that?

dirk-thomas:
> ... the CI is failing now

I restarted the build which failed due to a network issue. It passed this time.

Thank you for the improvement.

PR #564, opened 2018-07-30
update dpkg to fix installation problems with catkin-pkg

dirk-thomas:
This is necessary on Trusty to install `catkin_pkg` which is required by `ros-indigo-catkin`.


mikaelarguedas:
I guess this is needed because travis still runs on trusty machines?

Looks good to me as CI passes :+1:

dirk-thomas:
> I guess this is needed because travis still runs on trusty machines?

Yes.

PR #567, opened 2018-08-06
Aggregate Status

DLu:
First subcomponent of #555 

In short:
 * Merges all the [yamls](http://repositories.ros.org/status_page/yaml/) for a ROS distro (`ros_lunar_default`, `ros_lunar_ds`) and creates one large dictionary
 * Adds an English description for each package

Includes tests for the latter portion. Tests are not automatically called, and right now, none of the existing jobs call this code. 

CC: @dirk-thomas


DLu:
(I'm unclear why the CI is failing)

DLu:
CI is no longer failing. How's it look @dirk-thomas ?

DLu:
Any recommendations for easing the merging of this PR? Happy to do whatever I can to make this easier for y'all super busy Open Roboticists.

DLu:
2-Month checkup. @dirk-thomas Can you provide any sort of timeline on when you might be able to look at this?

ros_buildfarm/aggregate_status.py
```diff
@@ -0,0 +1,291 @@
...
+  all of the architectures. See examples below.
+
+  Naming Conventions
+   * distro refers to ROS Distro (e.g. indigo)
+   * machine refers to different release builds (e.g. default, uxhf, dsv8)
+   * os_name refers to the name of the Operating System (e.g. ubuntu, debian)
+   * os_flavor refers to specific distros of the os (e.g. xenial, jessie)
+   * cpu refers to the cpu architecture (or source), (e.g. i386, amd64, source)
+   * candidate refers to different candidate builds, (e.g. build, test, main)
+   * combo refers to the os_flavor + the cpu
```
dirk-thomas: These "naming conventions" sound confusing to me. To be consistent with the existing terms in use I suggest the following mapping:

* `distro` -> `rosdistro`
* `machine` -> `build_file_name`
* `os_flavor` -> `os_code_name`
* `cpu` -> `arch`
* `candidate` -> `repository_name`
* `combo` -> `target`

ros_buildfarm/aggregate_status.py
```diff
@@ -0,0 +1,291 @@
...
+
+  Example Descriptions:
+   - released - All candidates have the same version number
+   - waiting for new release - Build/test have the same version number, but there's nothing for main
+   - waiting for re-release - Build/test have the same version number, but there's an older version for main
+   - waiting for new/re-release - Build/test have the same version number, but there's either nothing or an older
+                                  version for main
+   - source builds, binary doesn't - If the source builds, but all the other cpus do not
+   - does not build on X - If the source and some of the cpus build, but some do not
```
dirk-thomas: The descriptions seem to be biased to a single perspective. From an end user point of view "waiting for new/re-release" might make sense. But from a package maintainer point of view the terminology is misleading. E.g. in the case of "waiting for new release" the release has already been made. The only step missing is a sync of the testing repo to the main repo to make the new package available.

There I would recommend a different description more towards what is actually missing / needs to happen.

---

Just an idea: for the generated yaml it might be good to just use an "enum" value to describe the state and then however the data is being "viewed" can decide on the precise string.

ros_buildfarm/aggregate_status.py
```diff
@@ -0,0 +1,291 @@
...
+"""
+
+CANDIDATES = ['build', 'test', 'main']  # Not imported from status_page to avoid potential circular dependency
+
+YAML_FOLDER = 'http://repositories.ros.org/status_page/yaml/'
```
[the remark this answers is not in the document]
DLu: I noted this in the [original PR](https://github.com/ros-infrastructure/ros_buildfarm/pull/555). I could not find anywhere in the [current production config](https://github.com/ros-infrastructure/ros_buildfarm_config/tree/production) that matched `'http://repositories.ros.org'` other than [some longer urls](https://github.com/ros-infrastructure/ros_buildfarm_config/blob/production/index.yaml#L85), but maybe I missed something. Where I can I get/derive this URL.

ros_buildfarm/aggregate_status.py
DLu: I found the naming of all these things confusing, so I appreciate you pointing me right. The only one I have trouble with is `candidate` -> `repository_name`. It doesn't seem like build, test and main are the names of (git?) repositories, so I was thinking of it more as build, test and main being different candidates for release. Is there another name besides repository name that suits you?

dirk-thomas: They are `apt` repositories.

DLu:
> How long does it take to run build_aggregate_status? 

On my machine, kinetic took 50 seconds and melodic took 22 seconds. That is primarily a function of loading the yaml files from the web (downloading and parsing). Kinetic's files are ~1M each and melodic is ~400K.  I honestly have no idea if that is relatively fast or slow compared to the other jobs.

ros_buildfarm/aggregate_status.py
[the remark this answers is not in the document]
dirk-thomas: The `prerequisites` is not the right information for this.

The generated files are being published via SSH and the only location where that target is specified atm is in the `buildfarm_deployment_config` afaik.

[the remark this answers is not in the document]
DLu: How would you recommend I tackle this then? Is there a way for me to grab them via ssh?

ros_buildfarm/aggregate_status.py
```diff
@@ -0,0 +1,262 @@
...
+                    version_map[version].append((os_name, os_code_name, arch, apt_repo))
+
+    # If there's only one version across all builds (and nothing is missing), then
+    # this package is completely synced and released
+    if len(version_map) == 1:
+        return 'released'
+
+    # If there are two different versions available
+    if len(version_map) == 2:
+        if apt_repo_map[None] == set(['main']):
```
dirk-thomas: A set can be defined with `{'main'}`.

Same a few times below.

test/test_aggregate_status.py
```diff
@@ -0,0 +1,20 @@
+import yaml
+from ros_buildfarm.aggregate_status import get_status_description
+
+def test_aggregate_status():
+    D = yaml.load(open('test/aggregate_status_tests.yaml'))
```
dirk-thomas: Please use lower case for variables.

test/test_aggregate_status.py
```diff
@@ -0,0 +1,20 @@
...
+    match = 0
+    for i, case in enumerate(D['test_cases']):
+        distro = case.get('distro', 'indigo')
+        blacklist = set(map(tuple, case.get('blacklist', [])))
+        status = get_status_description(case['build_status'], D['expected'][distro], blacklist)
+        assert status == case['result']
+        if status == case['result']:
```
dirk-thomas: These two lines don't seem to make much sense. The `else` case below can never be reached?

test/test_aggregate_status.py
```diff
@@ -0,0 +1,20 @@
...
+            match += 1
+        else:
+            print('Failure in test case #{}: {} != {}'.format(i + 1, status, case['result']))
```
dirk-thomas: (Will likely be removed) Error messages should be printed to `stderr`.

test/aggregate_status_tests.yaml
```diff
@@ -0,0 +1,143 @@
+test_cases:
```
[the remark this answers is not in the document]
dirk-thomas: This question is still pending.

ros_buildfarm/aggregate_status.py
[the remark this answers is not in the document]
dirk-thomas: Where ever these status pages are being configured there needs to be a config file (as for all the other job types) where this URL can be set. That part of the logic might not be in this PR as far as I can see since nothing uses the new Python module yet.

ros_buildfarm/aggregate_status.py
```diff
@@ -0,0 +1,262 @@
...
+Aggregate Status takes the yaml generated by the different CPU architectures (e.g. ros_lunar_default, ros_lunar_ds)
+and combines them into one yaml file per ROS distro.
+It also creates an English description of each package across all of the architectures.
+
+Example Descriptions:
+ - released - All apt_repos have the same version number
+ - waiting for new release - Build/test have the same version number, but there's nothing for main
+ - waiting for re-release - Build/test have the same version number, but there's an older version for main
+ - waiting for new/re-release - Build/test have the same version number, but there's either nothing or an older
+                                version for main
```
dirk-thomas: This line seems to be a combination of the two lines above?

ros_buildfarm/aggregate_status.py
```diff
@@ -0,0 +1,262 @@
...
+
+"""
+Aggregate Status takes the yaml generated by the different CPU architectures (e.g. ros_lunar_default, ros_lunar_ds)
+and combines them into one yaml file per ROS distro.
+It also creates an English description of each package across all of the architectures.
+
+Example Descriptions:
+ - released - All apt_repos have the same version number
+ - waiting for new release - Build/test have the same version number, but there's nothing for main
+ - waiting for re-release - Build/test have the same version number, but there's an older version for main
```
dirk-thomas: Similar to the original comment: https://github.com/ros-infrastructure/ros_buildfarm/pull/555#issuecomment-404328056

Both of the short labels "waiting for new release" / "waiting for re-release" are misleading to me. Under waiting for release I expect that someone has to do a release which is not the case here. Both are merely waiting for a sync to happen to make the latest release available in the `main` repo.

dirk-thomas:
A summary of an offline conversation @DLu and I had on how to proceed on this:

* Create a new status page (lets call it `compare2` for now). The goal is to replace the existing compare pages in the future.
  * I would suggest reusing as much as possible from [build_release_compare_page](https://github.com/ros-infrastructure/ros_buildfarm/blob/1d99c76f36a5ff2825c348e7de32dc2256dbdd53/ros_buildfarm/status_page.py#L874) and [build_release_status_page](https://github.com/ros-infrastructure/ros_buildfarm/blob/1d99c76f36a5ff2825c348e7de32dc2256dbdd53/ros_buildfarm/status_page.py#L41) - if necessary refactor some parts of these functions in order to reuse them rather than copy-n-paste / reimplement.
* Layout wise the new page will be similar to the existing `compare` page with the following changes:
  * Add a legend describing the colors with precise terminology.
  * While keeping the version numbers per ROS distro add colors to encode additional information (e.g. about the sync state, broken packages, etc.).
  * Have a tooltip describing the state in text form.
  * Make these cells links pointing to existing package release pages filtering by the package name of the row.

This should answer on a quick look if the same version is released across distros, the release is in a "good" state in each distro as well as when it's not show a tooltip with a summary and provide a link to the details somewhere else.

dirk-thomas:
@DLu Friendly ping. Are you still interested to work on this or should this ticket be closed?

DLu:
Closing for #811

PR #571, opened 2018-09-01
show who is maintainer

k-okada:
add row to show who is maintainer of the package
Currently we only have who is the maintainer of the blocked package, but I could not remember all package that I have to run `bloom-release`...

screenshot from 2018-09-01 15-43-50


dirk-thomas:
Closing due to no response.

PR #574, opened 2018-09-17
[prerelease] add ability to generate repos files for faster cloning

mikaelarguedas:
Opening this for discussion: not ready for review yet, just attaching code if it helps the discussion.

When running prerelease jobs, the cloning of the overlay can be pretty slow due to the fact that packages are cloned 1 by 1.

I would like to be able to generate a file (vcstool repos file or rosinstall file) so that users running the prerelease jobs can elect to invoke the ROS VCS tools manually for cloning the packages in the overlay.

Questions I'd like feedback on:
- What file format should be generated ? it looks like the rosinstall format would be supported by all tools and thus should be preferred over the vcstool format?
- Should the file always be generated (but not invoked) ? or should it be an option to the `generate_prerelease_script.py` script?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -74,6 +74,7 @@ matrix:
         - sudo apt-get update -qq
         - sudo apt-get install dpkg -y  # necessary for catkin-pkg to be installable
         - sudo apt-get install ros-indigo-catkin -y
+        - sudo apt-get install python3-vcstool -y
         - python setup.py install
         - mkdir job && cd job
         - ln -s .. ros_buildfarm
@@ -89,6 +90,10 @@ matrix:
         - docker
       env: JOB_TYPE=external_prerelease
       before_script:
+        - sudo sh -c 'echo "deb http://packages.ros.org/ros/ubuntu $(lsb_release -sc) main" > /etc/apt/sources.list.d/ros.list'
+        - sudo apt-key adv --keyserver hkp://ha.pool.sks-keyservers.net:80 --recv-key 421C365BD9FF1F717815A3895523BAEEB01FA116
+        - sudo apt-get update -qq
+        - sudo apt-get install python3-vcstool -y
         - python setup.py install
         - mkdir job && cd job
         - ln -s .. ros_buildfarm
@@ -173,6 +178,7 @@ matrix:
         - sudo apt-get update -qq
         - sudo apt-get install dpkg -y  # necessary for catkin-pkg to be installable
         - sudo apt-get install ros-indigo-catkin -y
+        - sudo apt-get install python3-vcstool -y
         - python setup.py install
         - mkdir job && cd job
         - ln -s .. ros_buildfarm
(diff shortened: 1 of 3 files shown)


dirk-thomas:
> What file format should be generated ? it looks like the rosinstall format would be supported by all tools and thus should be preferred over the vcstool format?

Either format is fine with me (as long as the logic ensures collisions in the rosinstall format can't happen).

I would suggest to use `vcstool` for fetch though since it is the lighter weight and more actibely maintained tool.

> Should the file always be generated (but not invoked) ? or should it be an option to the `generate_prerelease_script.py` script?

I would suggest to always use `vcstool` - I don't see a reason why you would want to op-out of it.

mikaelarguedas:
Updated to use vcstool in all cases and kept file generation as vcstool format for now.

`rosinstall_generator` currently doesn't leverage rosdistro's `RepositorySpecification` so there doesn't seem to be a straight forward way to generate rosinstall files with the current information available.

Note: if/when this gets merged the [Prerelease Instructions](http://wiki.ros.org/regression_tests#Prerelease_Instructions) should be updated to list `vcstool` as part of the prerequisites.

.travis.yml
```diff
@@ -89,6 +90,10 @@ matrix:
         - docker
       env: JOB_TYPE=external_prerelease
       before_script:
+        - sudo sh -c 'echo "deb http://packages.ros.org/ros/ubuntu $(lsb_release -sc) main" > /etc/apt/sources.list.d/ros.list'
+        - sudo apt-key adv --keyserver hkp://ha.pool.sks-keyservers.net:80 --recv-key 421C365BD9FF1F717815A3895523BAEEB01FA116
+        - sudo apt-get update -qq
+        - sudo apt-get install python3-vcstool -y
```
dirk-thomas: In the cases where the apt repo isn't necessary I would suggest to install the Python package via `pip` instead.

scripts/prerelease/generate_prerelease_overlay_script.py
```diff
@@ -57,9 +57,13 @@ def main(argv=sys.argv[1:]):
              "'catkin_workspace/src')"
     )
     parser.add_argument(
-        '--json', action='store_true',
+        '--json', action='store_true', default=False,
```
dirk-thomas: There is no need to add `default=False` here (or on the new option). The default default value is `None` which is sufficient.

scripts/prerelease/generate_prerelease_overlay_script.py
```diff
@@ -57,9 +57,13 @@ def main(argv=sys.argv[1:]):
...
         help='Output overlay information as JSON instead of a shell script'
     )
+    parser.add_argument(
+        '--vcstool', action='store_true', default=False,
+        help='Output overlay information as vcstool repos file'
+    )
```
dirk-thomas: The `json` and `--vcstool` options should be in a mutually exclusive group.

scripts/prerelease/generate_prerelease_overlay_script.py
```diff
@@ -98,14 +102,20 @@ def main(argv=sys.argv[1:]):
...
 
-    if not args.json:
+    if args.json:
+        print(json.dumps([vars(r) for r, p in scms], sort_keys=True, indent=2))
+    elif args.vcstool:
+        print('repositories:')
+        for r, p in scms:
+            print('  %s:\n    %s\n    %s\n    %s' % (
+                p, 'type: ' + r.type, 'url: ' + r.url, 'version: ' + r.version),
+            )
```
dirk-thomas: Three separate print calls would be easier to read.

mikaelarguedas:
This now requires https://github.com/dirk-thomas/vcstool/pull/77 to be able to pass CI

dirk-thomas:
Closing in favor of rebased patch in #600.

Issue #568, opened 2018-08-14
Ros2 rosindex generation and deploy


sloretz:
I'm looking for input about automating the generation and deployment of a ros2 rosindex equivalent website automatically. @mikaelarguedas gave me a an explanation of how this might look, and recommended opening a ticket for more comments.

Task background
----
rosindex is a static website using jekyll and a jekyll ruby plugin. Content is generated from [ros/rosdistro](https://github.com/ros/rosdistro). Once generated the result is pushed to a github repo to be hosted using github pages. The entire website is generated at once, meaning content for all ROS distros would be generated at the same time even if only one distro has changed. Generating bouncy + ardent takes less than a minute on my machine.

The website generation job would need to happen in a docker container that has ruby and some other dependencies installed.

Job Dependencies
----
The website should reflect released packages, so it would need to be regenerated any time a ROS distro is synced from testing to main. Since it generates content from all distros, something needs to be done to avoid generating content for other distros that have changed in `ros/rosdistro` but not been sync'd to main. Maybe a job could cache the ROS distro specific `distribution.yaml` file when the sync to main happens?

```
+-------------------+    +----------------------+
|sync to main ardent+--->+cache rosdistro ardent+-----+
+-------------------+    +----------------------+     |   +----------------------+
                                                      <-->+generate ros2 rosindex|
+-------------------+    +----------------------+     |   +----------------------+
|sync to main bouncy+--->+cache rosdistro bouncy+-----+
+-------------------+    +----------------------+
```

The output of `generate ros2 rosindex` is a commit to some github pages repository, so the job needs  a github user with write access. Are their other jobs that push to github which this job could be based off of?

gavanderhoorn:
With `rosindex`, are you referring to [rosindex](http://rosindex.github.io) by @jbohren?

dirk-thomas:
> I would recommend that we not try to couple this with the syncs.

:+1:

The `doc_independent-packages` job runs daily atm. Adding additional repos to this job can easily be done by configuration: see https://github.com/ros-infrastructure/ros_buildfarm_config/blob/195e00fec02727858afb75a4d99c1febd8a97365/doc-independent-build.yaml#L5-L12

While the job has a different scope at the moment it should be fairly easy to make it work for this use case:
[...]

PR #576, opened 2018-10-18
Support for a docker_build type of doc_independent build

hidmic:
Connected to #568. This pull request adds support for a new kind of package independent documentation builds, tailored to work with Jekyll/Middleman -generated static sites and relying on source side Docker containers to perform the actual build.


ros_buildfarm/config/doc_build_file.py
```diff
@@ -17,7 +17,9 @@
 DOC_TYPE_ROSDOC = 'rosdoc_lite'
 DOC_TYPE_MANIFEST = 'released_manifest'
 DOC_TYPE_MAKE = 'make_target'
-DOC_TYPES = [DOC_TYPE_ROSDOC, DOC_TYPE_MANIFEST, DOC_TYPE_MAKE]
+DOC_TYPE_EXTERNAL = 'external_site'
```
nuclearsandwich: I'm not sure if "external site" is best name for this doc type. The other types all describe the build mechanism not the built artifact. I would have expected something more like "docker_container" to similarly describe the how not the what.

Since we'll need to document the new documentation job type on https://github.com/ros-infrastructure/ros_buildfarm/blob/a9f0554840dda39272d76ff19f67c1675d77f8d6/doc/jobs/doc_jobs.rst#different-types-of-documentation that will probably be a good way to settle on a type name. It should match the description in documentation.

ros_buildfarm/doc_job.py
```diff
@@ -407,34 +408,49 @@ def configure_doc_independent_job(
...
-        'upload_root': build_file.upload_root,
 
         'notify_emails': build_file.notify_emails,
 
         'timeout_minutes': build_file.jenkins_job_timeout,
-
-        'credential_id': build_file.upload_credential_id,
     }
+
+    if build_file.documentation_type == 'make_target':
```
nuclearsandwich: I feel somewhat hesitant to tie the upload mechanism to the documentation type. It's possible that an external_site would still publish via ssh/rsync or that we may want to publish make_target docs via git.

This will need further discussion because I think ideally we'd preserve support for the existing upload configuration while deprecating it in favor of a nested structure that we can validate.

Something like below for the git publisher
```yaml
upload:
  type: git
  repository_url: git@github.com:ros-infrastructure/doc-target.git
  repository_branch: gh-pages
  credential_id: <credential_id>
```

and for the doc_independent job that uses make_target
```yaml
upload:
  type: rsync_ssh
  host: some-docs-server.ros.org
  root: /var/www/docs
  user: docsbot
  credential_id: <credential_id>
```

ros_buildfarm/templates/doc/doc_independent_site_job.xml.em
```diff
@@ -0,0 +1,135 @@
...
+))@
+@[end if]@
+@(SNIPPET(
+    'property_requeue-job',
+))@
+  </properties>
+@(SNIPPET(
+  'scm_git',
+   url=upload_repository_url,
+   branch_name=upload_repository_branch or 'gh-pages',
```
nuclearsandwich: Is there a reason to prefer setting a default in the template rather than setting it earlier on or not setting a default and asserting?

ros_buildfarm/templates/doc/doc_independent_site_job.xml.em
```diff
@@ -0,0 +1,135 @@
...
+))@
+  <assignedNode>@(node_label)</assignedNode>
+  <canRoam>false</canRoam>
+  <disabled>false</disabled>
+  <blockBuildWhenDownstreamBuilding>false</blockBuildWhenDownstreamBuilding>
+  <blockBuildWhenUpstreamBuilding>false</blockBuildWhenUpstreamBuilding>
+  <triggers>
+@(SNIPPET(
+    'trigger_timer',
+    spec='H 3 * * *',
```
nuclearsandwich: How long does the test job take and how often do we want it to run?

ros_buildfarm/config/doc_build_file.py
hidmic: I don't have a strong preference, but the current job is somewhat coupled to the deploy mechanism as a git commit happens (and it is expected to happen) within the container.

ros_buildfarm/doc_job.py
hidmic: Makes perfect sense and I do like it better. However, if we go down that route, no repository management can take place inside the container. Not even a git commit.

In any case, I fear that improvement may delay this new functionality reaching production. What do you think?

ros_buildfarm/templates/doc/doc_independent_site_job.xml.em
hidmic: It is actually being set before. A leftover, I'll remove it.

ros_buildfarm/templates/doc/doc_independent_site_job.xml.em
hidmic: I presume it'll take somewhere between one and two hours on a good network connection. And I understand we want this to be built on a daily basis (just as it is for other independent doc types). We did not discuss any other form of triggering.

hidmic:
@nuclearsandwich Just pushed minimal changes to trigger the discussion. If we get (really) picky:

1. You're right about fully decoupling build and deploy stages. No need for `make` or a containerized version of it to know anything about the upload.
2. Using 'external_site' is then a poor choice.
3. Having both `make_target` and `external_site` types hooked up of the same job logic is weird. Those would look better decoupled.
4. If all of the above holds, then source side containers could even be an alternative to the base container provided for the other doc types instead of a separate type. 

I'm open for changes, but I'd rather give up flexibility to have this up and running.

ros_buildfarm/doc_job.py
nuclearsandwich: As much as I want to take the time to improve things while we're touching them, I think the change is either a pre-requisite to this PR or just an enhancement to follow up on and not something that should be included in this work.

ros_buildfarm/templates/doc/doc_independent_site_job.xml.em
nuclearsandwich: I had to remind myself of the Jenkins cron spec behavior. This runs once a day at an idle time during 3-4AM. :+1:

nuclearsandwich:
[rendered view of the updated doc](https://github.com/ros-infrastructure/ros_buildfarm/blob/33282d7de17e07ba03b24ee1e4172a296cdacf33/doc/jobs/doc_jobs.rst)

ros_buildfarm/templates/doc/doc_independent_site_job.xml.em
```diff
@@ -0,0 +1,135 @@
...
+        ' --build-arg user=`whoami` --build-arg uid=`id -u`' +
+        ' -t %s-env .' % doc_repository_name,
+        'echo "# END SECTION"',
+        '',
+        'echo "# BEGIN SECTION: Build & update site - %s"' % doc_repository_name,
+        'docker run' +
+        ' --rm' +
+        ' --net=host' +
+        ' --cidfile=$WORKSPACE/docker_site/docker.cid' +
+        ' -v $WORKSPACE/repository:/tmp/scaffold' +
```
nuclearsandwich: Is there a reason the name `scaffold` was chosen here? Was there a reason not to just use repository?

hidmic: I was just reusing Jekyll terminology. No strong opinion either, we can use `repository` or any other name you find more appropriate.

ros_buildfarm/templates/doc/doc_independent_site_job.xml.em
```diff
@@ -0,0 +1,135 @@
...
+        ' -v $WORKSPACE/site:/tmp/site' +
+        ' %s-env update_site /tmp/scaffold /tmp/site' % doc_repository_name,
```
nuclearsandwich: I'm thinking about the external_site "API" between the buildfarm scripts and the source-side containers. The API right now is the update_site command with the paths to the scaffold and site directories as positional arguments.

A more "Docker-ish" way to define the API might be to pass the mount points in as environment variables and then let the Dockerfile use that information however it wants by specifying a [CMD](https://docs.docker.com/engine/reference/builder/#cmd) directive in the source Dockerfile.

Doing so would increase the coupling between ros_buildfarm and docker but decrease the coupling between the source-side containers and ros_buildfarm so I think it's a small net gain.

ros_buildfarm/templates/doc/doc_independent_site_job.xml.em
nuclearsandwich: Thinking more broadly on this, could we use `upload_repository`to describe the repository that holds the published content and `doc_repository` for the documentation source repository everywhere? The more consistent we are with naming the easier it is to grep through the codebase looking for things and I try to propagate names from the config through to their use as much as is feasible.

doc/jobs/doc_jobs.rst
```diff
@@ -31,6 +31,12 @@ Each *doc build files* can have one of three specific documentation types:
   honor an environment variable identifying the destination of the generated
   documentation.
 
+* the **external_site** type also operates on a single independent repository.
+  For each *doc build file* with that type a Jenkins job is created. Such repository
```
dirk-thomas: Each sentence should start on a new line (see the [developer guideline](https://github.com/ros2/ros2/wiki/Developer-Guide#markdown)).

Same below.

ros_buildfarm/config/doc_build_file.py
```diff
@@ -17,7 +17,9 @@
...
+DOC_TYPES = [DOC_TYPE_ROSDOC, DOC_TYPE_MANIFEST,
```
dirk-thomas: To avoid aligning please wrap after the open bracket.

ros_buildfarm/config/doc_build_file.py
```diff
@@ -72,9 +74,16 @@ def __init__(self, name, data):
...
 
-        # doc_repositories can only be used with doc type make_target
+        # doc_repositories can only be used with make_target and external_site
+        # doc types
         is_make_target_type = self.documentation_type == DOC_TYPE_MAKE
-        assert not self.doc_repositories or is_make_target_type
+        is_external_site_type = self.documentation_type == DOC_TYPE_EXTERNAL
+        are_doc_repositories_allowed = (is_make_target_type or
+                                        is_external_site_type)
+        assert not self.doc_repositories or are_doc_repositories_allowed
```
dirk-thomas: With the logic getting more complex I would suggest to decouple this like:

```
if self.documentation_type not in (DOC_TYPE_MAKE, DOC_TYPE_EXTERNAL):
    assert not self.doc_repositories
```

ros_buildfarm/config/doc_build_file.py
```diff
@@ -72,9 +74,16 @@ def __init__(self, name, data):
...
+
+        # Single doc_repository allowed for external_site doc type
+        assert not is_external_site_type or len(self.doc_repositories) == 1
```
dirk-thomas: Why is this limitation necessary? Wouldn't it be easy & convenient if multiple repositories could be listed as for doc independent one?

ros_buildfarm/doc_job.py
```diff
@@ -399,6 +399,7 @@ def configure_doc_independent_job(
 
     job_config = _get_doc_independent_job_config(
         config, config_url, job_name, build_file)
+
```
dirk-thomas: Avoid unrelated whitespace changes.

ros_buildfarm/doc_job.py
```diff
@@ -407,34 +408,49 @@ def configure_doc_independent_job(
...
+        template_name = 'doc/doc_independent_site_job.xml.em'
+        job_data.update({
+            'doc_repository_url': build_file.doc_repositories[0],
+            'upload_repository_url': build_file.upload_repository_url,
+            'upload_repository_branch': build_file.upload_repository_branch,
+            'upload_credential_id': build_file.upload_credential_id,
+        })
+    else:
+        raise JobValidationError(
+            "Not independent documentation_type: " +
```
dirk-thomas: Use single quotes.

ros_buildfarm/config/doc_build_file.py
nuclearsandwich: It occurred to me to raise this point as well. It would be reasonable to expect that one job could clone N doc_repositories then build and run N containers, even if the N we need is one.

I didn't raise this question as the use of the jenkins git publisher means that all containers be required to transform the same source repository, but that isn't actually different from all the doc jobs being published to the same rsync destination.

So unless there's something else Dirk and I are missing I think lifting this limitation and iterating over all doc_repositories to build their containers is reasonable.

hidmic: @nuclearsandwich @dirk-thomas AFAIK, and correct me if I'm wrong, but jenkins git publisher will only allow us to handle one destination repo, that in the case of multiple repos would not be the (most common) case. I'd expect each to commit to the same gh-pages branch in different repositories.

(I'm addressing comments, will reply the rest today)

nuclearsandwich: I think you're right about the limitation, but the existing independent packages jobs all push to the same rsync target. A set of external sites could potentially all push to subdirectories of the same destination repository and that could be useful.

hidmic: Hmm, I had not considered that possibility. Alright, fair enough, will add that then.

ros_buildfarm/doc_job.py
hidmic: Done.

ros_buildfarm/doc_job.py
hidmic: My bad, fixed.

ros_buildfarm/config/doc_build_file.py
hidmic: Good point, done.

ros_buildfarm/config/doc_build_file.py
hidmic: Sure. Done.

doc/jobs/doc_jobs.rst
hidmic: Good catch. Fixed.

doc/jobs/doc_jobs.rst
```diff
@@ -31,6 +31,12 @@ Each *doc build files* can have one of three specific documentation types:
...
+  must provide its own Dockerfile below `docker/image/`. Within that container, an
+  **update_site** command taking the repository path and the site repository path
+  as arguments must be available.
```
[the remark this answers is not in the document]
hidmic: Documentation updated to match the new environment variable based logic.

ros_buildfarm/templates/doc/doc_independent_site_job.xml.em
hidmic: I was going to argue against but I just read the CMD directives are overridden by explicitly given arguments. Logic changed.

ros_buildfarm/templates/doc/doc_independent_site_job.xml.em
hidmic: That sounds perfectly reasonable. Names changed.

hidmic:
@nuclearsandwich @dirk-thomas let me know if there's anything else you'd like to change.

ros_buildfarm/doc_job.py
```diff
@@ -407,34 +407,48 @@ def configure_doc_independent_job(
...
+            build_file.documentation_type
+        )
+
+    repository_args, script_generating_key_files = \
+        get_repositories_and_script_generating_key_files(config=config)
+
+    job_data.update({
+        'script_generating_key_files': script_generating_key_files,
+        'repository_args': repository_args,
+    })
```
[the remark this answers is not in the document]
hidmic: I was trying to avoid duplication, but it's ok anyways, I can roll it back.

ros_buildfarm/doc_job.py
```diff
@@ -407,34 +407,48 @@ def configure_doc_independent_job(
...
     job_data = {
         'job_priority': build_file.jenkins_job_priority,
         'node_label': get_node_label(build_file.jenkins_job_label),
 
-        'ros_buildfarm_repository': get_repository(),
-
         'doc_repositories': build_file.doc_repositories,
-
-        'script_generating_key_files': script_generating_key_files,
+        'ros_buildfarm_repository': get_repository(),
```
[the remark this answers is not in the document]
hidmic: Same as above, though ordering is orthogonal.

hidmic:
@dirk-thomas I'm OK with any naming you'd like. I chose 'external_site' because there is some coupling between the build and deploy phases (i.e. the container commits) but we can call it 'docker_build' or something like that.

scripts/doc/generate_doc_independent_job.py
```diff
@@ -25,6 +25,7 @@
 from ros_buildfarm.config import get_global_doc_build_files
 from ros_buildfarm.config import get_index
 from ros_buildfarm.config.doc_build_file import DOC_TYPE_MAKE
+from ros_buildfarm.config.doc_build_file import DOC_TYPE_EXTERNAL
```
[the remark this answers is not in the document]
hidmic: Good point.

ros_buildfarm/doc_job.py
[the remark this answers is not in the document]
hidmic: `job_data` is populated differently based on the documentation type. I've rearranged things to look more like they used to though. PTAL.

doc/jobs/doc_jobs.rst
```diff
@@ -31,6 +31,14 @@ Each *doc build files* can have one of three specific documentation types:
   honor an environment variable identifying the destination of the generated
   documentation.
 
+* the **docker_build** type also operates on independent repositories.
+  For each *doc build file* with that type a Jenkins job is created.
+  Each repository must provide its own Dockerfile below `docker/image/`,
```
nuclearsandwich: It'd be great to add an example to the docs. I opted not to permalink it to keep the example up-to-date in theory.

```suggestion
  Each repository must provide its own Dockerfile below `docker/image/` (Example: [ros2/rosindex](https://github.com/ros2/rosindex/tree/ros2/docker/image)),
```

nuclearsandwich:
@dirk-thomas since this is an independent job type do you think we should deploy it from this branch and test before merging?

dirk-thomas:
> I posted one suggestion to add an example to the docs.

I would skip this since in the future this example might change which would likely become outdated in the docs without noticing. The current buildfarm config should be a "good enough" example once this feature is rolled out.

> since this is an independent job type do you think we should deploy it from this branch and test before merging?

Absolutely, good idea.

hidmic:
> The current buildfarm config should be a "good enough" example

True, though there's no example on how the Dockerfile should look like. Documentation is explicit about it, but IMHO an example is always helpful.

dirk-thomas:
Just for the record: http://build.ros.org/job/doc_rosindex/

hidmic:
Updated to better handle the `targets` stanza in doc-build files, as [per request](https://github.com/ros-infrastructure/ros_buildfarm_config/pull/123#discussion_r228202996). @dirk-thomas @nuclearsandwich PTAL.

dirk-thomas:
The documentation for the doc build file needs to be updated to match the logic.

hidmic:
Ready for another round.

doc/jobs/doc_jobs.rst
```diff
@@ -31,6 +31,15 @@ Each *doc build files* can have one of three specific documentation types:
   honor an environment variable identifying the destination of the generated
   documentation.
 
+* the **docker_build** type also operates on independent repositories.
+  For each *doc build file* with that type a Jenkins job is created.
+  Each repository must provide its own Dockerfile below `docker/image/`
+  (Example: [ros2/rosindex](https://github.com/ros2/rosindex/tree/ros2/docker/image)),
```
chapulina:
```suggestion
  (Example: [ros-infrastructure/rosindex](https://github.com/ros-infrastructure/rosindex/tree/ros2/docker/image)),
```

hidmic:
Thanks for the reviews @nuclearsandwich and @dirk-thomas !

dirk-thomas:
The `external-site-job-support` branch in this repo should probably be deleted and the deployed Jenkins job reconfigured to use the master branch.

hidmic:
Branch is on my fork, but I can drop it if you want. I don't have permission to redeploy the job though, @nuclearsandwich may I ask you to do so?

nuclearsandwich:
> Branch is on my fork, but I can drop it if you want. I don't have permission to redeploy the job though, @nuclearsandwich may I ask you to do so?

I should have clarified that this request was fulfilled.

In order to easily branch-deploy this I pushed a copy of your PR branch to the this fork (ros-infrastructure/ros_buildfarm) and deployed from there. I've since updated the config to use the master branch and deleted that copy of your PR branch. You're welcome to maintain any branches on your fork indefinitely or remove them at your leisure.

ros_buildfarm/templates/snippet/publisher_publish-over-git.xml.em
```diff
@@ -0,0 +1,12 @@
...
+  <pushMerge>false</pushMerge>
+  <pushOnlyIfSuccess>true</pushOnlyIfSuccess>
+  <forcePush>false</forcePush>
+  <branchesToPush>
+    <hudson.plugins.git.GitPublisher_-BranchToPush>
+      <targetRepoName>origin</targetRepoName>
+      <branchName>@branch</branchName>
+    </hudson.plugins.git.GitPublisher_-BranchToPush>
+  </branchesToPush>
+</hudson.plugins.git.GitPublisher>
```
dirk-thomas: The indentation is wrong resulting in a diff when regenerating the job. See #619 for the fix.

PR #577, opened 2018-10-20
remove catkin-specific naming from directories and variables

dirk-thomas:
With more and more changes coming in I think it makes sense to plan for integrating the `colcon` branch from #548 into the default branch with a configuration option to choose the build tool. While that might require some effort the longer the separation stays as is we will continue to have to port patches between the two main branches.

This patch is a precursor of that to minimize the changes:

* The first commit removes catkin-specific naming of directories as well as (environment) variables.
* The second commit passes the make flags via the environment instead of as a command line argument.
* The third commit moves the workspace functions into a module without `catkin` in the name.
* The fourth commit creates a `call_build_tool` function with a different API than `call_catkin_make_isolated` to allow switching the build tool by parameter.
* The fifth commit renames the `catkin_make_isolated_and_*` scripts.
* The sixth commit updates the code to use the renamed scripts and makes them use the new `call_build_tool` function.
* The seventh commit updates the `call_build_tool` function to support invoking `colcon`.
* The eighth commit adds the configuration option `build_tool` to source and doc build files and passes the information through the numerous layers.


dirk-thomas:
Closing in favor of #585.

PR #580, opened 2018-10-22
adding ccache for devel job execution, it's not needed in the dockerfile generation

### LINKED ISSUE #579: ccache not shared between docker builds?
Although, `ccache` is installed and enabled in all docker containers, the cache doesn't seem to be saved and restored between subsequent builds.
Only the stage _"Run Dockerfile - generating devel tasks"_ forwards the ccache folder from the host to the docker container via `-v /home/jenkins-agent/.ccache:/home/buildfarm/.ccache`, which is a prerequisite for sharing them. I don't know about the AWS setup, but on Travis we need to explicitly enable saving and restoring of cache folders.

tfoote:
Fixes #579


dirk-thomas:
Please squash merge changes (which don't have multiple commits which need to be preserved) in the future.

tfoote:
Sorry will do.

PR #582, opened 2018-10-23
consider CCACHE_DIR variable from environment

rhaschke:
In private pre-release scenarios, it's useful to consider the `CCACHE_DIR` variable from the environment.
In my particular use case, my home is on a slow network volume, and thus I redirect `CCACHE_DIR` to some local, fast drive. However, without this patch `pre-release.sh` yields the following error (because the home is also Kerberos-protected and even root cannot write there):
```
docker: Error response from daemon: error while creating mount source path '/homes/rhaschke/.ccache': mkdir /homes/rhaschke/.ccache: permission denied.
```

### DIFF
--- a/ros_buildfarm/templates/devel/devel_job.xml.em
+++ b/ros_buildfarm/templates/devel/devel_job.xml.em
@@ -191,7 +191,7 @@ if pull_request:
         ' -e=TRAVIS=$TRAVIS' +
         ' -v $WORKSPACE/ros_buildfarm:/tmp/ros_buildfarm:ro' +
         ' -v $WORKSPACE/catkin_workspace:/tmp/catkin_workspace' +
-        ' -v ~/.ccache:/home/buildfarm/.ccache' +
+        ' -v $CCACHE_DIR:/home/buildfarm/.ccache' +
         ' devel_build_and_install.%s_%s' % (rosdistro_name, source_repo_spec.name.lower()),
         'cd -',  # restore pwd when used in scripts
         'echo "# END SECTION"',
@@ -219,7 +219,7 @@ if pull_request:
         ' -e=TRAVIS=$TRAVIS' +
         ' -v $WORKSPACE/ros_buildfarm:/tmp/ros_buildfarm:ro' +
         ' -v $WORKSPACE/catkin_workspace:/tmp/catkin_workspace' +
-        ' -v ~/.ccache:/home/buildfarm/.ccache' +
+        ' -v $CCACHE_DIR:/home/buildfarm/.ccache' +
         ' devel_build_and_test.%s_%s' % (rosdistro_name, source_repo_spec.name.lower()),
         'cd -',  # restore pwd when used in scripts
         'echo "# END SECTION"',
--- a/ros_buildfarm/templates/devel/devel_job.xml.em
+++ b/ros_buildfarm/templates/devel/devel_job.xml.em
@@ -191,7 +191,7 @@ if pull_request:
         ' -e=TRAVIS=$TRAVIS' +
         ' -v $WORKSPACE/ros_buildfarm:/tmp/ros_buildfarm:ro' +
         ' -v $WORKSPACE/catkin_workspace:/tmp/catkin_workspace' +
-        ' -v $CCACHE_DIR:/home/buildfarm/.ccache' +
+        ' -v ~/.ccache:/home/buildfarm/.ccache' +
         ' devel_build_and_install.%s_%s' % (rosdistro_name, source_repo_spec.name.lower()),
         'cd -',  # restore pwd when used in scripts
         'echo "# END SECTION"',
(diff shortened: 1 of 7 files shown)


dirk-thomas:
How is the `CCACHE_DIR` variable being set in the non-local execution?

rhaschke:
> How is the CCACHE_DIR variable being set in the non-local execution?

Via `CCACHE_DIR=${CCACHE_DIR:-~/.ccache}`

like here: https://github.com/ros-infrastructure/ros_buildfarm/pull/582/files#diff-6f573f74c0f3d23bdf21f66f838d7301R12
However, I'm not sure, I found all relevant scripts to set this.

dirk-thomas:
The `*_script.sh.em` templates are the ones being used when you generate a script for local execution. On Jenkins those are not used so I would expect the variable to not be defined in a Jenkins job.

dirk-thomas:
I am confused. The default value of `CCACHE_DIR` is `$HOME/.ccache` - see https://ccache.samba.org/manual/latest.html#_configuration_settings. What does the updated patch accomplish? It looks like a no-op.

rhaschke:
The default value of `CCACHE_DIR` is not being set at all. In this case, the default value used by ccache is indeed `$HOME/.ccache`. However, when you want to use the variable `CCACHE_DIR` in the scripts, it needs to be defined - either to the existing value or to the default use by `ccache` itself.
Sorry, I thought that would be obvious.

dirk-thomas:
> However, when you want to use the variable `CCACHE_DIR` in the scripts, it needs to be defined - either to the existing value or to the default use by `ccache` itself.
> Sorry, I thought that would be obvious.

I am sorry for still not getting it. The patch only sets the environment variable but I don't see where it is being used (beside by `ccache` directly which will use the default if it isn't set explicitly).

ros_buildfarm/templates/devel/devel_job.xml.em
```diff
@@ -191,7 +191,7 @@ if pull_request:
         ' -e=TRAVIS=$TRAVIS' +
         ' -v $WORKSPACE/ros_buildfarm:/tmp/ros_buildfarm:ro' +
         ' -v $WORKSPACE/catkin_workspace:/tmp/catkin_workspace' +
-        ' -v ~/.ccache:/home/buildfarm/.ccache' +
+        ' -v $CCACHE_DIR:/home/buildfarm/.ccache' +
```
rhaschke: > ... but I don't see where it is being used ...

@dirk-thomas Originally, the variable `CCACHE_DIR` was used here to mount the user's ccache folder into the docker container. However, your comment in https://github.com/ros-infrastructure/ros_buildfarm/pull/582#issuecomment-432342615 made me revert these changes. 
But, when I grep for `~/.ccache:` only those *_job.xml.em files are listed. Hence, they should get this change!
```
ros_buildfarm/templates/release/binarydeb_job.xml.em:        ' -v ~/.ccache:/home/buildfarm/.ccache' + \
ros_buildfarm/templates/release/binarydeb_job.xml.em:        ' -v ~/.ccache:/home/buildfarm/.ccache' +
ros_buildfarm/templates/devel/devel_job.xml.em:        ' -v ~/.ccache:/home/buildfarm/.ccache' +
ros_buildfarm/templates/devel/devel_job.xml.em:        ' -v ~/.ccache:/home/buildfarm/.ccache' +
```

dirk-thomas: In the first version the environment variable was only set in local scripts but not for buildfarm jobs but unconditionally used to mount that directory. Hence my first comments: https://github.com/ros-infrastructure/ros_buildfarm/pull/582#issuecomment-432319236 and https://github.com/ros-infrastructure/ros_buildfarm/pull/582#issuecomment-432342615

Then you removed the usage during the mount which leaves just the environment variable without any effect: hence https://github.com/ros-infrastructure/ros_buildfarm/pull/582#issuecomment-432430274 and https://github.com/ros-infrastructure/ros_buildfarm/pull/582#issuecomment-433968179

The patch needs to be updated to set the environment variable (when run locally as well as on the buildfarm) and then it can be used in the mount parameter.

rhaschke: OK, I reverted the revert. If there is anything more missing, I think you are by far the better expert on this. I just noticed, that the `CCACHE_DIR` is not considered for mounting the appropriate folder into the docker container. I guess, all the bits and pieces are there and I'am sure you will quickly finalize this.

dirk-thomas: The patch is now back in the original state. While that works when you run the build locally on Jenkins the environment variable won't be defined yet and therefore the mount argument `-v $CCACHE_DIR:/home/buildfarm/.ccache` will expand to `-v :/home/buildfarm/.ccache` and therefore fail.

The environment variable needs to be not only defined in the local scripts but also in the case of running within a Jenkins job.

rhaschke: I agree. Please add the missing definition in the corresponding file. I don't know about the details of the build farm scripts to do this.

rhaschke:
Closing due to inactivity.

PR #587, opened 2018-11-01
Updates for working with newer Jenkins versions

nuclearsandwich:
Incorporating these changes will require updates to the buildfarm deployment and live Jenkins instance. As I document the process a guide will be available in ros-infrastructure/buildfarm_deployment#207 and the final version will get published to discourse.

Connects to ros-infrastructure/buildfarm_deployment#207

### DIFF
--- a/ros_buildfarm/templates/release/trigger_jobs.groovy.em
+++ b/ros_buildfarm/templates/release/trigger_jobs.groovy.em
@@ -21,7 +21,7 @@ add_job_names_@(int(i / group_size) + 1)(job_names)
 
 @(FILE('snippet/check_recursive_upstream_projects.groovy'))@
 
-println "Triggering " + job_names.size + " jobs..."
+println "Triggering " + job_names.size() + " jobs..."
 triggered = 0
 skipped = 0
 for (job_name in job_names) {
--- a/ros_buildfarm/templates/dashboard_view_all_jobs.xml.em
+++ b/ros_buildfarm/templates/dashboard_view_all_jobs.xml.em
@@ -15,7 +15,7 @@
     <hudson.views.LastSuccessColumn/>
     <hudson.views.LastFailureColumn/>
     <hudson.views.LastDurationColumn/>
-    <jenkins.plugins.extracolumns.BuildDescriptionColumn plugin="extra-columns@@1.18">
+    <jenkins.plugins.extracolumns.BuildDescriptionColumn plugin="extra-columns@@1.20">
       <columnWidth>80</columnWidth>
       <forceWidth>true</forceWidth>
     </jenkins.plugins.extracolumns.BuildDescriptionColumn>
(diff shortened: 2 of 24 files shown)


ros_buildfarm/templates/release/trigger_jobs.groovy.em
```diff
@@ -21,7 +21,7 @@ add_job_names_@(int(i / group_size) + 1)(job_names)
 
 @(FILE('snippet/check_recursive_upstream_projects.groovy'))@
 
-println "Triggering " + job_names.size + " jobs..."
+println "Triggering " + job_names.size() + " jobs..."
```
dirk-thomas: Why does Groovy code need parenthesis for functions without arguments now? If that is the case the same pattern is present in multiple Groovy snippets.

nuclearsandwich: > Why does Groovy code need parenthesis for functions without arguments now?

The Groovy sandbox used by the script-security plugin has a whitelist of methods and fields. When running scripts containing that line the sandbox was treating `size` as an invalid field access rather than a method and raising an error. I haven't looked upstream yet to determine if this is a bug or feature.

nuclearsandwich: The parentheses here work around a bug in the 1.44 script security plugin. It's referenced in https://issues.jenkins-ci.org/browse/JENKINS-50470 but I think it merits a separate issue as it is still outstanding. I'll open that issue but in the meantime I think it's worth adding the parentheses here. With a comment noting the upstream bug if desired.

It doesn't appear to be generally required to put parentheses on functions, just this specific case.

nuclearsandwich:
I've completed a manual sweep of xml template updates resulting from the updated jenkins and plugin versions. Barring the xml version question below this should be all the changes necessary to run on the latest Jenkins LTS with an updated set of plugins.

Because of the new publish-over plugin and the added configuration default in parameterized-trigger I am not sure these changes are backwards compatible with a buildfarm_deployment configuration based on master today.

Common among all updated configs and is the move to xml 1.1 with the updated XML pseudo tag `<?xml version='1.1' encoding='UTF-8'?>` but I've not found in the ros_buildfarm library where the xml version is set. @dirk-thomas if you have pointers that would be helpful otherwise I'll go exploring until I find it.

ros_buildfarm/templates/release/trigger_jobs.groovy.em
nuclearsandwich: Issue is open upstream https://issues.jenkins-ci.org/browse/JENKINS-54952 @dirk-thomas would you like a comment in source linking to this issue until it's resolved?

dirk-thomas: Not necessary. I was mostly curious why it only happens in this case.

nuclearsandwich:
> Common among all updated configs and is the move to xml 1.1 with the updated XML pseudo tag `<?xml version='1.1' encoding='UTF-8'?>` but I've not found in the ros_buildfarm library where the xml version is set.

The above comment is still outstanding although it doesn't prevent the buildfarm from operating and I think that can be addressed in a future PR if desired.

dirk-thomas:
I don't think this package every generates a XML declaration line. So I assume it either gets prepended in the Python library communicating with Jenkins or on the server side.

nuclearsandwich:
Preserving the development branch for a week to allow others to migrate to master (or a release including this PR).

PR #588, opened 2018-11-09
gracefully work for existing job configs with using code from master

dirk-thomas:
Fixes temporary regression of #585.

Existing job configuration neither pass `--build-tool` nor `--ros-version` to the `run_devel_job.py` or `run_doc_job.py` scripts. Though the updated scripts currently require these arguments. And aexample of a failing job due to this: http://build.ros.org/job/Mdev__common_msgs__ubuntu_bionic_amd64/11/

Once the jobs are being reconfigured the option `--build-tool` and `--ros-version` are always being passed so the problem goes away. In order to fix this in the meantime this patch relaxes the input requirements for these two scrips and falls back to the previous behavior. See succeeding build: http://build.ros.org/job/Mdev__common_msgs__ubuntu_bionic_amd64/13/

### DIFF
--- a/ros_buildfarm/argument.py
+++ b/ros_buildfarm/argument.py
@@ -292,7 +292,8 @@ def add_argument_build_tool(parser, required=False):
         help='The build tool to use' + default_help)
 
 
-def add_argument_ros_version(parser):
+def add_argument_ros_version(parser, required=False):
     parser.add_argument(
-        '--ros-version', type=int, required=True,
+        '--ros-version', type=int, required=required,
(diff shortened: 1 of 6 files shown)


dirk-thomas:
Never mind - there are other information hard coded in each job which will make it impossible to get existing job configs work with the current master. So the jobs need to be reconfigured anyway to make this work...

dirk-thomas:
The build after reconfiguring the job passed: http://build.ros.org/job/Mdev__common_msgs__ubuntu_bionic_amd64/14/

Issue #581, opened 2018-10-23
build-and-test stage uses install space from build-and-install stage


rhaschke:
In https://github.com/ros-planning/moveit/issues/1106 we observed a strange failure of the pre-release test.
The problem was on the one hand induced [by circular package dependencies](https://github.com/ros-planning/moveit/issues/1106#issuecomment-431831861).
However, there is another problem, which I first thought is [related to catkin's](https://github.com/ros-planning/moveit/issues/1106#issuecomment-432017181) function `list_insert_in_workspace_order()`.

However, digging deeper I found that the issue is actually related to the build farm. This also explains that I only observed the issue when building a full pre-release test, but not when building only the build-and-test stage.

Running **`build-and-install`**, catkin uses these variables:
```
CATKIN_DEVEL_PREFIX=/tmp/catkin_workspace/devel_isolated/moveit_moveit_ros_move_group
CATKIN_WORKSPACES=/tmp/catkin_workspace/install_isolated;/opt/ros/indigo
```
and finds libraries (and includes) at the correct location: 
* raw libs: `/tmp/catkin_workspace/install_isolated/lib;/opt/ros/indigo/lib`
* ordered libs: `/tmp/catkin_workspace/install_isolated/lib;/opt/ros/indigo/lib`

Subsequently running **`build-and-test`** yields:
```
CATKIN_DEVEL_PREFIX=/tmp/catkin_workspace/devel_isolated/moveit_moveit_ros_move_group
CATKIN_WORKSPACES=/tmp/catkin_workspace/devel_isolated/moveit_*;/opt/ros/indigo
```
However, includes and libs are still found in `install_isolated`, which shouldn't happen I think, and yields the wrong order of directories (because install_isolated is handled as "non-catkin" by `list_insert_in_workspace_order()`:
* raw libs: `/tmp/catkin_workspace/install_isolated/lib;/opt/ros/indigo/lib`
* ordered libs: `/opt/ros/indigo/lib;/tmp/catkin_workspace/install_isolated/lib`

Finally, running **`build-and-test` _alone_** yields the correct paths:
* raw libs: `/tmp/catkin_workspace/devel_isolated/*/lib;/opt/ros/indigo/lib;/tmp/catkin_workspace/devel_isolated/*/lib`
* ordered libs: `/tmp/catkin_workspace/devel_isolated/*/lib;/opt/ros/indigo/lib`

[...]

PR #591, opened 2018-11-13
wipe CMake caches before build-and-test

dirk-thomas:
Might address #581.


rhaschke:
Thanks for tackling this!

PR #617, opened 2019-04-08
add new dependency on xmlschema for rep repo

tfoote:
Added in https://github.com/ros-infrastructure/rep/pull/190

http://build.ros.org/job/doc_independent-packages/1207/console
```
14:35:57 python xsdValid.py
14:35:57 Traceback (most recent call last):
14:35:57   File "xsdValid.py", line 3, in <module>
14:35:57     import xmlschema
14:35:57 ImportError: No module named xmlschema
14:35:57 Makefile:26: recipe for target 'xsdvalid' failed
14:35:57 make[1]: *** [xsdvalid] Error 1
14:35:57 make[1]: Leaving directory '/tmp/repositories/rep'
14:35:57 Makefile:6: recipe for target 'html' failed
14:35:57 make: *** [html] Error 2
```

### DIFF
--- a/ros_buildfarm/templates/doc/doc_independent_task.Dockerfile.em
+++ b/ros_buildfarm/templates/doc/doc_independent_task.Dockerfile.em
@@ -38,7 +38,7 @@ RUN echo "@today_str"
 
 RUN python3 -u /tmp/wrapper_scripts/apt.py update-install-clean -q -y make python-catkin-pkg-modules python-dateutil python-pip python-wstool python-yaml
 # Pin sphinx at 1.8.5 re: https://github.com/ros-infrastructure/ros_buildfarm/issues/614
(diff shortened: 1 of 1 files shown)


dirk-thomas:
I am not convinced hard coding these dependencies in this repo is a good approach. E.g. nobody external could use this approach to add the `rep` repo to their buildfarm.

It would be better to configure the dependencies in the configuration file: [doc-independent-build.yaml](https://github.com/ros-infrastructure/ros_buildfarm_config/blob/0c664bb08000635c174ae679532f514d5d63e159/doc-independent-build.yaml)

tfoote:
I agree that would be better. I don't have time at the moment to do a refactor like that. I am just trying to get the build back to green. The other option is to roll back the work in 190 but I think this is better than that alternative.

dirk-thomas:
Superseded by #618.

PR #624, opened 2019-04-26
GPU support for build files

j-rivero:
The PR is the initial work to be able to run tests that require GPU only in the nodes where gpu support is present. Changes:
 * Add a basic method to detect gpu support
 * Add support for including ctest arguments in `call_build_tool`
 * Use ctest LT and LE parameters to filter gpu or tests.

To test the changes I used a custom branch in `gazebo_ros_pkgs`:
 1. `mkdir -p /tmp/lala/src && cd /tmp/lala/src`
 1. `git clone https://github.com/ros-simulation/gazebo_ros_pkg -b crystal_gpu_test`
 1.  `./build_and_test.py --build-tool colcon --workspace-root /tmp/lala/ --rosdistro-name crystal`

We would need to test the whole thing in a real jenkins enviroment to see how things go there.


ros_buildfarm/templates/devel/devel_create_tasks.Dockerfile.em
```diff
@@ -18,6 +18,11 @@ ENV DEBIAN_FRONTEND noninteractive
 
 RUN useradd -u @uid -m buildfarm
 
+# Todo: check to only include this in the case of gpu jobs
```
j-rivero: although this snippet can probably stay in any docker image (even those without GPU support) the elegant way would be to make it dependent of a required or present GPU support. I'm not sure how to do it in empy or if we need another template file.

ros_buildfarm/templates/devel/devel_job.xml.em
```diff
@@ -187,7 +187,7 @@ if pull_request:
         'echo "# END SECTION"',
         '',
         'echo "# BEGIN SECTION: Run Dockerfile - build and install"',
-        'docker run' +
+        'docker run --env=DISPLAY=:0.0 --env=QT_X11_NO_MITSHM=1 --volume=/tmp/.X11-unix:/tmp/.X11-unix:rw -env=XAUTHORITY= --volume=: --runtime=nvidia' +
```
j-rivero: This command needs to conditional to the gpu support. I'm again not sure what is the best why of doing it. @nuclearsandwich any pointers are really appreciated.

j-rivero:
I've added the support to use `nvidia-docker2` both in Dockerfiles and docker run calls. Although some tests are failing , I can not see any problem related to a wrong GPU support or rendering: http://54.174.3.45/job/Cdev_gpu__gazebo_ros_pkgs__ubuntu_bionic_amd64/15/consoleFull . I made a couple of comments that need your help @nuclearsandwich .

ros_buildfarm/templates/devel/devel_job.xml.em
nuclearsandwich: You can use `has_gpu_support()` to populate an option passed to the template generator and then check it with an empy conditional like

```
@[if use_nvidia_runtime ]@
docker run ...
@[else]@
docker run
@[end if]@
```

j-rivero: I think I found the place. I've modified the template to check for gpu `build_source_name` and use the special parameters in that call. 97fc01a

ros_buildfarm/templates/devel/devel_create_tasks.Dockerfile.em
j-rivero: I think I found the way 0f4ddc2 (bad variable name corrected in a second commit).

j-rivero: > with that in mind, does this Dockerfile need this setup at all or is it something that should be added to the task-performing dockerfiles rather than the task-creating one?

Given that the commands needs to be added in the dockerfile I understand that the task managing the creation of them is the one that should take care of this action.

ros_buildfarm/templates/devel/devel_job.xml.em
j-rivero: > You can use `has_gpu_support()` to populate an option passed to the template generator and then check it with an empy conditional like

Sorry I read your comment after committed the changes. I've used the `build_source_name` to do it, while it is easier than populate a new parameter it could result more fragile than your proposed solution. Whatever you prefer.

ros_buildfarm/templates/devel/devel_job.xml.em
```diff
@@ -43,6 +43,16 @@ if pull_request:
         },
     ]
 }@
+@[if source_build_name == "gpu"]@
+@{
+docker_run_cmd = 'docker run --env=DISPLAY=:0.0 --env=QT_X11_NO_MITSHM=1 --volume=/tmp/.X11-unix:/tmp/.X11-unix:rw -env=XAUTHORITY= --volume=: --runtime=nvidia'
```
nuclearsandwich: What is the purpose of `--volume=:`?

ros_buildfarm/templates/devel/devel_job.xml.em
nuclearsandwich: We should have a dedicated variable and check the build source file name is arbitrary and there may be more than one GPU config file. For example gpu and gpu_arm64.

j-rivero: Agree. I changed both templates checks: 56c22e5

ros_buildfarm/templates/devel/devel_job.xml.em
j-rivero: ops, I cleaned that up ce7ff31

ros_buildfarm/workspace.py
```diff
@@ -99,6 +100,13 @@ def call_build_tool(
     if cmake_args:
         cmd += ['--cmake-args'] + cmake_args
 
+    if ctest_args:
+        if build_tool == 'colcon':
+            cmd += ['--ctest-args'] + ctest_args
+        # No ctest parameters are passed in catkin_make_isolated
+        # because in ROS 1 the test target of the makefile is being
+        # not ctest
```
nuclearsandwich: Looks like you might have accidentally a whole word?

```suggestion
        # invoked not ctest
```

ros_buildfarm/common.py
```diff
@@ -533,3 +534,12 @@ def get_system_architecture():
     if machine == 'aarch64':
         return 'armv8'
     raise RuntimeError('Unable to determine architecture')
+
+
+def has_gpu_support():
+    # It detects only nvidia support. The implementation is to check if
+    # /dev/nvidia* links are present. Further work could be to use a
+    # more robust method (i.e: https://github.com/ntpeters/pyvidia)
+    if glob.glob('/dev/nvidia*'):
```
nuclearsandwich: If this check runs inside a non-nvidia runtime docker container (as I expect it would during the create_devel_task then the /dev/nvidia* devices aren't available).

j-rivero: That is true. I think that this situation may refer to the use of [devel_create_tasks.Dockerfile.em](https://github.com/ros-infrastructure/ros_buildfarm/pull/624/files#diff-d2dacd6d9707bab9ae565f1625d5ecf2R21). If I understand correctly the nvidia-docker support can be removed from there since it's not being used (not necessary). Is there any more places in these changes where this function is called as part of the `create_devel_task` process?

nuclearsandwich: >  If I understand correctly the nvidia-docker support can be removed from there since it's not being used (not necessary)

I think that optimally, the devel_create_tasks container _shouldn't_ need the nvidia runtime. But to accomplish that I think we need to run `has_gpu_support()` on from the host system and pass  an argument through a command-line argument through to the build and test containers.

j-rivero: That is true. I think that this situation may refer to the use of [devel_create_tasks.Dockerfile.em](https://github.com/ros-infrastructure/ros_buildfarm/pull/624/files#diff-d2dacd6d9707bab9ae565f1625d5ecf2R21). If I understand correctly the nvidia-docker support can be removed from there since it's not being used (not necessary). Is there any more places in these changes where this function is called as part of the `create_devel_task` process?

j-rivero:
I moved the check for gpu support from templates to bash in order to get it during runtime while running tests https://github.com/ros-infrastructure/ros_buildfarm/pull/624/commits/d0c7cad18169f968138869a5a5043c23289d10c4. With these changes I was able to test:

 * gpu execution of tests for `gazebo_ros_pkgs` is working as expected (some failures also present in my local tests): http://54.174.3.45/job/Cdev_gpu__gazebo_ros_pkgs__ubuntu_bionic_amd64/24/ 
 * normal execution of other software (`sros2` for testing) with the changes in this PR: http://54.174.3.45/job/Cdev__sros2__ubuntu_bionic_amd64/3/

dirk-thomas:
@j-rivero Friendly ping

ros_buildfarm/common.py
j-rivero: Following this up and reviewing my old changes:
 1. The [current implementation](https://github.com/ros-infrastructure/ros_buildfarm/pull/624/files#diff-5cf979d1d9687b7cad0d95e630cc87c6R190) delegates the check to run docker with appropriate arguments to the runtime execution of the `devel_job.xml` . It's not very elegant but it is probably the best moment to check for the gpu support.
 1. The PR still hosts the support to detect the gpu in the devel_create_tasks container. If 1. is good, I think we can remove it. If 1. is no good then I can add support to pass the value of `has_gpu_support` to scripts in the form of a parameter.

@nuclearsandwich what do you think?

j-rivero:
> @j-rivero Friendly ping

The failure in the CI was caused by temporary problem in the Travis infrastructure. Now green.

scripts/devel/build_and_test.py
```diff
@@ -73,6 +74,10 @@ def main(argv=sys.argv[1:]):
                 '-DBUILD_TESTING=1',
                 '-DCATKIN_ENABLE_TESTING=1', '-DCATKIN_SKIP_TESTING=0',
                 '-DCATKIN_TEST_RESULTS_DIR=%s' % test_results_dir]
+            if has_gpu_support():
+                ctest_args = ['-L "gpu_test"']
+            else:
+                ctest_args = ['-LE "gpu_test"']
```
dirk-thomas: First, I would think the arguments should be two items in the list (`['-L', 'gpu_test']`) - not one item including a space?

Second, I don't understand how the presence of GPU support is supposed to toggle tests with/without the `gpu_test` label. How would you choose the behavior from the outside (e.g. from a build file)? How would you run all tests - including the GPU tests?

dirk-thomas:
The previous CI was failing and this needs another rebase.

scripts/devel/build_and_test.py
j-rivero: > First, I would think the arguments should be two items in the list (`['-L', 'gpu_test']`) - not one item including a space?

Right, fixed here 7932b2f

j-rivero:
Following @dirk-thomas suggestion:

> Second, I don't understand how the presence of GPU support is supposed to toggle tests with/without the `gpu_test` label. How would you choose the behavior from the outside (e.g. from a build file)? How would you run all tests - including the GPU tests?

I've implemented the support for two parameters in the build files:
  * ``tests_require_gpu``: indicate if software tests needs gpu support to run correctly.
  * ``run_only_gpu_tests:`` indicate if only gpu tagged tests will be run.

With them we can cover the use cases of: require gpu support for tests and run all tests, require gpu support for tests and run only gpu tagged tests and run everything but tests tagged as gpu.

I've been running some tests with the [gpu tests annotation branch of gazebo_ros_pkgs](https://github.com/ros-simulation/gazebo_ros_pkgs/compare/eloquent...eloquent_gpu_test):
 * [Build Status](http://18.206.154.49/view/Edev_gpu/job/Edev_gpu__gazebo_ros_pkgs__ubuntu_bionic_amd64/17/) gazebo_ros_pkgs running only tagged gpu tests (using [this configuration](https://github.com/j-rivero/ros_buildfarm_config/compare/ros2...ros2_gpu_tests))
 * [Build Status](http://18.206.154.49/view/Edev_gpu/job/Edev_gpu__gazebo_ros_pkgs__ubuntu_bionic_amd64/15/) gazebo_ros_pkgs running all tests (using [this configuration](https://github.com/j-rivero/ros_buildfarm_config/compare/ros2...ros2_gpu_all_tests)) with respect to previous configuration).

If I'm not wrong, we are going to need a REP for these changes.

doc/configuration_options.rst
```diff
@@ -270,6 +270,13 @@ The following options are valid in version ``2`` (beside the generic options):
   * ``default``: a boolean flag as described for *test_commits*.
   * ``force``: a boolean flag as described for *test_commits*.
 
+* ``tests_require_gpu``: a dictionary to indicate if software tests needs gpu
+  support to run correctly.
+  * ``default``: a boolean flag as described for *test_commits*.
+  * ``run_only_gpu_tests:``` a dictionary flag to indicate if only gpu tagged
+  tests will be run.
+    * ``default``: a boolean flag as described for *test_commits*.
```
dirk-thomas: The indentation of the last two bullets seems off. The bullet containing `run_only_gpu_tests` needs an empty line before.

ros_buildfarm/argument.py
```diff
@@ -356,6 +356,15 @@ def add_argument_run_abichecker(parser):
         help='Run the ABI checker when compiling packages')
 
 
+def add_argument_require_gpu_support(parser):
+    parser.add_argument(
+        '--require-gpu-support', action='store_true',
+        help='GPU acceleration is needed in the build container')
+    parser.add_argument(
+        '--run-only-gpu-tests', action='store_true',
+        help='GPU acceleration is needed in the build container')
```
dirk-thomas: Wrong help text.

doc/configuration_options.rst
j-rivero: I've change it in d9a43bd

ros_buildfarm/argument.py
j-rivero: fixed in 60ac7c6

dirk-thomas:
Just to summarize today's offline 

* If `tests_require_gpu` is set but the machine doesn't have a GPU the job should fail.
* We remove `run_only_gpu_tests` and instead always run all tests within one package.

j-rivero:
> If `tests_require_gpu` is set but the machine doesn't have a GPU the job should fail.

I've [implemented the check](https://github.com/ros-infrastructure/ros_buildfarm/pull/624/files#diff-5cf979d1d9687b7cad0d95e630cc87c6R193) on `/dev/nvidia[0-9]` device and make the job to fail if not present. 
 * Tested: [Build Status](http://18.206.154.49/view/Edev_gpu/job/Edev_gpu__gazebo_ros_pkgs__ubuntu_bionic_amd64/32/)

> We remove `run_only_gpu_tests` and instead always run all tests within one package.
Changed in 440946f. Removed the only gpu option and all ctest_args support.

* Tested: [Build Status](http://18.206.154.49/view/Edev_gpu/job/Edev_gpu__gazebo_ros_pkgs__ubuntu_bionic_amd64/33/)

ros_buildfarm/templates/devel/devel_job.xml.em
```diff
@@ -219,7 +222,8 @@ if pull_request:
         'echo "# BEGIN SECTION: Run Dockerfile - build and test"',
         ''
         'if [ ! -d "$HOME/.ccache" ]; then mkdir $HOME/.ccache; fi',
-        'docker run' +
+        '',
+        ('docker run --env=DISPLAY=:0.0 --env=QT_X11_NO_MITSHM=1 --volume=/tmp/.X11-unix:/tmp/.X11-unix:rw  --gpus all' if require_gpu_support else 'docker run') +
```
dirk-thomas: Let's keep the common part - `docker run` - outside of the condition and only conditional inserr the gpu specific.

j-rivero: agree b0d0736

PR #627, opened 2019-05-02
prerelease: correct pre-seeded workspace name.

### LINKED ISSUE #626: Running prerelease locally with pre-seeded workspace builds 0 pkgs
Being a dumb user I essentially just copy-paste the last set of commands given in the [*Run for "custom" repositories*](https://github.com/ros-infrastructure/ros_buildfarm/blob/c63ad85b21470f3262086fcd987528a0efc0cf6d/doc/jobs/prerelease_jobs.rst#run-for-custom-repositories) section (so to run a prerelease on a pre-seeded workspace).

While nothing complains and the Docker images all get build, in the end it would appear no packages are found and nothing is build and/or tested:

```
No packages found in source space: /tmp/catkin_workspace/src
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
~~  traversing 0 packages in topological order:
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

and:

```
Build step 10

Test results of underlay workspace

Invoking: catkin_test_results /tmp/prerelease/catkin_workspace/test_results --all

Test results directory "/tmp/prerelease/catkin_workspace/test_results" does not exist

Skipping empty overlay workspace
```

This is the contents of `/tmp/prerelease` before running `prerelease.sh -y`:

```
ws/
└── src
    └── ros_buildfarm
        └── dummy_package
            ├── CMakeLists.txt
            └── package.xml

3 directories, 2 files
```
Which would seem to be the things that the documentation implies should be there.

Before diving into this: is this use of the prerelease scripts still supported?

gavanderhoorn:
As per subject.

Fix #626.

### DIFF
--- a/doc/jobs/prerelease_jobs.rst
+++ b/doc/jobs/prerelease_jobs.rst
@@ -150,7 +150,7 @@ The following commands are all it takes to run a prerelease job for a custom rep
 .. code:: sh
 
   mkdir /tmp/prerelease && cd /tmp/prerelease
-  git clone -b dummy_package https://github.com/ros-infrastructure/ros_buildfarm ws/src/ros_buildfarm
+  git clone -b dummy_package https://github.com/ros-infrastructure/ros_buildfarm catkin_workspace/src/ros_buildfarm
   generate_prerelease_script.py https://raw.githubusercontent.com/ros-infrastructure/ros_buildfarm_config/production/index.yaml kinetic default ubuntu xenial amd64 --output-dir .
   # the argument -y suppresses the question if you want to continue with content already present in the workspace
   ./prerelease.sh -y


dirk-thomas:
I don't think the current docs are wrong. The master branch doesn't use the directory name `catkin_workspace` anywhere anymore. Could it be that you are using the latest release but looking at the docs from master?

gavanderhoorn:
I'm using [this version](https://github.com/ros-infrastructure/ros_buildfarm/blob/c63ad85b21470f3262086fcd987528a0efc0cf6d/doc/jobs/prerelease_jobs.rst#run-for-custom-repositories) of the docs.

Apparently there hasn't been a releas since 2018-04-30; I wasn't aware of that.

Would it be an idea to do a release again? It's been a year and one of the first things the docs ask is to run `pip3 install ros_buildfarm`, which seems to use the released version. If that doesn't match the documentation that could have led to my confusion.

gavanderhoorn:
Closing as apparently `ws` is the correct name.

dirk-thomas:
> Would it be an idea to do a release again?

Absolutely, I am not sure when I will get to it but I hope within the next couple of weeks.

dirk-thomas:
> Would it be an idea to do a release again?

See #639 for the upcoming release.

PR #634, opened 2019-05-30
set ROS_PYTHON_VERSION when evaluating conditions

dirk-thomas:
Necessary to fix http://build.ros2.org/view/Ddev/job/Ddev__osrf_pycommon__ubuntu_bionic_amd64/4/ since it uses [conditional dependencies](https://github.com/osrf/osrf_pycommon/blob/8ba78744d9df7de70d824b4d76c11afe9e81ccd1/package.xml#L13-L14) based on the `ROS_PYTHON_VERSION`.

### DIFF
--- a/scripts/devel/create_devel_task_generator.py
+++ b/scripts/devel/create_devel_task_generator.py
@@ -73,7 +73,11 @@ def main(argv=sys.argv[1:]):
              'and instead of installing the tests are ran')
     args = parser.parse_args(argv)
 
-    condition_context = dict(os.environ)
+    condition_context = {}
+    for t in args.env_vars:
+        parts = t.split('=', 1)
+        assert len(parts) == 2, '--env-vars argument lacks equal sign: ' + t
+        condition_context[parts[0]] = parts[1]
     condition_context['ROS_DISTRO'] = args.rosdistro_name
     condition_context['ROS_VERSION'] = args.ros_version
 
--- a/scripts/devel/create_devel_task_generator.py
+++ b/scripts/devel/create_devel_task_generator.py
@@ -73,6 +73,10 @@ def main(argv=sys.argv[1:]):
              'and instead of installing the tests are ran')
     args = parser.parse_args(argv)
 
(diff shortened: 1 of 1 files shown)


nuclearsandwich:
Rather than hard-code this into ros_buildfarm source. Is it something we could add to build configuration files?

dirk-thomas:
> Is it something we could add to build configuration files?

Sure, that is possible too.

nuclearsandwich:
> Sure, that is possible too.

I think that approach would be more flexible as it would allow for Python 2 and Python 3 builds of packages with the same rosdistro name. It also allows the python version to be set for custom rosdistros without forking the ros_buildfarm library.

dirk-thomas:
I updated the patch to pass through `os.environ` and set `ROS_PYTHON_VERSION` in ros2/ros_buildfarm_config#26.

scripts/devel/create_devel_task_generator.py
```diff
@@ -73,17 +73,18 @@ def main(argv=sys.argv[1:]):
              'and instead of installing the tests are ran')
     args = parser.parse_args(argv)
 
+    condition_context = dict(os.environ)
```
nuclearsandwich: I think you actually want to use the build environment variables argument rather than the environment this script will be run in.

This is how the build arguments are passed through to the running container. (Introduced in #554)

```suggestion
    condition_context = args.env_vars
```

dirk-thomas: Good point. Fixed in 9c5e875ba72b3531b385428f6ae9c0b2a02f5c58.

`args.env_vars` is a list of strings so it needs to be converted into a `dict`.

dirk-thomas:
Before: [Build Status](http://build.ros2.org/job/Ddev__osrf_pycommon__ubuntu_bionic_amd64/4/)
After: [Build Status](http://build.ros2.org/job/Ddev__osrf_pycommon__ubuntu_bionic_amd64/6/)

PR #635, opened 2019-06-03
Embed repository GPG key.

nuclearsandwich:
The keyserver network is increasingly unreliable. Embedding the key
means changing it whenever the trusted key changes, but pulling a new
key from the keyserver also involves changing the key ID to be pulled so
this at least removes a moving part.

I'm not sure yet what Travis will do with this syntax so this is still a WIP.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -124,7 +124,7 @@ matrix:
             h+sLkzGf+my2IbsMCuc+3aeNMJ5Ej/vlXefCH/MpPWAHCqpQhe2DET/jRSaM53US
             AHNx8kw4MPUkxExgI7Sd
             =4Ofr
-            -----END PGP PUBLIC KEY BLOCK-----' | apt-key add -
+            -----END PGP PUBLIC KEY BLOCK-----' | sudo apt-key add -
         - sudo apt-get update -qq
         - sudo apt-get install dpkg -y  # necessary for catkin-pkg to be installable
         - sudo apt-get install ros-kinetic-catkin -y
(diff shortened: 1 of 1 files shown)


dirk-thomas:
If we still want to embed the key this PR needs to be updated since #641 switched to the new key.

nuclearsandwich:
Now that the key has been propagated across the keyserver network embedding the key is less urgent. Doing so protects against the key being poisoned or a keyserver being down when queried. Changing the key directly takes about as much effort as changing the fingerprint of the key to fetch from the keyserver so this approach doesn't introduce any new downsides. That being said I don't mind if we decide just to close this PR.

dirk-thomas:
> I don't mind

In the past it was very common that at least one of the many Travis CI builds failed and needed to be retriggered manually due to the keyserver failing to respond. Therefore I would expect this still to be valuable.

Atm the patch seems to only change one of the numerous tests which use the key. Please update the PR to change all of them.

dirk-thomas:
@nuclearsandwich Friendly ping.

dirk-thomas:
Is it possible to not duplicate the key N times in this file?

nuclearsandwich:
> Is it possible to not duplicate the key N times in this file?

I was just thinking that myself. I'm looking at ways to embed it in the .travis.yaml and reuse it as well as just sticking the key in .travis/ros.asc

Another option is fetching the key via https from a hosted location in ros/rosdistro which will be more reliable than the keyserver network but less reliable than just having it on hand.

tfoote:
I think I have a solution to the duplication in #669

dirk-thomas:
Is this ready to be merged?

nuclearsandwich:
With @tfoote's change the key line is replaced with a reference. If that's satisfactory then since tests are passing. I would say yes.

PR #637, opened 2019-06-06
Add all Ubuntu EOL distros back to boxturtle to old release template

mikaelarguedas:
Modify the `old_release_set.Dockerfile.em` template to behave as expected for all distros listed in the ROS snapshot repository.

Original motivation for this PR is to allow to reuse this template for the generation of the official docker images.

Related to https://github.com/osrf/docker_images/pull/273 and https://github.com/osrf/docker_templates/pull/65

Arguably this does not necessarily belong in the `ros_buildfarm` package as no buildfarm will be ran against these distros. If so a copy of this snippet can be hosted on https://github.com/osrf/docker_templates but would result in duplicating maintenance effort

### DIFF
--- a/ros_buildfarm/templates/snippet/old_release_set.Dockerfile.em
+++ b/ros_buildfarm/templates/snippet/old_release_set.Dockerfile.em
@@ -1,3 +1,5 @@
+@# Ubuntu raring and older are checked for official docker images generation
+@# https://github.com/osrf/docker_images/pull/273
 @[if os_name == 'ubuntu' and os_code_name in ['lucid', 'maverick', 'natty', 'oneiric', 'precise', 'quantal', 'raring', 'saucy', 'utopic', 'vivid', 'wily', 'yakkety', 'zesty', 'artful']]@
 RUN find /etc/apt/ -name *.list -exec sed -i -e 's/archive.ubuntu.com\|security.ubuntu.com/old-releases.ubuntu.com/g' {} \;
 @[end if]@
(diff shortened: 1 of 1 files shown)


dirk-thomas:
Can you elaborate why this is necessary / useful? Those old distributions are partially even using `rosbuild` instead of `catkin`. Is there any particular distribution you are trying to / actively run a buildfarm for?

mikaelarguedas:
>  Is there any particular distribution you are trying to / actively run a buildfarm for?

As per original description I am not and dont know of anyone who would try to run a buildfarm for these distributions: 

The only reason for submitting it here is to avoid duplicating and maintaining the same template in the `osrf/docker_templates` repository (docker_templates tries to reuse as much as possible the snippets maintained in ros_buildfarm).
The change itself is needed to be able to expand this template for all ubuntu distros having packages in the snapshot.ros.org repository.

As mentioned in original description, as this change is not really buildfarm related, I'm happy to consider another place to store such template if it is preferred

dirk-thomas:
If the plan is to build Docker images for all EOL distros and reuse this snippet that sounds fine with me.

Can you please add a comment above the condition mentioning the Docker images. Otherwise I would be worried this get truncated in the near future since the use is not obvious.

mikaelarguedas:
> Can you please add a comment above the condition mentioning the Docker images. Otherwise I would be worried this get truncated in the near future since the use is not obvious.

Comment added in https://github.com/ros-infrastructure/ros_buildfarm/pull/637/commits/a3a8ea3687250de568a2d4237551287775c7fb18

dirk-thomas:
Thanks.

PR #641, opened 2019-06-13
Update GPG key and move to supported platform for CI builds

tfoote:
This was failing due to the new GPG location. As Indigo is going EOL I've updated this to use the existing parameter $ROS_DISTRO_NAME instead of hard coding indigo.

Added gpg key change too. Cross reference: https://github.com/ros-infrastructure/roswiki/issues/276


dirk-thomas:
Retriggering CI...

PR #646, opened 2019-07-17
set git user email and name for 'git merge' to work

dirk-thomas:
Before: [Build Status](http://build.ros2.org/job/Eci__overlay_ubuntu_bionic_amd64/3/)
After: [Build Status](http://build.ros2.org/job/Eci__overlay_ubuntu_bionic_amd64/9/) (passing beyond the `git merge` call)

### DIFF
--- a/ros_buildfarm/templates/ci/create_workspace.Dockerfile.em
+++ b/ros_buildfarm/templates/ci/create_workspace.Dockerfile.em
@@ -69,6 +69,10 @@ RUN echo "@today_str"
     dependency_versions=dependency_versions,
 ))@
 
+# needed for 'vcs custom --git --args merge' invocation
+RUN python3 -u /tmp/wrapper_scripts/apt.py update-install-clean -q --no-install-recommends sudo
+RUN sudo -H -u buildfarm -- git config --global user.email "jenkins@@ros.invalid" && sudo -H -u buildfarm -- git config --global user.name "Jenkins ROS"
+
 @(TEMPLATE(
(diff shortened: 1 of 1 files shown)


ros_buildfarm/templates/ci/create_workspace.Dockerfile.em
```diff
@@ -69,6 +69,10 @@ RUN echo "@today_str"
     dependency_versions=dependency_versions,
 ))@
 
+# needed for 'vcs custom --git --args merge' invocation
+RUN python3 -u /tmp/wrapper_scripts/apt.py update-install-clean -q --no-install-recommends sudo
+RUN sudo -H -u buildfarm -- git config --global user.email "jenkins@@ros.invalid" && sudo -H -u buildfarm -- git config --global user.name "Jenkins ROS"
```
cottsay: Why is this done using `sudo` instead of just putting it after the `USER` entry on line 77?

dirk-thomas: I just did the same as in the ROS 2 CI repo: https://github.com/ros2/ci/blob/573f0dedc9cc2d6b14380ff67a2177fd3701ddbd/linux_docker_resources/Dockerfile#L177

dirk-thomas:
> Maybe we can consider changing that pattern in the future.

Please feel free to improve this in a follow up PR. I was only focused on getting the CI overlay builds to work in order to use them.

PR #656, opened 2019-07-26
Use --no-log-init to prevent Docker build from hanging

### LINKED ISSUE #654: Excessive CPU usage when building Dockerfiles
When building some Dockerfiles as part of the prerelease flow, we are getting stuck at the step:
```
Step 9/25 : RUN useradd -u 629291812 -m buildfarm
```
with excessively high CPU usage. This is apparently an issue with Docker (https://forums.docker.com/t/run-adduser-seems-to-hang-with-large-uid/27371/2)

As a workaround can you please modify your Dockerfile to do something like
```
useradd -u @uid -l -m buildfarm
```
instead?

mm318:
Fixes #654.

This implements the workaround found here: https://forums.docker.com/t/run-adduser-seems-to-hang-with-large-uid/27371/2

### DIFF
--- a/ros_buildfarm/templates/ci/ci_create_reconfigure_task.Dockerfile.em
+++ b/ros_buildfarm/templates/ci/ci_create_reconfigure_task.Dockerfile.em
@@ -11,7 +11,7 @@ ENV DEBIAN_FRONTEND noninteractive
     timezone=timezone,
 ))@
 
-RUN useradd -u @uid -m buildfarm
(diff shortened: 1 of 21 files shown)


dirk-thomas:
Thank you for the patch.

Please use the magic phrase "Fixes #xxx" in your description in the future. That will ensure the referenced ticket will be closed automatically when the PR is merged. (I updated your description this time.)

mm318:
Thanks!

PR #658, opened 2019-08-02
environment: update virtual env setup instructions.

### LINKED ISSUE #657: should the virtualenv setup instructions be updated?
The instructions in the [Virtualenv](https://github.com/ros-infrastructure/ros_buildfarm/blob/f1eaeb85e5536e03c0764895d6ed782701906d1f/doc/environment.rst#virtualenv) section of the *Setup environment to deploy configuration* use `pyvenv-3.4` with the `--no-pip` option to work around a bug.

I've always just used `virtualenv` itself or `python3 -m venv` and it appears that works for me.

The following steps result in a working venv with `ros_buildfarm` installed:

```bash
$ virtualenv -p python3 $HOME/venv_ros_buildfarm
$ source $HOME/venv_ros_buildfarm/bin/activate
$ pip3 install empy
$ pip3 install jenkinsapi
$ pip3 install rosdistro
$ pip3 install ros_buildfarm
```

Would it make sense to update the *Virtualenv* section with similar steps to avoid the mention of the `3.4` specific version (has confused people in the past) and the `pyvenv-3.4` use?

gavanderhoorn:
As per subject. Fixes #657.

The situation around `venv`, `pyvenv`, `pyenv`, `virtualenv` and friends appears to be a bit nebulous, but apparently the `venv` module is the way to go (at least according to [this](https://www.reddit.com/r/learnpython/comments/4hsudz/pyvenv_vs_virtualenv/) and [this](https://stackoverflow.com/questions/41573587) discussion).

It's part of the standard-library for Python 3 as well (but has to be installed separately on Ubuntu for reasons unknown) so should be available on all targeted systems.

I've removed the mention of Trusty as that is no longer a supported platform.

### DIFF
--- a/doc/environment.rst
+++ b/doc/environment.rst
@@ -19,21 +19,16 @@ The easiest way is to use a Python virtual environment.
 This avoids collisions between the required Python 3 packages and already
 installed Python 2 packages used by ROS.
 
-The following commands can be used on Ubuntu Trusty::
+The following commands can be used on Ubuntu Xenial and newer::
 
-  sudo apt update && sudo apt install python3 python3-all python3-pip
+  sudo apt update && sudo apt install python3 python3-all python3-pip python3-venv
 
   mkdir /tmp/deploy_ros_buildfarm
   cd /tmp/deploy_ros_buildfarm
 
-  # the option '--without-pip' is used to work around a Python bug:
-  # https://bugs.launchpad.net/ubuntu/+source/python3.4/+bug/1290847
-  pyvenv-3.4 --without-pip venv
+  python3 -m venv venv
 
   . venv/bin/activate
-
-  # necessary because of '--without-pip'
-  curl https://bootstrap.pypa.io/get-pip.py | python3
 
   pip3 install empy
   pip3 install jenkinsapi


gavanderhoorn:
I've left the separate commands for creation of the `/tmp/deploy_ros_buildfarm` even though it could be created as part of the `venv` invocation.

That leads to the somewhat awkward `python3 -m venv venv` command when creating the virtual environment.

dirk-thomas:
> That leads to the somewhat awkward `python3 -m venv venv` command when creating the virtual environment.

What part of this invocation are you considering to be awkward?

--

Also please use a phrase like "Fixes #657" in the original description in the future to automatically close the issue when the PR is merged. (I updated it in this case.)

gavanderhoorn:
> > That leads to the somewhat awkward `python3 -m venv venv` command when creating the virtual environment.
> 
> What part of this invocation are you considering to be awkward?

the fact that the `venv` arg is repeated (which looks like a copy-pasta if you're not paying attention).

> Also please use a phrase like "Fixes #657" in the original description in the future to automatically close the issue when the PR is merged. (I updated it in this case.)

The `fix` keyword was already part of the commit comment, which does the same thing. I did not believe there was a need to add it to the PR description as well.

dirk-thomas:
> The fix keyword was already part of the commit comment, which does the same thing.

:+1: it just isn't visible to the reviewer and doesn't create a connection between the issues. With the reference spelled out in a comment the issue now has a reference to the PR. Anyway all fine now.

dirk-thomas:
Thanks for the improvement.

PR #666, opened 2019-08-16
workarounds to get the Noetic CI jobs using Python 3 to turn over

dirk-thomas:
* The first commit ensures that `python3` is used to invoke `catkin_make_isolated` when `ROS_PYTHON_VERSION=3` is set in the environment (in case the shebang line still references `python`).
* The second commit rewrites the shebang lines of source files to use `python3` instead of `python`
  * only when `catkin_make_isolated` is being used and
  * only in the build-and-test step of CI jobs

And again: the third commit specifying the custom branch should be reverted before merging this!

### DIFF
--- a/ros_buildfarm/templates/ci/ci_job.xml.em
+++ b/ros_buildfarm/templates/ci/ci_job.xml.em
@@ -345,6 +345,11 @@ parameters = [
         ' "ccache -s"',
         'echo "# END SECTION"',
         '',
+    ] + ([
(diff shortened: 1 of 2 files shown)


nuclearsandwich:
This is an appropriate PR number for workarounds. :japanese_ogre:

dirk-thomas:
> This is an appropriate PR number for workarounds. :japanese_ogre:

Now I am a bit hesitant to merge this :worried:

> And again: the third commit specifying the custom branch should be reverted before merging this!

Removed from the PR.

Merging without squashing since we might revert the commits individually in the future...

PR #667, opened 2019-08-16
make order of build env vars deterministic

dirk-thomas:
Avoid unnecessary shuffling of the `build_environment_variables` on every reconfigure job (e.g. http://build.ros2.org/job/Dci_reconfigure-jobs/117/).

### DIFF
--- a/ros_buildfarm/ci_job.py
+++ b/ros_buildfarm/ci_job.py
@@ -249,7 +249,7 @@ def _get_ci_job_config(
     if build_file.build_environment_variables:
         build_environment_variables = [
             '%s=%s' % (var, value)
-            for var, value in build_file.build_environment_variables.items()]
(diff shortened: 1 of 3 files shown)


dirk-thomas:
> I would much rather see the order preserved

When the data is extracted from `yaml` it already is a `dict` which doesn't guarantee order (at least in earlier Python versions). So I don't think this goal is feasible.

nuclearsandwich:
> I would much rather see the order preserved

I think the "order" of environment variables would be a smell. Also since it seems we're using lexical order in all our build files this enforces that lexical order.

dirk-thomas:
And finally a deterministic reconfigure without the need to shuffle config every night a little bit: http://build.ros2.org/job/Dci_reconfigure-jobs/121/consoleFull

PR #673, opened 2019-08-23
Add new Ubuntu and Debian repositories.

nuclearsandwich:
Is it intentional that ros_buildfarm is not released to Debian?


dirk-thomas:
Simply because there was never a need to.

PR #674, opened 2019-08-23
Add scripts to build blocked source entries page

sloretz:
This also somewhat refactors the blocked releases page generation where convenient for code reuse.

This page shows which repositories need source entries, similar to the blocked releases page. The purpose is to determine which repos need to be worked on next for the Noetic release.

Fix Python 2.7 CI

### DIFF
--- a/ros_buildfarm/status_page.py
+++ b/ros_buildfarm/status_page.py
@@ -589,6 +589,32 @@ def build_blocked_releases_page(
     additional_resources(output_dir, copy_resources=copy_resources)
 
 
+def build_blocked_source_entries_page(
+        config_url, rosdistro_name,
+        output_dir, copy_resources=False
+):
+    start_time = time.localtime()
+
+    repos_info = _get_blocked_source_entries_info(config_url, rosdistro_name)
+
+    # Expand a template for the webpage
+    repos_data = [_format_repo_table_row(name, data) for name, data in sorted(repos_info.items())]
+    template_name = 'status/blocked_source_entries_page.html.em'
+    data = {
+        'title': 'ROS %s - blocked source entries' % rosdistro_name,
+        'start_time': time.strftime('%Y-%m-%d %H:%M:%S %z', start_time),
+        'resource_hashes': get_resource_hashes(),
+        'rosdistro_name': rosdistro_name.capitalize(),
+        'repos_data': repos_data,
+    }
+    output_filename = os.path.join(
+        output_dir, 'blocked_source_entries_%s.html' % rosdistro_name)
+    print("Generating blocked source entries page '%s':" % output_filename)
+    with open(output_filename, 'w') as h:
+        h.write(expand_template(template_name, data))
+    additional_resources(output_dir, copy_resources=copy_resources)
+
+
 def _div_wrap(data, name=None):
     if_attr = ' id="%s"' % name if name is not None else ''
     return '<div{0}>{1}</div>'.format(if_attr, data)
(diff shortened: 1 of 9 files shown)


ros_buildfarm/status_page.py
```diff
@@ -873,6 +925,96 @@ def _get_blocked_releases_info(config_url, rosdistro_name, repo_names=None):
...
+
+def _source_entry_repos(distro):
+    for repo_name in distro.repositories.keys():
+        if _has_source_entry(repo_name, distro):
+            yield repo_name
+
+
+def _package_dependencies(walker, package):
+    try:
+        yield from walker.get_recursive_depends(
```
dirk-thomas: This line result in a `SyntaxError` in the CI builds atm.

sloretz: Passes python 2 CI in 509ac76

ros_buildfarm/templates/status/blocked_source_entries_page_job.xml.em
```diff
@@ -0,0 +1,110 @@
...
+    'property_job-priority',
+    priority=20,
+))@
+@(SNIPPET(
+    'property_requeue-job',
+))@
+  </properties>
+@(SNIPPET(
+    'scm_git',
+    url='https://github.com/ros-infrastructure/ros_buildfarm',
```
dirk-thomas: Use variables as in the other templates for this.

Same for the `branch_name`.

sloretz: Using variables for url and branch name on both this page and the blocked_releases page in 9501fce

ros_buildfarm/status_page.py
```diff
@@ -39,6 +39,111 @@
 from .templates import get_template_path
 
 
+def build_blocked_source_entries_page(
```
dirk-thomas: Maybe move below the existing function `build_blocked_releases_page`?

ros_buildfarm/status_page.py
```diff
@@ -39,6 +39,111 @@
...
+        repos_info[repo_name]['released'] = False
+        repos_info[repo_name]['version'] = 'no'
+        repos_info[repo_name]['repos_blocked_by'] = {}
+        repos_info[repo_name]['maintainers'] = defaultdict(dict)
+
+    for repo_name in prev_released_repos:
+        repos_info[repo_name]['url'] = _repo_url(prev_dist, repo_name)
+
+    # has a source entry? Lets call that 'released' with a 'version' to reuse
+    # logic for the blocked releases page
```
dirk-thomas: The naming is really awkward. Can you point to what exact parts are being reused to get an idea if it is worth the weird naming?

ros_buildfarm/status_page.py
```diff
@@ -39,6 +39,111 @@
...
+        'resource_hashes': get_resource_hashes(),
+        'rosdistro_name': rosdistro_name.capitalize(),
+        'repos_data': repos_data,
+    }
+    output_filename = os.path.join(
+        output_dir, 'blocked_source_entries_%s.html' % rosdistro_name)
+    print("Generating blocked source entries page '%s':" % output_filename)
+    with open(output_filename, 'w') as h:
+        h.write(expand_template(template_name, data))
+    additional_resources(output_dir, copy_resources=copy_resources)
```
dirk-thomas: How about splitting this function in a similar way as the existing one for blocked releases: `_get_blocked_releases_info` focuses on collecting the data and `build_blocked_releases_page` which calls the former takes care of the template part and the formatting?

ros_buildfarm/status_page.py
```diff
@@ -873,6 +928,98 @@ def _get_blocked_releases_info(config_url, rosdistro_name, repo_names=None):
...
+    except ValueError:
+        raise ValueError('Distribution key not found in list of valid distributions.')
+
+    if i == 0:
+        raise ValueError('No previous distribution found.')
+
+    return valid_rosdistro_names[i - 1]
+
+
+def _rosdistro_or_latest(index, rosdistro_name=None):
```
dirk-thomas: I don't see where this function is ever used with a `rosdistro_name` being `None`?

ros_buildfarm/status_page.py
```diff
@@ -873,6 +928,98 @@ def _get_blocked_releases_info(config_url, rosdistro_name, repo_names=None):
     return repos_info
 
 
+def _has_source_entry(repo_name, dist_file):
```
dirk-thomas: Since this function is only used in a single place I would suggest to just place the condition in that location. (As an alternative at least move this function directly below the function it is used in.)

*Update:* I noticed that it follows the existing function `_is_released`. With the current patch that function is only called in a single place (`_released_repos`). Therefore I suggest the same for that function.

ros_buildfarm/status_page.py
```diff
@@ -873,6 +928,98 @@ def _get_blocked_releases_info(config_url, rosdistro_name, repo_names=None):
...
+    try:
+        for dep in walker.get_recursive_depends(
+                package, ['build', 'buildtool', 'run', 'test'], ros_packages_only=True,
+                limit_depth=1):
+            yield dep
+    except AssertionError as e:
+        print(e, file=sys.stderr)
+
+
+def _released_packages(distro, repo_names=None):
```
dirk-thomas: I don't see where this function is ever used with a `repo_names` being `None`?

ros_buildfarm/status_page.py
```diff
@@ -39,6 +39,111 @@
...
+            package_dependencies.update(_package_dependencies(prev_walker, pkg))
+
+        for dep in package_dependencies:
+            if dep in missing_packages:
+                blocking_repo = prev_dist.release_packages[dep].repository_name
+                if blocking_repo == repo_name:
+                    # ignore packages in the same repo
+                    continue
+                url = _repo_url(prev_dist, repo_name)
+                repos_info[repo_name]['repos_blocked_by'][blocking_repo] = url
```
dirk-thomas: I would suggest to avoid the single use variable `url` and directly assign the return value of the function call.

Same for `new_maintainers` below.

sloretz: Removed single use variables in 07d3052

ros_buildfarm/status_page.py
sloretz: Hmm I don't either. I copied it from here, but I'll remove that altogether. https://github.com/ros-infrastructure/ros_buildfarm/blob/467c34fb812281c32307c6b6339c2da13d394907/ros_buildfarm/status_page.py#L695-L698

sloretz: Removed `_rosdistro_or_latest()` in 5c2b1f3

ros_buildfarm/status_page.py
sloretz: Made `repo_names` required argument in 08a74fba9a428b7528cb67816e246e7c3bc15598

ros_buildfarm/status_page.py
sloretz: I moved `_has_source_entry()` and `_is_released()` below where they're used in 9109dc09da9f3821955c2f1a65203463729be16f and 2ec5472787fbe4da42879693346f7633daeea0f9. I chose that over inlining because the function name was helpful to me when trying to understand the logic.

ros_buildfarm/status_page.py
sloretz: I think it's just reusing [`_format_repo_table_row`](https://github.com/ros-infrastructure/ros_buildfarm/blob/467c34fb812281c32307c6b6339c2da13d394907/ros_buildfarm/status_page.py#L644)

dirk-thomas: Since `_format_repo_table_row` is fairly long I am fine with keeping it reusing that. Maybe update the comment to mention that function / line explicitly.

ros_buildfarm/status_page.py
sloretz: Moved in 5d615c3

ros_buildfarm/status_page.py
sloretz: +1 to smaller functions. Refactored in 334837e

ros_buildfarm/status_page.py
sloretz: Updated comment to mention `_format_repo_table_row` in 6667374

sloretz:
> It would be good to group the changes in the following commits so that they can be merged without squashing:

@dirk-thomas Changes re-grouped ( I made sure `git diff` with the old head showed no differences before force pushing this branch)

* 1b92242 - Refactor's blocked releases page for code reuse
* 4f11b47 - Makes Blocked releases page use variables for the buildfarm config URL and branch name
* 5ba2025 - Copies files from blocked releases page without modifying them
* a590fd6 - All the changes to add the blocked source entries page
* b59b1ee - Temporary branch name change. This will be dropped before merging.

ros_buildfarm/status_page.py
```diff
@@ -805,35 +792,21 @@ def _get_blocked_releases_info(config_url, rosdistro_name, repo_names=None):
...
-                            pkg = parse_package_string(pkg_xml)
-                        except InvalidPackage:
-                            pass
-                        else:
-                            pkg_maintainers = {m.name: m.email for m in pkg.maintainers}
-                            if unreleased_repo_name not in maintainers:
-                                maintainers[unreleased_repo_name] = {}
-                            maintainers[unreleased_repo_name].update(pkg_maintainers)
+                    new_maintainers = dict(_maintainers(prev_distribution, pkg_name))
+                    maintainers[unreleased_repo_name].update(new_maintainers)
```
dirk-thomas: Maybe eliminate the single use var `new_maintainers`?

sloretz: Removed single use variable and squashed into first commit fc9041c .

PR #677, opened 2019-09-02
Documentation improvements and a warning

max-krichenbauer:
Hello!

These are just some minor improvements to the documentation on how to set up a build farm, as well as adding a warning when users are about to make a common mistake in specifying a local file system path as config URL. 

The related ROS Answers thread is here:
https://answers.ros.org/question/331213/buildfarm-jobs-cant-access-ros_buildfarm-ros_buildfarm_config-repos/

### DIFF
--- a/doc/basic_configuration.rst
+++ b/doc/basic_configuration.rst
@@ -17,9 +17,16 @@ configuration to generate the necessary Jobs on the Jenkins master.
 Create / fork configuration repository
 --------------------------------------
 
-First you need to either fork the
+First you need to either fork or clone the
 `ros_buildfarm_config <https://github.com/ros-infrastructure/ros_buildfarm_config>`_
 repository or create a repository containing the same configuration files.
+
+**Important:**
+Since ``ros_buildfarm_config`` files need to be accessed from within docker images,
+you **can not** use a local file system path (``file:/``) to reference them.
+You have to provide an http server where the configuration files can be accessed.
+(Note that on the build farm master, Jenkins usually occupies the standard port 80).
+
 
 Then you must update the configuration files:
 
@@ -35,8 +42,7 @@ The following options must be set in *all* build files::
 
     notifications:
       committers: false
-    maintainers
-      committers: false
+      maintainers: false
 
 
 Update administrator notification
@@ -44,7 +50,7 @@ Update administrator notification
 
 Change the email address which gets notified about any administrative jobs.
 
-In the entry point yaml::
+In [ros-infrastructure/ros_buildfarm_config](https://github.com/ros-infrastructure/ros_buildfarm_config)'s `index.yaml`::
 
   notification_emails:
   - your_email@example.com
@@ -55,17 +61,33 @@ In *all* build files::
     emails:
     - your_email@example.com
 
+Note that you need to have a local smtp service configured for email notifications.
+
+If you do not require email notifications, remove the configuration line entries.
+
 
 Update URLs to point to custom build farm
 -----------------------------------------
 
-Change the Jenkins URL in the entry point yaml to point to your earlier
-provisioned Jenkins master::
+Change the ``rosdistro_index_url`` in the [ros-infrastructure/ros_buildfarm_config](https://github.com/ros-infrastructure/ros_buildfarm_config)'s 
+``index.yaml`` to point to the ``rosdistro`` repository that defines the distro(s) that your buildfarm should build.
+
+**Important:** Note that this the [rosdistro](https://github.com/ros/rosdistro)'s ``index.yaml``
+(which points to the files that define which packages are included in the distro and respective caches),
+and **not** the ``index.yaml`` in ``ros_buildfarm_config`` which you are editing.
+
+This will usually be on the provisioned ``repo`` host, but can be the official [ros/rosdistro](https://github.com/ros/rosdistro)'
+
+    rosdistro_index_url: http://repo_hostname.example.com/rosdistro/index.yaml
+
+
+Change the Jenkins URL in the [ros-infrastructure/ros_buildfarm_config](https://github.com/ros-infrastructure/ros_buildfarm_config)'s 
+``index.yaml`` to point to your earlier provisioned Jenkins master::
 
   jenkins_url: http://jenkins_hostname.example.com:8080
 
-Change the repository URLs in the entry point yaml to point to your earlier
-provisioned ``repo`` host::
+Change the repository URLs in the [ros-infrastructure/ros_buildfarm_config](https://github.com/ros-infrastructure/ros_buildfarm_config)'s
+``index.yaml`` to point to your earlier provisioned ``repo`` host::
 
   status_page_repositories:
     CUSTOM_NAME:
@@ -100,3 +122,49 @@ repository::
 
 You can extract the repository PGP key from the ``buildfarm_deployment``
 configuration.
+
+
+Update URLs to point to required build tool repositories
+--------------------------------------------------------
+
+During job execution, access to repositories which contain the necessary tools to run the ROS build farm is required.
+These can be main ROS repository or mirrors of it.
+Note that the number of ``debian_repositories`` and ``debian_repository_keys`` must match, as well as the order in which they're specified.
+The keys can usually be found on the repository (for example: http://packages.ros.org/ros.asc )
+
+    prerequisites:
+      debian_repositories:
+      - http://packages.ros.org/ros/ubuntu
+      debian_repository_keys:
+      - |
+        -----BEGIN PGP PUBLIC KEY BLOCK-----
+        Version: GnuPG v1
+
+        mQINBFzvJpYBEADY8l1YvO7iYW5gUESyzsTGnMvVUmlV3XarBaJz9bGRmgPXh7jc
+        VFrQhE0L/HV7LOfoLI9H2GWYyHBqN5ERBlcA8XxG3ZvX7t9nAZPQT2Xxe3GT3tro
+        u5oCR+SyHN9xPnUwDuqUSvJ2eqMYb9B/Hph3OmtjG30jSNq9kOF5bBTk1hOTGPH4
+        K/AY0jzT6OpHfXU6ytlFsI47ZKsnTUhipGsKucQ1CXlyirndZ3V3k70YaooZ55rG
+        aIoAWlx2H0J7sAHmqS29N9jV9mo135d+d+TdLBXI0PXtiHzE9IPaX+ctdSUrPnp+
+        TwR99lxglpIG6hLuvOMAaxiqFBB/Jf3XJ8OBakfS6nHrWH2WqQxRbiITl0irkQoz
+        pwNEF2Bv0+Jvs1UFEdVGz5a8xexQHst/RmKrtHLct3iOCvBNqoAQRbvWvBhPjO/p
+        V5cYeUljZ5wpHyFkaEViClaVWqa6PIsyLqmyjsruPCWlURLsQoQxABcL8bwxX7UT
+        hM6CtH6tGlYZ85RIzRifIm2oudzV5l+8oRgFr9yVcwyOFT6JCioqkwldW52P1pk/
+        /SnuexC6LYqqDuHUs5NnokzzpfS6QaWfTY5P5tz4KHJfsjDIktly3mKVfY0fSPVV
+        okdGpcUzvz2hq1fqjxB6MlB/1vtk0bImfcsoxBmF7H+4E9ZN1sX/tSb0KQARAQAB
+        tCZPcGVuIFJvYm90aWNzIDxpbmZvQG9zcmZvdW5kYXRpb24ub3JnPokCVAQTAQoA
+        PhYhBMHPbjHmut6IaLFytPQu1vurF8ZUBQJc7yaWAhsDBQkDwmcABQsJCAcCBhUK
+        CQgLAgQWAgMBAh4BAheAAAoJEPQu1vurF8ZUkhIP/RbZY1ErvCEUy8iLJm9aSpLQ
+        nDZl5xILOxyZlzpg+Ml5bb0EkQDr92foCgcvLeANKARNCaGLyNIWkuyDovPV0xZJ
+        rEy0kgBrDNb3++NmdI/+GA92pkedMXXioQvqdsxUagXAIB/sNGByJEhs37F05AnF
+        vZbjUhceq3xTlvAMcrBWrgB4NwBivZY6IgLvl/CRQpVYwANShIQdbvHvZSxRonWh
+        NXr6v/Wcf8rsp7g2VqJ2N2AcWT84aa9BLQ3Oe/SgrNx4QEhA1y7rc3oaqPVu5ZXO
+        K+4O14JrpbEZ3Xs9YEjrcOuEDEpYktA8qqUDTdFyZrxb9S6BquUKrA6jZgT913kj
+        J4e7YAZobC4rH0w4u0PrqDgYOkXA9Mo7L601/7ZaDJob80UcK+Z12ZSw73IgBix6
+        DiJVfXuWkk5PM2zsFn6UOQXUNlZlDAOj5NC01V0fJ8P0v6GO9YOSSQx0j5UtkUbR
+        fp/4W7uCPFvwAatWEHJhlM3sQNiMNStJFegr56xQu1a/cbJH7GdbseMhG/f0BaKQ
+        qXCI3ffB5y5AOLc9Hw7PYiTFQsuY1ePRhE+J9mejgWRZxkjAH/FlAubqXkDgterC
+        h+sLkzGf+my2IbsMCuc+3aeNMJ5Ej/vlXefCH/MpPWAHCqpQhe2DET/jRSaM53US
+        AHNx8kw4MPUkxExgI7Sd
+        =4Ofr
+        -----END PGP PUBLIC KEY BLOCK-----
+
--- a/doc/basic_configuration.rst
+++ b/doc/basic_configuration.rst
@@ -76,7 +76,7 @@ and **not** the ``index.yaml`` in ``ros_buildfarm_config`` which you are editing
 
 This can be the official `ros/rosdistro <https://github.com/ros/rosdistro>`_'s ``index.yaml`` 
 if you intend to build the default set of packages, or your personal configuration where you intend to host
-``rosdistro`` (for example a GitHub fork of ``ros/rosdistro`` or your ``repo`` host) ::
+``rosdistro`` (for example a GitHub fork of ``ros/rosdistro`` or your ``repo`` host)::
 
   rosdistro_index_url: https://raw.githubusercontent.com/ros/rosdistro/master/index.yaml
 
--- a/doc/basic_configuration.rst
+++ b/doc/basic_configuration.rst
@@ -23,7 +23,7 @@ repository or create a repository containing the same configuration files.
 
 **Important:**
 Since ``ros_buildfarm_config`` files need to be accessed from within docker images,
-you **can not** use a local file system path (``file:/``) to reference them.
+you **can not** use a local file system path (`file://`) to reference them.
 You have to provide an http server where the configuration files can be accessed.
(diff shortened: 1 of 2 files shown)


doc/basic_configuration.rst
```diff
@@ -17,10 +17,17 @@ configuration to generate the necessary Jobs on the Jenkins master.
...
 --------------------------------------
 
-First you need to either fork the
+First you need to either fork or clone the
 `ros_buildfarm_config <https://github.com/ros-infrastructure/ros_buildfarm_config>`_
 repository or create a repository containing the same configuration files.
 
+**Important:**
+Since ``ros_buildfarm_config`` files need to be accessed from within docker images,
+you **can not** use a local file system path (``file:/``) to reference them.
```
gavanderhoorn:
```suggestion
you **can not** use a local file system path (`file://`) to reference them.
```

gavanderhoorn:
General comment: not entirely sure how strictly it is enfored in these documents, but there is a single-sentence-to-a-line policy in O(S)R(F) documentation sources. I believe there are some parts of these changes where that is not followed.

doc/basic_configuration.rst
max-krichenbauer: Thanks @gavanderhoorn 
I always thought the second dash was a idiosyncrasy of http and not required for ``file``.
Well, in practice it isn't required, but it sill seems to be the usual way of doing it:
https://en.wikipedia.org/wiki/File_URI_scheme

max-krichenbauer:
> General comment: not entirely sure how strictly it is enfored in these documents, but there is a single-sentence-to-a-line policy in O(S)R(F) documentation sources. I believe there are some parts of these changes where that is not followed.

Thanks! I didn't know.
I can change it if requested, but the original file didn't follow a single-sentence-to-a-line policy either, so I would have to change those parts that I didn't add as well (?).

doc/basic_configuration.rst
```diff
@@ -17,10 +17,17 @@ configuration to generate the necessary Jobs on the Jenkins master.
 Create / fork configuration repository
 --------------------------------------
 
-First you need to either fork the
+First you need to either fork or clone the
```
dirk-thomas: The note you mentioned is the main reason why simply cloning the repo to your local file system isn't sufficient. Therefore I think this should stay "fork"-only.

doc/basic_configuration.rst
```diff
@@ -17,10 +17,17 @@ configuration to generate the necessary Jobs on the Jenkins master.
...
 `ros_buildfarm_config <https://github.com/ros-infrastructure/ros_buildfarm_config>`_
 repository or create a repository containing the same configuration files.
 
+**Important:**
+Since ``ros_buildfarm_config`` files need to be accessed from within docker images,
+you **can not** use a local file system path (`file://`) to reference them.
+You have to provide an http server where the configuration files can be accessed.
+(Note that on the build farm master, Jenkins usually occupies the standard port 80).
```
dirk-thomas: I don't see how the port of Jenkins is relevant here? I don't think you want to start an HTTP server for this.

doc/basic_configuration.rst
```diff
@@ -35,16 +42,15 @@ The following options must be set in *all* build files::
...
+      maintainers: false
 
 
 Update administrator notification
 ---------------------------------
 
 Change the email address which gets notified about any administrative jobs.
 
-In the entry point yaml::
+In `ros-infrastructure/ros_buildfarm_config <https://github.com/ros-infrastructure/ros_buildfarm_config>`_'s `index.yaml`::
```
dirk-thomas: I don't think this should reference a specific repository since that is not the same for any custom buildfarm. Also the referenced branch in this case is incorrect.

Mentioning the filename `index.yaml` explicitly would indeed be good (rather than calling it "entry point yaml").

doc/basic_configuration.rst
```diff
@@ -55,17 +61,32 @@ In *all* build files::
     emails:
     - your_email@example.com
 
+Note that you need to have a local smtp service configured for email notifications.
```
dirk-thomas: This is part of the buildfarm deployment instructions. Maybe reference them here?

doc/basic_configuration.rst
```diff
@@ -55,17 +61,32 @@ In *all* build files::
...
+
+If you do not require email notifications, remove the configuration line entries.
```
dirk-thomas: Even when you remove these global config settings some jobs will still send notification email to package specific email addresses.

doc/basic_configuration.rst
```diff
@@ -55,17 +61,32 @@ In *all* build files::
...
 
 Update URLs to point to custom build farm
 -----------------------------------------
 
-Change the Jenkins URL in the entry point yaml to point to your earlier
-provisioned Jenkins master::
+Change the ``rosdistro_index_url`` in the `ros-infrastructure/ros_buildfarm_config <https://github.com/ros-infrastructure/ros_buildfarm_config>`_'s 
+``index.yaml`` to point to the ``rosdistro`` repository that defines the distro(s) that your buildfarm should build.
+
+**Important:** Note that this the `rosdistro <https://github.com/ros/rosdistro>`_'s ``index.yaml``
```
dirk-thomas: "Note that this the" ?

doc/basic_configuration.rst
```diff
@@ -55,17 +61,32 @@ In *all* build files::
...
+(which points to the files that define which packages are included in the distro and respective caches),
+and **not** the ``index.yaml`` in ``ros_buildfarm_config`` which you are editing.
+
+This will usually be on the provisioned ``repo`` host, but can be the official `ros/rosdistro <https://github.com/ros/rosdistro>`_'s ``index.yaml`` if you intend to build the default set of packages::
+
+  rosdistro_index_url: http://repo_hostname.example.com/rosdistro/index.yaml
```
dirk-thomas: Why would the `rosdistro` fork commonly be on the `repo` host? That sounds rather uncommon to me.

doc/basic_configuration.rst
```diff
@@ -100,3 +121,50 @@ repository::
 
 You can extract the repository PGP key from the ``buildfarm_deployment``
 configuration.
+
+
+Update URLs to point to required build tool repositories
```
dirk-thomas: Please don't use the term "build tool" for the repository here. "build tool" has a different semantic meaning. Maybe just: "Update URLs to point to required repositories".

doc/basic_configuration.rst
```diff
@@ -100,3 +121,50 @@ repository::
...
+      DiJVfXuWkk5PM2zsFn6UOQXUNlZlDAOj5NC01V0fJ8P0v6GO9YOSSQx0j5UtkUbR
+      fp/4W7uCPFvwAatWEHJhlM3sQNiMNStJFegr56xQu1a/cbJH7GdbseMhG/f0BaKQ
+      qXCI3ffB5y5AOLc9Hw7PYiTFQsuY1ePRhE+J9mejgWRZxkjAH/FlAubqXkDgterC
+      h+sLkzGf+my2IbsMCuc+3aeNMJ5Ej/vlXefCH/MpPWAHCqpQhe2DET/jRSaM53US
+      AHNx8kw4MPUkxExgI7Sd
+      =4Ofr
+      -----END PGP PUBLIC KEY BLOCK-----
+
+
+Note that the number of ``debian_repositories`` and ``debian_repository_keys`` must match, as well as the order in which they're specified.
```
dirk-thomas: This constraint is already documented in the `configuration_options` page. Maybe link to it instead?

scripts/generate_all_jobs.py
```diff
@@ -58,6 +58,13 @@ def main(argv=sys.argv[1:]):
     print('')
 
     config = get_index(args.config_url)
+    if args.config_url.startswith("file:"):
```
dirk-thomas: Nitpick: use single quotes here.

scripts/generate_all_jobs.py
```diff
@@ -58,6 +58,13 @@ def main(argv=sys.argv[1:]):
...
+        print('!WARNING!')
+        print('! Local file system path used for configuration.')
+        print('! Configuration will not be accessible to jobs during execution.')
+        print('! Consider using a web(http) hosted configuration repository.')
+        print('')
```
dirk-thomas: Please print warnings to `stderr`. Also a shorter message would be nice. I don't think this should be five lines long and contain `!` around each line.

doc/basic_configuration.rst
```diff
@@ -55,17 +61,32 @@ In *all* build files::
...
+
+If you do not require email notifications, remove the configuration line entries.
+
 
 Update URLs to point to custom build farm
 -----------------------------------------
 
-Change the Jenkins URL in the entry point yaml to point to your earlier
-provisioned Jenkins master::
+Change the ``rosdistro_index_url`` in the `ros-infrastructure/ros_buildfarm_config <https://github.com/ros-infrastructure/ros_buildfarm_config>`_'s
```
dirk-thomas: Same as above: I don't think this should reference the one repo used for the ROS buildfarm. The common case will be that a user setting up a custom buildfarm has its own fork of this so the added link isn't really helpful.

Same below another time.

doc/basic_configuration.rst
max-krichenbauer: It doesn't need to be a fork though, it doesn't even need to be a clone. A normal download is sufficient if even can be on the same machine. The only requirement is that it is accessible through http, since file system references won't work.

dirk-thomas: That is true. The documentation just describes the most common (and easy) use case. The user is always free to diverge from that recommendation. Setting up a HTTP server to serve the files isn't something I think this documentation should describe.

doc/basic_configuration.rst
max-krichenbauer: Well, you **have** to have an http server. Whether it's github (if you forked the ros_buildfarm_config repo), another server or localhost.
Since you create the jobs on the buildfarm master, I assume some users may try to use http://localhost, which is however occupied by Jenkins.

doc/basic_configuration.rst
max-krichenbauer: Thanks! Good point. I'll remove the link and make it more clear.

doc/basic_configuration.rst
max-krichenbauer: Thanks! Will add the reference.

doc/basic_configuration.rst
max-krichenbauer: Thanks. I didn't know. I'll add a note.

doc/basic_configuration.rst
max-krichenbauer: Maybe I'm mistaken, but I thought the meaning of this document was to list all the configuration items that **must** be addressed to set up a simple (standard) build farm. And as the ros_buildfarm_config stands right now, this item must be changed for the build farm to work (at least the one listed in the repo right now is not reachable for me).
So I thought it would be good to explain this here.

doc/basic_configuration.rst
max-krichenbauer: Probably I was mistaken - I though ``ros_buildfarm_deployment`` automatically created folders ``rosdistro`` and ``rosdistro_cache`` on the ``repo`` host for this purpose. But now I'm not sure anymore.
Where do people usually host ``rosdistro`` if not on the repo host?

scripts/generate_all_jobs.py
max-krichenbauer: Thanks for the comment. Updated.

doc/basic_configuration.rst
max-krichenbauer: Thanks! Updated. I hope it's clear enough now.

doc/basic_configuration.rst
max-krichenbauer: Thanks! Fixed.

doc/basic_configuration.rst
max-krichenbauer: Thanks! Fixed.

doc/basic_configuration.rst
max-krichenbauer: Thanks! Fixed.

max-krichenbauer:
Somehow, the travis-ci check keeps failing, but it seems to be unrelated to my changes. The error message is:
````
    _____________________ ERROR collecting test/test_pep257.py _____________________
    ImportError while importing test module 
    '/tmp/ws3/src/ament/ament_lint/ament_copyright/test/test_pep257.py'.
    Hint: make sure your test modules/packages have valid Python names.
    Traceback:
    test/test_pep257.py:15: in <module>
        from ament_pep257.main import main
    /tmp/ws2/install_isolated/ament_pep257/lib/python3.6/site-packages/ament_pep257/main.py:26: 
    in <module>
        import pydocstyle
    E   ModuleNotFoundError: No module named 'pydocstyle'
    !!!!!!!!!!!!!!!!!!! Interrupted: 1 errors during collection !!!!!!!!!!!!!!!!!!!!
````
I'm unable to do anything about that and would the thankful if someone could let me know if there is any problem.

File: doc/basic_configuration.rst (hunk 8 above)
Comment (dirk-thomas) [reply doc/basic_configuration.rst]: > Where do people usually host `rosdistro` if not on the repo host?

On GitHub - same as the upstream repo.

File: doc/basic_configuration.rst (hunk 10 above)
Comment (dirk-thomas) [reply doc/basic_configuration.rst]: You can certainly mention the configuration item here. But I don't think it makes sense to re-document how to use them if that is already covered in other parts of the documentation. A reference to that doc avoids duplicating information.

File: scripts/generate_all_jobs.py [hunk 14]
```diff
@@ -58,6 +58,12 @@ def main(argv=sys.argv[1:]):
     print('')
 
     config = get_index(args.config_url)
+    if args.config_url.startswith('file:'):
+        sys.stderr.write('WARNING: Local file system path used for ',
```
dirk-thomas: You should use `sys.stderr` for this since e.g. now you lack a trailing newline. Instead use: `print('...', file=sys.stderr)`

Also please don't align subsequent line.

dirk-thomas:
> the travis-ci check keeps failing, but it seems to be unrelated to my changes. The error message is:

I didn't find the error message in the logs. Can you please point to the exact build.

One of the Travis sub-builds (`.4` - `JOB_TYPE=ci`) is known to fail atm. #675 aims to address that but is waiting for a sync of ROS 2 Dashing which ships a fix.

max-krichenbauer:
> > the travis-ci check keeps failing, but it seems to be unrelated to my changes. The error message is:
> 
> I didn't find the error message in the logs. Can you please point to the exact build.
> 
> One of the Travis sub-builds (`.4` - `JOB_TYPE=ci`) is known to fail atm. #675 aims to address that but is waiting for a sync of ROS 2 Dashing which ships a fix.

Yes, I think that's the one:
``Build 1004.4 JOB_TYPE=ci UNDERLAY1_PACKAGES="ament_flake8" UNDERLAY2_PACKAGES="ament_pep257" OVERLAY_PACKAGES="ament_cmake_ros"``
I wasn't sure if it was something I was doing wrong.

scripts/generate_all_jobs.py
max-krichenbauer: Thanks. I changed it to ``print('...', file=sys.stderr)`` .
About the subsequent lines, this is the indentation enforced by both flake8 and pylint, so I don't know how to change it without failing more checks.

doc/basic_configuration.rst
max-krichenbauer: Thanks. Changed it.

doc/basic_configuration.rst
max-krichenbauer: Thanks. I changed the wording accordingly.

doc/basic_configuration.rst
```diff
@@ -55,17 +59,34 @@ In *all* build files::
     emails:
     - your_email@example.com
 
+You need to have a `local smtp service configured <https://github.com/ros-infrastructure/buildfarm_deployment#setup-master-for-email-delivery>`_ for email notifications.
```
dirk-thomas: This note lacks any context that it is meant for the Jenkins master.

doc/basic_configuration.rst
```diff
@@ -55,17 +59,34 @@ In *all* build files::
...
+Note that even when you remove these global email notification settings
+some jobs will still send notification emails to package specific email addresses.
```
dirk-thomas: Maybe note that this is configuration dependent. It can be disabled - by default it will send notification emails though.

doc/basic_configuration.rst
```diff
@@ -55,17 +59,34 @@ In *all* build files::
...
 -----------------------------------------
 
-Change the Jenkins URL in the entry point yaml to point to your earlier
-provisioned Jenkins master::
+Change the ``rosdistro_index_url`` in your ``ros_buildfarm_config``'s 
+``index.yaml`` to point to the ``rosdistro`` repository that defines the distro(s) that your buildfarm should build.
+
+**Important:** Note that this is the `rosdistro <https://github.com/ros/rosdistro>`_'s ``index.yaml``
+(which points to the files that define which packages are included in the distro and respective caches),
+and **not** the ``index.yaml`` in ``ros_buildfarm_config`` which you are editing.
```
dirk-thomas: Isn't the previous sentence already clarifying which repository the `index.yaml` file belongs to? This seems redundant to me.

doc/basic_configuration.rst
```diff
@@ -100,3 +121,12 @@ repository::
...
 configuration.
+
+
+Update URLs to point to required repositories
+---------------------------------------------
+
+During job execution, access to repositories which contain the necessary tools to run the ROS build farm is required.
+These must be specified in your ``ros_buildfarm_config``'s ``index.yaml`` as ``prerequisites``.
+You can use the official ROS repository or mirrors of it.
+See the `Configuration Options <https://github.com/max-krichenbauer/ros_buildfarm/blob/documentation-improvements/doc/configuration_options.rst#entry-point-yaml>`_ documentation page for details.
```
dirk-thomas: This shouldn't link to your fork but be relative within this repo.

scripts/generate_all_jobs.py
```diff
@@ -58,6 +58,13 @@ def main(argv=sys.argv[1:]):
     print('')
 
     config = get_index(args.config_url)
+    if args.config_url.startswith('file:'):
+        print('WARNING: Local file system path used for ',
+              'configuration. Configuration will not be ',
+              'accessible to jobs during execution. Consider ',
+              'using a web(http) hosted configuration repository.',
+              file=sys.stderr)
```
dirk-thomas: Just as a note - the current state can be kept as is:

> About the subsequent lines, this is the indentation enforced by both flake8 and pylint, so I don't know how to change it without failing more checks.

If you don't wrap right after the open parenthesis the following lines must be indented to the same level. Commonly if any arguments are wrapped we ensure to wrap *all* of them, so having a newline right after the open parenthesis. That way all arguments are indented exactly one level.

doc/basic_configuration.rst
max-krichenbauer: Thanks. Added.

doc/basic_configuration.rst
max-krichenbauer: Thanks. Added a note.

doc/basic_configuration.rst
max-krichenbauer: Okay. I removed it in the last commit.
Personally I felt that since both files have the same name, stressing the difference here would be a good idea to avoid confusion.

doc/basic_configuration.rst
max-krichenbauer: Oh, sorry - seems like I copy-pasted from the wrong browser tab.
Fixed now.

scripts/generate_all_jobs.py
max-krichenbauer: Oh, now I understand what you mean. Thanks.
While I was already editing the code I now changed it as well.

doc/basic_configuration.rst
dirk-thomas: Please use a relative link instead.

max-krichenbauer: Done.

dirk-thomas:
Thanks for improving the docs.

Issue #678, opened 2019-09-12
Implementation of abichecker for ROS2


j-rivero:
To implement the ABI checking of ROS2 package I have some doubts that would like to resolve before starting the works:

* Which scenarios should be covered by the ABI checking? I'm pretty sure that we probably want it for PR testing to check changes against current stable version. Any other context (check testing against stable, ...)?

* How are we going to identify which packages needs ABI checking? Adding an option in build/release files is the first thing that comes to my mind, not sure if we have a better idea.

* Do we need to create new jobs for ABI checking or should we try to reuse the existing ones? In the case of PR testing, the abi-checking can be added to existing `[A-Z]pr__` jobs. Please note that ABI checking requires the use of `-g -Og` flags during compilation so we can not reuse the current pr__ build.

j-rivero:
After speaking with @chapulina and start working on the project, I've some updates for some of my questions:

>     * Which scenarios should be covered by the ABI checking? I'm pretty sure that we probably want it for PR testing to check changes against current stable version. Any other context (check testing against stable, ...)?

The PR testing jobs are main and natural scope for running the ABI checking. The best option is to test ABI against the stable .deb packages.

>     * How are we going to identify which packages needs ABI checking? Adding an option in build/release files is the first thing that comes to my mind, not sure if we have a better idea.

Still an open question.

>     * Do we need to create new jobs for ABI checking or should we try to reuse the existing ones? In the case of PR testing, the abi-checking can be added to existing `[A-Z]pr__` jobs. Please note that ABI checking requires the use of `-g -Og` flags during compilation so we can not reuse the current pr__ build.

I've checked that the tool can perfectly run on code compiling without special debug flags so the fastest and easiest way would be to use the compiled code in the existing PRs.

dirk-thomas:
> How are we going to identify which packages needs ABI checking?

It probable depends on how much overhead this check adds to a job. If it is fairly minor I don't see a reason why not to do it for every package / repo.

> the abi-checking can be added to existing `[A-Z]pr__` jobs

It might be good to not only run the check in PR jobs but also in devel jobs to keep seeing the notice / warning more than once. Otherwise after merging a PR there will be no further notifications.

j-rivero:
Summarizing: 

[...]

PR #681, opened 2019-10-02
Implement the abichecker run on devel builds

j-rivero:
The PR tries to implement the ABI checking for ROS2 packages (issue #678).

The approach followed was to use implement the option of abi checking in the `build_and_install.py` script used by the devel scripts. I've implement an argument in the script named `--abi-checker` that runs the [auto-abi-checker](https://github.com/osrf/auto-abi-checker) tool using:
 1. The workspace installation space
 1. The debian packages associated to the ROS repository under testing.

**Testing:**
 * Using the `build_and_install.py` script directly on my local system the runs of simple packages (`rclcpp`) was succesful.
 * I'm not sure about the best wait to test this in the ROS2 build farm. Last time I used a testing instance of a buildfarm (I think @nuclearsandwich is generating one these days). If there is a faster approach, please let me know.

**Open questions and TODO:**
 * Still need to resolve the question in #678 about how are we going to identify which packages needs ABI. The easiest thing would be the abi-checker always. If the package does not provide any `.h` or `.so` the checker will probably do nothing and report success.
 * Export results in Jenkins (the osrf buildfarm uses [HTML publisher](https://wiki.jenkins.io/display/JENKINS/HTML+Publisher+Plugin) + [Console parser](https://wiki.jenkins.io/display/JENKINS/Console+Parser+Plugin) to make build unstable.)
 * The local test build fails miserably on packages that has msgs generation support due to the lack of headers related to context and opensplice (not sure if the buildfarm is installing all the support for them). Support to remove the headers of these DDS implementations can be added to the `auto-abi-checker`.
 * There are problems with packages (i.e: `gazebo_ros_pkgs`) that introduce boost headers into the mix of headers to check (some expected definitions are missing). I'm still investigating them.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -197,8 +197,6 @@ matrix:
         - docker
       env: JOB_TYPE=external_prerelease
       before_script:
-        - sudo sh -c 'echo "deb http://packages.ros.org/ros/ubuntu $(lsb_release -sc) main" > /etc/apt/sources.list.d/ros.list'
-        - *rosgpgkeyadd
         - sudo apt-get update
         - sudo apt-get install -y python3-pip python3-setuptools
         - sudo pip3 install vcstool
@@ -302,8 +300,6 @@ matrix:
         - docker
       env: JOB_TYPE=external_prerelease
       before_script:
-        - sudo sh -c 'echo "deb http://packages.ros.org/ros/ubuntu $(lsb_release -sc) main" > /etc/apt/sources.list.d/ros.list'
-        - *rosgpgkeyadd
         - sudo apt-get update
         - sudo apt-get install -y python3-pip python3-setuptools
         - sudo pip3 install vcstool
@@ -407,8 +403,6 @@ matrix:
         - docker
       env: JOB_TYPE=external_prerelease
       before_script:
-        - sudo sh -c 'echo "deb http://packages.ros.org/ros/ubuntu $(lsb_release -sc) main" > /etc/apt/sources.list.d/ros.list'
-        - *rosgpgkeyadd
         - sudo apt-get update
         - sudo apt-get install -y python3-pip  python3-setuptools
         - sudo pip3 install vcstool
@@ -491,8 +485,6 @@ matrix:
         - docker
       env: JOB_TYPE=prerelease UNDERLAY_REPOSITORY_NAMES="roscpp_core" OVERLAY_PACKAGE_NAMES=roscpp
       before_script:
-        - sudo sh -c 'echo "deb http://packages.ros.org/ros/ubuntu $(lsb_release -sc) main" > /etc/apt/sources.list.d/ros.list'
-        - *rosgpgkeyadd
         - sudo apt-get update
         - sudo apt-get install -y python3-pip python3-setuptools
         - sudo pip3 install EmPy
@@ -517,8 +509,6 @@ matrix:
         - docker
       env: JOB_TYPE=external_prerelease
       before_script:
-        - sudo sh -c 'echo "deb http://packages.ros.org/ros/ubuntu $(lsb_release -sc) main" > /etc/apt/sources.list.d/ros.list'
-        - *rosgpgkeyadd
         - sudo apt-get update
         - sudo apt-get install -y python3-pip python3-setuptools
         - sudo pip3 install EmPy vcstool
--- a/.travis.yml
+++ b/.travis.yml
@@ -197,8 +197,10 @@ matrix:
         - docker
       env: JOB_TYPE=external_prerelease
       before_script:
+        - sudo sh -c 'echo "deb http://packages.ros.org/ros/ubuntu $(lsb_release -sc) main" > /etc/apt/sources.list.d/ros.list'
+        - *rosgpgkeyadd
         - sudo apt-get update
-        - sudo apt-get install -y python3-pip python3-setuptools
+        - sudo apt-get install -y python3-pip python3-setuptools python3-catkin-pkg-modules
         - sudo pip3 install vcstool
         - python setup.py install
         - mkdir job && cd job
@@ -285,6 +287,7 @@ matrix:
         - sudo apt-get update -qq
         - sudo apt-get install dpkg -y  # necessary for catkin-pkg to be installable
         - sudo apt-get install ros-kinetic-catkin -y
+        - sudo apt-get install python3-catkin-pkg-modules -y
         - python setup.py install
         - mkdir job && cd job
         - ln -s .. ros_buildfarm
@@ -300,8 +303,10 @@ matrix:
         - docker
       env: JOB_TYPE=external_prerelease
       before_script:
+        - sudo sh -c 'echo "deb http://packages.ros.org/ros/ubuntu $(lsb_release -sc) main" > /etc/apt/sources.list.d/ros.list'
+        - *rosgpgkeyadd
         - sudo apt-get update
-        - sudo apt-get install -y python3-pip python3-setuptools
+        - sudo apt-get install -y python3-pip python3-setuptools python3-catkin-pkg-modules
         - sudo pip3 install vcstool
         - python setup.py install
         - mkdir job && cd job
@@ -388,6 +393,7 @@ matrix:
         - sudo apt-get update -qq
         - sudo apt-get install dpkg -y  # necessary for catkin-pkg to be installable
         - sudo apt-get install ros-kinetic-catkin -y
+        - sudo apt-get install python3-catkin-pkg-modules -y
         - python setup.py install
         - mkdir job && cd job
         - ln -s .. ros_buildfarm
@@ -403,8 +409,10 @@ matrix:
         - docker
       env: JOB_TYPE=external_prerelease
       before_script:
+        - sudo sh -c 'echo "deb http://packages.ros.org/ros/ubuntu $(lsb_release -sc) main" > /etc/apt/sources.list.d/ros.list'
+        - *rosgpgkeyadd
         - sudo apt-get update
-        - sudo apt-get install -y python3-pip  python3-setuptools
+        - sudo apt-get install -y python3-pip  python3-setuptools python3-catkin-pkg-modules
         - sudo pip3 install vcstool
         - python setup.py install
         - mkdir job && cd job
@@ -485,8 +493,10 @@ matrix:
         - docker
       env: JOB_TYPE=prerelease UNDERLAY_REPOSITORY_NAMES="roscpp_core" OVERLAY_PACKAGE_NAMES=roscpp
       before_script:
+        - sudo sh -c 'echo "deb http://packages.ros.org/ros/ubuntu $(lsb_release -sc) main" > /etc/apt/sources.list.d/ros.list'
+        - *rosgpgkeyadd
         - sudo apt-get update
-        - sudo apt-get install -y python3-pip python3-setuptools
+        - sudo apt-get install -y python3-pip python3-setuptools python3-catkin-pkg-modules
         - sudo pip3 install EmPy
         # install catkin for test results
         - sudo sh -c 'echo "deb http://packages.ros.org/ros/ubuntu $(lsb_release -sc) main" > /etc/apt/sources.list.d/ros.list'
@@ -509,8 +519,10 @@ matrix:
         - docker
       env: JOB_TYPE=external_prerelease
       before_script:
+        - sudo sh -c 'echo "deb http://packages.ros.org/ros/ubuntu $(lsb_release -sc) main" > /etc/apt/sources.list.d/ros.list'
+        - *rosgpgkeyadd
         - sudo apt-get update
-        - sudo apt-get install -y python3-pip python3-setuptools
+        - sudo apt-get install -y python3-pip python3-setuptools python3-catkin-pkg-modules
         - sudo pip3 install EmPy vcstool
         - python setup.py install
         - mkdir job && cd job
--- a/ros_buildfarm/templates/devel/devel_create_tasks.Dockerfile.em
+++ b/ros_buildfarm/templates/devel/devel_create_tasks.Dockerfile.em
@@ -78,7 +78,7 @@ cmd = \
     ' --ros-version ' + str(ros_version) + \
     ' --env-vars ' + ' ' .join(env_vars)
 if run_abichecker:
-    cmd += " --run-abichecker"
+    cmd += ' --run-abichecker'
 cmds += [
     cmd +
     ' --dockerfile-dir /tmp/docker_build_and_install',
--- a/ros_buildfarm/templates/devel/devel_create_tasks.Dockerfile.em
+++ b/ros_buildfarm/templates/devel/devel_create_tasks.Dockerfile.em
@@ -83,6 +83,7 @@ cmds += [
     cmd +
     ' --dockerfile-dir /tmp/docker_build_and_test' +
     ' --testing',
+    ' --abi-checking',
 ]
 }@
 CMD ["@(' && '.join(cmds))"]
--- a/ros_buildfarm/templates/devel/devel_create_tasks.Dockerfile.em
+++ b/ros_buildfarm/templates/devel/devel_create_tasks.Dockerfile.em
@@ -82,7 +82,7 @@ cmds += [
     ' --dockerfile-dir /tmp/docker_build_and_install',
     cmd +
     ' --dockerfile-dir /tmp/docker_build_and_test' +
-    ' --testing',
+    ' --testing' +
     ' --abi-checking',
 ]
 }@
--- a/ros_buildfarm/templates/devel/devel_create_tasks.Dockerfile.em
+++ b/ros_buildfarm/templates/devel/devel_create_tasks.Dockerfile.em
@@ -83,7 +83,7 @@ cmds += [
     cmd +
     ' --dockerfile-dir /tmp/docker_build_and_test' +
     ' --testing' +
-    ' --abi-checking',
+    ' --run-abichecker',
 ]
 }@
 CMD ["@(' && '.join(cmds))"]
--- a/ros_buildfarm/templates/devel/devel_create_tasks.Dockerfile.em
+++ b/ros_buildfarm/templates/devel/devel_create_tasks.Dockerfile.em
@@ -77,13 +77,14 @@ cmd = \
     ' --build-tool ' + build_tool + \
     ' --ros-version ' + str(ros_version) + \
     ' --env-vars ' + ' ' .join(env_vars)
+if run_abichecker:
+    cmd += " --run-abichecker"
 cmds += [
     cmd +
     ' --dockerfile-dir /tmp/docker_build_and_install',
     cmd +
     ' --dockerfile-dir /tmp/docker_build_and_test' +
-    ' --testing' +
-    ' --run-abichecker',
+    ' --testing'
 ]
 }@
 CMD ["@(' && '.join(cmds))"]
--- a/ros_buildfarm/templates/devel/devel_job.xml.em
+++ b/ros_buildfarm/templates/devel/devel_job.xml.em
@@ -111,11 +111,6 @@ if pull_request:
     'builder_shell_key-files',
     script_generating_key_files=script_generating_key_files,
 ))@
-@[if run_abichecker]
-@{ str_abichecker = ' --run-abichecker' }
-@[else]
-@{ str_abichecker = '' }
-@[end if]
 @(SNIPPET(
     'builder_shell',
     script='\n'.join([
@@ -143,7 +138,7 @@ if pull_request:
         ' ' + ' '.join(repository_args) +
         ' --build-tool ' + build_tool +
         ' --ros-version ' + str(ros_version) +
-        str_abichecker +
+        (' --run-abichecker' if run_abichecker else '') +
         ' --env-vars ' + ' '.join(build_environment_variables) +
         ' --dockerfile-dir $WORKSPACE/docker_generating_dockers',
         'echo "# END SECTION"',
--- a/ros_buildfarm/templates/devel/devel_job.xml.em
+++ b/ros_buildfarm/templates/devel/devel_job.xml.em
@@ -111,6 +111,11 @@ if pull_request:
     'builder_shell_key-files',
     script_generating_key_files=script_generating_key_files,
 ))@
+@[if run_abichecker]
+@{ str_abichecker = ' --run-abichecker' }
+@[else]
+@{ str_abichecker = '' }
+@[end if]
 @(SNIPPET(
     'builder_shell',
     script='\n'.join([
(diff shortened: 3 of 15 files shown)


ros_buildfarm/workspace.py
```diff
@@ -41,6 +42,32 @@ def clean_workspace(workspace_root):
     if os.path.exists(test_results_dir):
         shutil.rmtree(test_results_dir)
 
+def call_abi_checker(workspace_root, rosdistro_name, env):
+    # TODO: pkgs detection, code based on create_devel_task_generator.py
+    condition_context = {}
+    condition_context['ROS_DISTRO'] = rosdistro_name
```
dirk-thomas: The dict needs to be populate with more information: `ROS_VERSION`, `ROS_PYTHON_VERSION`.

ros_buildfarm/workspace.py
```diff
@@ -41,6 +42,32 @@ def clean_workspace(workspace_root):
...
+        source_space = os.path.join(ws_root, 'src')
+        ws_pkgs = find_packages(source_space)
+        for pkg in ws_pkgs.values():
+            pkg.evaluate_conditions(condition_context)
+        pkgs.update(ws_pkgs)
+    pkg_names = [pkg.name for pkg in pkgs.values()]
+
+    # TODO: workspace_root[0] to be compatible with a code in
+    # create_devel_task_generator for a future refactor. To fix
+    # it, implement the support for multiple local-dir in auto-abi tool
```
dirk-thomas: Instead of a TODO comment how about asserting that `len(workspace_root) == 1`? That way if it is being used with multiple paths in the future it will actually fail instead of silently using only the first.

scripts/devel/build_and_install.py
```diff
@@ -21,7 +21,7 @@
 from ros_buildfarm.argument import add_argument_build_tool
 from ros_buildfarm.argument import add_argument_build_tool_args
 from ros_buildfarm.common import Scope
-from ros_buildfarm.workspace import call_build_tool
+from ros_buildfarm.workspace import call_build_tool, call_abi_checker
```
dirk-thomas: Nitpick: one import per line - in alphabetical order.

scripts/devel/build_and_install.py
```diff
@@ -53,20 +53,26 @@ def main(argv=sys.argv[1:]):
         action='store_true',
         help='The flag if the workspace should be cleaned after the '
              'invocation')
+    parser.add_argument(
+        '--run-abichecker',
+        dest='abichecking',
```
dirk-thomas: Nitpick: please remove the custom `dest`. The default destination `run_abichecker` is just fine.

scripts/devel/build_and_install.py
```diff
@@ -77,7 +83,16 @@ def main(argv=sys.argv[1:]):
...
 
-    return rc
+    if not args.abichecking:
+        return rc
+
+    with Scope('SUBSECTION', 'use abi checker'):
+        rc = call_abi_checker([args.workspace_root], args.rosdistro_name, env=env)
+
+    if rc != 0:
+        print("Failure during the execution of abi-checking")
```
dirk-thomas: Nitpick: use single quotes.

scripts/devel/build_and_install.py
```diff
@@ -77,7 +83,16 @@ def main(argv=sys.argv[1:]):
...
+        return rc
+
```
dirk-thomas: Nitpick: remove extra empty line.

scripts/devel/create_devel_task_generator.py
```diff
@@ -71,6 +71,11 @@ def main(argv=sys.argv[1:]):
         action='store_true',
         help='The flag if the workspace should be built with tests enabled '
              'and instead of installing the tests are ran')
+    parser.add_argument(
+        '--run-abichecker',
+        dest='abichecking',
```
dirk-thomas: Nitpick: please remove the custom `abichecking`. The default destination `run_abichecker` is just fine.

scripts/devel/create_devel_task_generator.py
```diff
@@ -186,6 +191,7 @@ def main(argv=sys.argv[1:]):
         'install_lists': [],
 
         'testing': args.testing,
+        'abichecking': args.abichecking,
```
dirk-thomas: Nitpick: please match the key with the argument `run_abichecker`.

dirk-thomas:
Btw. the current state severely breaks CI which needs to be address before it even makes sense to test deploy this.

ros_buildfarm/workspace.py
j-rivero: good idea, 8641225

scripts/devel/build_and_install.py
j-rivero: cb68012

scripts/devel/build_and_install.py
j-rivero: ff5cffd

scripts/devel/build_and_install.py
j-rivero: 2ab7c8d

scripts/devel/create_devel_task_generator.py
j-rivero: 7d60961

scripts/devel/create_devel_task_generator.py
j-rivero: 7d60961

ros_buildfarm/workspace.py
j-rivero: Thanks for the hint, 1d6db93

j-rivero:
> **Open questions and TODO:**
> 
>     * Still need to resolve the question in #678 about how are we going to identify which packages needs ABI. The easiest thing would be the abi-checker always. If the package does not provide any `.h` or `.so` the checker will probably do nothing and report success.
> 

Following Dirk's advice on https://github.com/ros-infrastructure/ros_buildfarm/issues/678#issuecomment-540821716, it runs on all jobs. I added a feature to measure time in auto-abi tool: 
 * For a non trivial c++ package like `aws_common` the overhead is 7:50 minutes.
 * For non c++ packages like `rclpy` the overhead is 2 seconds

>     * Export results in Jenkins (the osrf buildfarm uses [HTML publisher](https://wiki.jenkins.io/display/JENKINS/HTML+Publisher+Plugin) + [Console parser](https://wiki.jenkins.io/display/JENKINS/Console+Parser+Plugin) to make build unstable.)
>

HTML report publication has been added https://github.com/ros-infrastructure/ros_buildfarm/pull/681/files#diff-c4331714bf7e4850ab74f65a38e6bb45 . We need to remember to install the Jenkins plugin.
 
>     * The local test build fails miserably on packages that has msgs generation support due to the lack of headers related to context and opensplice (not sure if the buildfarm is installing all the support for them). Support to remove the headers of these DDS implementations can be added to the `auto-abi-checker`.
> 

Support to filter non `fastrtps` message support has been added to the auto-abi tool. Message package can be checked fine now: http://18.206.154.49/job/Ddev__ackermann_msgs__ubuntu_bionic_amd64/

**Testing and TODO:**

I've imported all the packages in the ROS2 buildfarm and run the abi on all of them:
http://18.206.154.49/ . Some packages are failing for different reasons: 
 * Some are valid ABI breakages (dashing packages against current dev branch)
 * Most of the failures are related to missing headers probably due to errors in dependencies declared upstream or other problems related to upstream. The list is not short.
 * Other errors can probably be fixed adding support into the auto-abi tool.

At this point we can decide: 
 1. To invest time hunting all errors, one by one, sending corresponding patches to upstream where needed.
 1. Make the failure of the ABI checker to trigger an unstable state and merge things how they are now. Ask maintainers to react to the unstable state or dedicate time to fix unstable states by our side.
 1. Run abi-checker without failing the build on errors and keep fixing errors or reporting them to upstream before we change to enable errors again.

chapulina:
Thanks, @j-rivero !

> At this point we can decide:

If we're sure the false positives are not coming from the infrastructure, my suggestion would be to go with number 2 as a way to warn maintainers. But I'll defer the decision to someone more experienced with the ROS buildfarm and the implicit contract between us and maintainers.

nuclearsandwich:
> Most of the failures are related to missing headers probably due to errors in dependencies declared upstream or other problems related to upstream. The list is not short.

Can you put the list in a gist or other shareable paste with links to the testfarm jobs. The problem with accepting instability is that it makes spotting new issues that much more difficult since the overall job state doesn't change.
We certainly don't want to deploy ABI checking on Dashing against mainline as mainline has moved on, the Dashing ABI check should target the dashing devel branch but as a test that the ABI checker catches legitimate breaks it's great.

dirk-thomas:
> We certainly don't want to deploy ABI checking on Dashing against mainline as mainline

I assumed that was implicit. Every dev/PR job should only check against the Debian package of the same distro.

I would also pick option 2. The general decision if ABI checking should be performed should be specified in a build file though (similar to `test_commits` and `test_pull_requests`: https://github.com/ros-infrastructure/ros_buildfarm/blob/3def27c13479693b538bc212c66bb23a34e1c201/doc/configuration_options.rst#specific-options-in-source-build-files). That way we can opt in to the feature on a per build file level (likely for future distros first, and after some parallel testing maybe for existing distros).

j-rivero:
I've implemented the option to make the builds not to fail but make them unstable if the abi checker failed. It used a console output parser to make the build unstable, not sure if you are using something different in the build farm to make builds unstable. See test job with [ml_classifiers](http://18.206.154.49/job/Ddev__ml_classifiers__ubuntu_bionic_amd64/10/parsed_console/)

> Can you put the list in a gist or other shareable paste with links to the testfarm jobs. The problem with accepting instability is that it makes spotting new issues that much more difficult since the overall job state doesn't change.

https://gist.github.com/j-rivero/73294c471c61ff46dba635c3d2020110#gistcomment-3063972 . Most of the entries include a comment I made about why is failing.

> I assumed that was implicit. Every dev/PR job should only check against the Debian package of the same distro.

It was implicit for me. I used [this file](https://raw.githubusercontent.com/j-rivero/ros_buildfarm_config_steven/deployment/2019-10-07/index.yaml ) with generate_devel_script to get only Dashing jobs. Not sure if I did something wrong. In the previous gist you can grep for `compat report is bad`. Not fully sure that are real breakages.

> The general decision if ABI checking should be performed should be specified in a build file though (similar to `test_commits` and `test_pull_requests`)

I still need to implement this.

j-rivero:
> > The general decision if ABI checking should be performed should be specified in a build file though (similar to `test_commits` and `test_pull_requests`)
> 
> I still need to implement this.

This was implemented in 4b5517e

j-rivero:
I've managed to fix all the CI options and send the pull request to implement the `test_abi` option in rosdistro files. Ready for review.

dirk-thomas:
> Some packages are failing for different reasons:
>
> 1. Some are valid ABI breakages (dashing packages against current dev branch)
> 2. Most of the failures are related to missing headers probably due to errors in dependencies declared upstream or other problems related to upstream. The list is not short.
> 3. Other errors can probably be fixed adding support into the auto-abi tool.

@j-rivero Can you clarify what the latest status on these are?

ros_buildfarm/templates/devel/devel_job.xml.em
```diff
@@ -361,13 +367,21 @@ if pull_request:
 @(SNIPPET(
     'publisher_groovy-postbuild_maintainer-notification',
 ))@
-@[ end if]@
+@[end if]@
```
nuclearsandwich: I think this space was intentional to show the nesting level of the conditional without producing an artifact in the output file.

ros_buildfarm/templates/devel/devel_task.Dockerfile.em
```diff
@@ -62,7 +62,7 @@ RUN pip3 install -U setuptools
 @[if ros_version == 2]@
 RUN python3 -u /tmp/wrapper_scripts/apt.py update-install-clean -q -y ros-@(rosdistro_name)-ros-workspace
 @[end if]@
-RUN python3 -u /tmp/wrapper_scripts/apt.py update-install-clean -q -y ccache
+RUN python3 -u /tmp/wrapper_scripts/apt.py update-install-clean -q -y ccache python3-catkin-pkg-modules
```
nuclearsandwich: Should `python3-catkin-pkg-modules` be moved to the apt invocation below so that it's only installed when needed for the ABI checker? That might also reduce the need to install it for some of the travis tests.

scripts/devel/build_and_install.py
```diff
@@ -16,16 +16,51 @@
...
+        source_space = os.path.join(ws_root, 'src')
+        ws_pkgs = find_packages(source_space)
+        for pkg in ws_pkgs.values():
+            pkg.evaluate_conditions(condition_context)
+        pkgs.update(ws_pkgs)
+    pkg_names = [pkg.name for pkg in pkgs.values()]
+    assert(pkg_names), 'No packages found in the workspace'
+
+    assert(len(workspace_root) == 1), 'auto-abi tool needs the implementation of multiple local-dir'
+    cmd = ['ROS_DISTRO=' + rosdistro_name + ' ' +
```
nuclearsandwich: It seems strange to be passing the environment through and also using the cmdline to set an environment variable. Is there a reason not to set this in the env object?

scripts/devel/build_and_install.py
```diff
@@ -77,7 +117,22 @@ def main(argv=sys.argv[1:]):
...
+        rc = call_abi_checker(
+            [args.workspace_root],
+            args.ros_version,
+            args.rosdistro_name,
+            env)
+    # Never fail a build because of abi errors but make them
+    # unstable by printing MAKE_BUILD_UNSTABLE. Jenkins will
+    # use a`plugin to make it
+    if rc != 0:
+        print('MAKE_BUILD_UNSTABLE')
```
nuclearsandwich: Since it appears that we control the parser for this string. Can we use a more precise one like "ABI_CHECK_FAILED" so that it's immediately clear where to look for the failure?

ros_buildfarm/templates/devel/devel_job.xml.em
j-rivero: Thanks. Fixed in dbbb07c

ros_buildfarm/templates/devel/devel_task.Dockerfile.em
j-rivero: It was also a good idea to me. After implemented it, travis pointed me that [we are including the abi-checker code unconditionally inside the build_and_install.py](https://github.com/ros-infrastructure/ros_buildfarm/pull/681/files#diff-14a2d9d1f397cfe07f871fdb61dc0483R22) file.

j-rivero:
> One overall nitpick is that the argument/variable for tracking whether ABI checks should be performed changes a number of times when it doesn't need to `test_abi` -> `enable_abi` -> `abi_checking` -> `run_abichecker`. While I don't think they _all_ need to be condensed into one. I see value in using either `run_abichecker` or `test_abi` through the middle layers so that we only switch terms once.

[I moved](https://github.com/j-rivero/ros_buildfarm/pull/5/commits/94eb8a92a38575f1b86d6c968a0f6efd7b10b455) all references to code to use `run_abichecker`

scripts/devel/build_and_install.py
j-rivero: No reason. Makes sense to me. [Implemented](https://github.com/j-rivero/ros_buildfarm/pull/5/commits/e8ee6a346a97e83a2e80653e8ecd9732007482ba).

scripts/devel/build_and_install.py
j-rivero: I can easily move the print to something more verbose. The problem is that we will lose the option of reusing the same snippet of code and have to create one every time we need to use it. The message happens just after we display the error: http://18.206.154.49/job/Ddev__gazebo_ros_pkgs__ubuntu_bionic_amd64/13/parsed_console/, I think it's pretty easy to find where it comes from. Anyway I'm not strongly against changing it.

j-rivero:
> @j-rivero Can you clarify what the latest status on these are?

I parsed all the errors and categorized them. Here https://gist.github.com/j-rivero/73294c471c61ff46dba635c3d2020110#gistcomment-3063972:
 * _Upstream action probably fixes the ABI checker_ means that I think that upstream code can fix the problem somehow. We could also add some per-package/per-repo configurations to the abi-checker to make it work (that would be harder).
* _Symbols are not 100% the same although code is the same (probably due to compiled code with different compilers or standard)_ self explanatory, I can probably make them to pass with some work.
* _Under inspection_ I'm still looking into them.

scripts/devel/build_and_install.py
nuclearsandwich: I went back and looked at a few ABI failing jobs and this string is not as forward in the UI as I expected it to be so I'm less concerned.

ros_buildfarm/templates/devel/devel_task.Dockerfile.em
nuclearsandwich: Is this a situation where it is worth moving the include into the function definition? catkin-pkg is not a heavyweight dependency per se but it wasn't needed before and is clearly creating some churn in our tests.

ros_buildfarm/templates/devel/devel_create_tasks.Dockerfile.em
```diff
@@ -77,12 +77,14 @@ cmd = \
...
     ' --env-vars ' + ' ' .join(env_vars)
+if run_abichecker:
+    cmd += " --run-abichecker"
 cmds += [
     cmd +
     ' --dockerfile-dir /tmp/docker_build_and_install',
     cmd +
     ' --dockerfile-dir /tmp/docker_build_and_test' +
-    ' --testing',
+    ' --testing'
```
nuclearsandwich: If there's no need to remove the trailing comma may as well just keep it.

scripts/devel/build_and_install.py
nuclearsandwich: Looks like 2ab7c8d was a different empty line?

scripts/devel/build_and_install.py
```diff
@@ -77,7 +117,21 @@ def main(argv=sys.argv[1:]):
         if args.clean_after:
             clean_workspace(args.workspace_root)
 
-    return rc
+    if not args.run_abichecker:
```
nuclearsandwich: If the ABI checker runs then the possibly nonzero return code from the build is discarded.
Do we need to check that before continuing?

scripts/devel/build_and_install.py
```diff
@@ -77,7 +117,21 @@ def main(argv=sys.argv[1:]):
...
+            [args.workspace_root],
+            args.ros_version,
+            env)
+    # Never fail a build because of abi errors but make them
+    # unstable by printing MAKE_BUILD_UNSTABLE. Jenkins will
+    # use a`plugin to make it
+    if rc != 0:
+        print('MAKE_BUILD_UNSTABLE')
+
+    return 0
```
nuclearsandwich: Are we intentionally suppressing the ABI checker return code by not returning `rc` here?
Should we return the ABI checker code, or even perhaps the build command code?

scripts/devel/build_and_install.py
j-rivero: I think it was right. Note that code has changed a lot since that comment. Anyway I don't think that we have the problem anymore.

scripts/devel/build_and_install.py
j-rivero: > Are we intentionally suppressing the ABI checker return code by not returning `rc` here?

Yes. The approach in this pull request is that abi-checker can return unstable but never fails the build. 

>  Should we return the ABI checker code, or even perhaps the build command code?

My impression is that if we do it the build will fail with `build_and_install.py` returning non zero code.

nuclearsandwich: In that case, should the ABI checker run if the build command rc != 0?
Because right now it appears to run without checking the build command rc at all.

scripts/devel/build_and_install.py
j-rivero: I think you are right. f0161fe7bc4b4535e1844dfe1910c7048303af59

ros_buildfarm/templates/devel/devel_create_tasks.Dockerfile.em
j-rivero: ouch, reverted 5e6f456e4c5caa0581de56d51cba3283c8cdc696

ros_buildfarm/templates/devel/devel_task.Dockerfile.em
j-rivero: Not a bad idea. See 6e92b5b054db132ec0be3cb2f058415b38b0bf8a , 43528c579d6ee045adcc3b00cc25a45e5513c052 and 955ba5bb1ca93b013e32b42edca6480b847ec404

scripts/devel/build_and_install.py
j-rivero: that was indeed a bug, good catch f0161fe7bc4b4535e1844dfe1910c7048303af59

.travis.yml
```diff
@@ -197,6 +197,8 @@ matrix:
         - docker
       env: JOB_TYPE=external_prerelease
       before_script:
+        - sudo sh -c 'echo "deb http://packages.ros.org/ros/ubuntu $(lsb_release -sc) main" > /etc/apt/sources.list.d/ros.list'
```
nuclearsandwich: Are these travis changes still required for all these tests after moving the import into the ABI checker-specific method?

scripts/devel/build_and_install.py
```diff
@@ -16,16 +16,53 @@
...
 from ros_buildfarm.workspace import clean_workspace
 from ros_buildfarm.workspace import ensure_workspace_exists
 
 
+def call_abi_checker(workspace_root, ros_version, env):
+    # import the module only if the function is being called to reduce the
+    # number of mandatory dependencies
+    from catkin_pkg.packages import find_packages
+
+    # TODO: pkgs detection, code based on create_devel_task_generator.py
```
dirk-thomas: If this is already worth a TODO why not deduplicate the logic?

j-rivero: not really. Waiting travis to pass through the changes for this (and for fixing the CI) before push them here.

j-rivero: Refactored here 9d8b0c5

.travis.yml
j-rivero: you are totally right, I think they are not needed anymore, removed here bac1c75

j-rivero:
Not sure what happened with Travis but seems like it is not running on my last changes. [I can promise that they pass ](https://travis-ci.org/j-rivero/ros_buildfarm/builds/624377266?utm_source=github_status&utm_medium=notification):) .

After/before the merge we need to install the Jenkins plugins:
 *  [HTML publisher](https://wiki.jenkins.io/display/JENKINS/HTML+Publisher+Plugin) 
 * [Console parser](https://wiki.jenkins.io/display/JENKINS/Console+Parser+Plugin)

scripts/devel/build_and_install.py
dirk-thomas: I refactored the new function a bit in c8673893c435541640ba7f36324d1a394b899a2c.

PR #683, opened 2019-10-24
update plugin versions

dirk-thomas:
After noticing several plugins are using newer versions on the deployed Jenkins instances.

@nuclearsandwich FYI please make sure to keep the plugin versions on both Jenkins instance (build.ros.org and build.ros2.org) in sync and when you update them also update the information in this repo. Otherwise the nightly reconfigure jobs spend an unnecessary time reconfiguring every job.


nuclearsandwich:
> @nuclearsandwich FYI please make sure to keep the plugin versions on both Jenkins instance (build.ros.org and build.ros2.org) in sync and when you update them also update the information in this repo

Acknowledged. Updating the ros_buildfarm scripts has been sitting on my todo queue and keeps getting pre-empted. Next time it will just be part of the Jenkins upgrade process. I'll open another PR when the drift between both systems is reconciled.

PR #685, opened 2019-11-12
Dev jobs publish test reports to doc server

max-krichenbauer:
This PR adds a feature to dev jobs to upload output files of automated test to the doc server so they can be accessed easily at the same place with all the information regarding the package.

Currently, the wiki will show a "CI badge" informing people of passed or failed recent CI tests, but the details of each test require visiting the build farm Jenkins master.
In the future, we suggest that the ROS Wiki or index.ros.org hosts the most recent test results for easy access.
The update to the Wiki to display links to the test reports is under review here:
https://github.com/droter/roswiki/pull/1

This PR is based on  previous similar PR to both the Wiki and the build farm that introduced the CI badge. ( https://github.com/ros-infrastructure/roswiki/pull/246 and https://github.com/ros-infrastructure/ros_buildfarm/pull/541 )


dirk-thomas:
The current upload seems to not remove old existing test results? So I would assume they will accumulate over time and the wiki would show a mixture of current and outdated information.

dirk-thomas:
If there is no good reason to choose a different term I would suggest to `test_results` throughout the patch - instead of `test_reports`.

dirk-thomas:
Also can you give an estimate how much additional data we are roughly talking about to store on docs.ros.org?

max-krichenbauer:
@dirk-thomas Thank you for your review!

> The current upload seems to not remove old existing test results? So I would assume they will accumulate over time and the wiki would show a mixture of current and outdated information.

Good point! I added in removing the remote data via the `cleanRemote` option of `publisher_publish-over-ssh`.

> If there is no good reason to choose a different term I would suggest to test_results throughout the patch - instead of test_reports.

Got it. Changed it to "test results".

> Also can you give an estimate how much additional data we are roughly talking about to store on docs.ros.org?

Sorry, I don't really have access to that data.

ros_buildfarm/templates/snippet/publisher_publish-over-ssh.xml.em
```diff
@@ -13,7 +13,7 @@
                 <removePrefix>@remove_prefix</removePrefix>
                 <remoteDirectorySDF>false</remoteDirectorySDF>
                 <flatten>false</flatten>
-                <cleanRemote>false</cleanRemote>
+                <cleanRemote>@clean_remote</cleanRemote>
```
dirk-thomas: To avoid having to explicitly pass `clean_remote='false'` in all cases where this snippet is used you can check if the snippet was invoked with the argument and fall back to `false`:

```
@(vars().get('clean_remote', 'false'))
```

That will also ensure to not break users having custom templates using the snippet without the new argument.

max-krichenbauer: Thank you! That's a really good point.
I changed it accordingly.

ros_buildfarm/templates/devel/devel_job.xml.em
```diff
@@ -356,6 +356,18 @@ if pull_request:
...
+@[if (not pull_request) and publish_test_results]@
+@(SNIPPET(
+    'publisher_publish-over-ssh',
+    config_name='docs',
+    remote_directory='%s/devel_jobs/%s/test_results' % (rosdistro_name, source_repo_spec.name),
+    source_files=[
+        'ws/test_results/**'
+    ]
+    remove_prefix='ws/test_results',
+    clean_remote='true',
```
dirk-thomas: Nitpick: I would suggest passing a Python boolean here `True` and updating the template accordingly.

max-krichenbauer: Thanks! Good point!
I changed it accordingly.

dirk-thomas:
The patch itself looks fine to me. Thanks for iterating.

The question now seems to be if we still want these test result files to be copied from the Jenkins server (see related discussion on https://github.com/droter/roswiki/pull/1#issuecomment-565473665)? It seems that publishing the html files from HAROS might already satisfy the current use case.

max-krichenbauer:
@dirk-thomas Thank you for the review and sorry for the late reply.

Personally, I don't understand enough about the ROS infrastructure to argue for one option over another, but it seems like both options (copying test results to the doc server or showing html-based test results on Jenkins) are both being somewhat opposed: https://github.com/ros-infrastructure/ros_buildfarm/issues/693

Could we merge the two discussions and find some consensus?

ros-discourse:

https://discourse.ros.org/t/next-ros-quality-assurance-working-group-february-2020/12571/1

kscottz:
The QAWG asked if I could help get this merged. I don't know enough about the build farm to safely merge this, but I would like to learn. Anyone want to walk me through the review and testing?

dirk-thomas:
@kscottz Please read through the above conversation.

The pending question is:

> The question now seems to be if we still want these test result files to be copied from the Jenkins server (see related discussion on droter/roswiki#1 (comment))? It seems that publishing the html files from HAROS might already satisfy the current use case.

ros-discourse:

https://discourse.ros.org/t/next-ros-quality-assurance-working-group-march-2020/12914/1

tfoote:
Overall there's two competing resources here. The first element is storage space. We have lots of space on docs.ros.org but in general content that gets added there stays there. On the other side jenkins can also host the data and make it available. But it generates so much data that all jobs have retention policies for cleaning up older builds after some period or number of cycles.

To that end we cannot make a decision on this without knowing the actual data volumes we're talking about. If it's low enough that we can afford to persist the integrated data for all builds ad-infanitum uploading to docs.ros.org with full history is reasonable. However that seems unlikely. For the documentation we only keep the latest version of content for each package (per rosdistro) so the data volume doesn't grow over time. 

If we want to keep some level of history it's probably better to keep it on jenkins which has mechanisms for setting retention policies. The main drawback to that is that if this becomes something that's moderately high volume of consumption it could relatively easily overwhelm the jenkins instance with load serving the content. It's already by far our biggest server for hosting and is the one that we have to pay the most attention to to keep it running.

This is why we boiled it down to needing to know what the data looks like and how much volume of access we're expecting to experience. We can agree that hosting these files is a good idea. But until we can understand what the costs will be (storage space, and server load) it's impossible for us to evaluate the tradeoffs for this request and the other side of it here: https://github.com/ros-infrastructure/ros_buildfarm/issues/693

dirk-thomas:
And at the moment publishing the results to another server yields no obvious benefit (beside bypassing the Jenkins retention limits).

kscottz:
In my mind having the build results for all builds for all of time is a "nice to have" but not really necessary for testing to provide utility.  A simpler and incremental improvement is to have just last N build test results (most of the time you're looking at the delta, and the results should nominally be 100% passing). Are Jenkins retention policies per-artifact or per-build? If the latter something between the previous 3-7 previous builds seems sufficient to me. I take it there is no "test" build farm where you could test this and see how it performs. 

It sounds like the action item is to pick a dozen or so packages and run the test suite over them and then characterize the output so we can upper bound the amount of retained data.

tfoote:
You can set a max number of builds or a max time. If either are surpassed it will get pruned. I believe there's typically an exception for one build results in the time based threshold.

dirk-thomas:
@max-krichenbauer @kscottz What is the status on this?

I am inclined to close the ticket since there is no clear rational why the artifacts should be uploaded / how the uploaded artifacts would be used.

max-krichenbauer:
Sorry, I'm no longer participating in the QA working group.
From my point of view the issue can be closed.

PR #694, opened 2019-12-18
Skip unreleased packages from ABI checking

j-rivero:
The current implementation of abi-checking is getting all packages present in the catkin workspace under analysis to inject them into the auto-abi call. The problem is that packages marked "not to be released" are not filtered so the python script is trying to get packages from them and failing.

**Tested:**
 * Failing [with current code](http://18.206.154.49/view/Mdev/job/Mdev__ros_comm__ubuntu_bionic_amd64/3/consoleFull#console-section-13)
 * Fixed [with this PR](http://18.206.154.49/view/Mdev/job/Mdev__ros_comm__ubuntu_bionic_amd64/9/consoleFull#console-section-13)

### DIFF
--- a/ros_buildfarm/templates/devel/devel_task.Dockerfile.em
+++ b/ros_buildfarm/templates/devel/devel_task.Dockerfile.em
@@ -81,7 +81,7 @@ RUN python3 -u /tmp/wrapper_scripts/apt.py update-install-clean -q -y ccache
 ))@
 
 @[if run_abichecker]@
-RUN python3 -u /tmp/wrapper_scripts/apt.py update-install-clean -q -y python3 abi-compliance-checker python3-catkin-pkg-modules
+RUN python3 -u /tmp/wrapper_scripts/apt.py update-install-clean -q -y python3-pip abi-compliance-checker python3-catkin-pkg-modules
 RUN pip3 install -U auto_abi_checker
 @[end if]@
 
--- a/ros_buildfarm/templates/devel/devel_task.Dockerfile.em
+++ b/ros_buildfarm/templates/devel/devel_task.Dockerfile.em
@@ -81,7 +81,7 @@ RUN python3 -u /tmp/wrapper_scripts/apt.py update-install-clean -q -y ccache
 ))@
 
 @[if run_abichecker]@
-RUN python3 -u /tmp/wrapper_scripts/apt.py update-install-clean -q -y python3-pip abi-compliance-checker python3-catkin-pkg-modules
+RUN python3 -u /tmp/wrapper_scripts/apt.py update-install-clean -q -y python3 python3-pip abi-compliance-checker python3-catkin-pkg-modules
 RUN pip3 install -U auto_abi_checker
 @[end if]@
 
--- a/ros_buildfarm/templates/devel/devel_task.Dockerfile.em
+++ b/ros_buildfarm/templates/devel/devel_task.Dockerfile.em
@@ -81,7 +81,7 @@ RUN python3 -u /tmp/wrapper_scripts/apt.py update-install-clean -q -y ccache
 ))@
 
 @[if run_abichecker]@
-RUN python3 -u /tmp/wrapper_scripts/apt.py update-install-clean -q -y python3 python3-pip abi-compliance-checker python3-catkin-pkg-modules
+RUN python3 -u /tmp/wrapper_scripts/apt.py update-install-clean -q -y abi-compliance-checker python3 python3-catkin-pkg-modules python3-pip 
 RUN pip3 install -U auto_abi_checker
 @[end if]@
 
--- a/scripts/devel/build_and_install.py
+++ b/scripts/devel/build_and_install.py
@@ -60,7 +60,7 @@ def call_abi_checker(workspace_root, ros_version, env):
     assert len(workspace_root) == 1, 'auto-abi tool needs the implementation of multiple local-dir'
     # ROS_DISTRO is set in the env object
     cmd = ['auto-abi.py ' +
-           '--orig-type ros-pkg --orig ' + ",".join(pkg_names) + ' ' +
+           '--orig-type ros-pkg --orig ' + ",".join(pkg_names_released) + ' ' +
            '--new-type ros-ws --new ' + os.path.join(workspace_root[0], 'install_isolated') + ' ' +
            '--report-dir ' + workspace_root[0] + ' ' +
            '--no-fail-if-empty ' +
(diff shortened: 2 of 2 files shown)


ros_buildfarm/templates/devel/devel_task.Dockerfile.em
```diff
@@ -81,7 +81,7 @@ RUN python3 -u /tmp/wrapper_scripts/apt.py update-install-clean -q -y ccache
 ))@
 
 @[if run_abichecker]@
-RUN python3 -u /tmp/wrapper_scripts/apt.py update-install-clean -q -y python3 abi-compliance-checker python3-catkin-pkg-modules
+RUN python3 -u /tmp/wrapper_scripts/apt.py update-install-clean -q -y python3-pip abi-compliance-checker python3-catkin-pkg-modules
```
j-rivero: oh, the pip addition was needed when I was testing the failure on ROS1 packages

dirk-thomas: `python3` is still needed, no? So this should just add `python3-pip` to the list and not rely on it pulling in `python3` implicitly.

scripts/devel/build_and_install.py
```diff
@@ -40,10 +42,25 @@ def call_abi_checker(workspace_root, ros_version, env):
...
+    pkg_names_released = []
+    for pkg_name in pkg_names:
+        try:
+            if distro_file.get_release_package_xml(pkg_name):
+                pkg_names_released.append(pkg_name)
+        except KeyError:
+            # skip unreleased packages, nothing to do with ABI checking
+            pass
+    if not pkg_names_released:
+        print("No released packages found in the workspace. Skipping abi-checker run")
```
dirk-thomas: Nitpick: single quotes.

scripts/devel/build_and_install.py
```diff
@@ -40,10 +42,25 @@ def call_abi_checker(workspace_root, ros_version, env):
     pkg_names = [pkg.name for pkg in pkgs.values()]
     assert pkg_names, 'No packages found in the workspace'
 
+    # Filter packages in source space that has been released
+    index_file = rosdistro.get_index(rosdistro.get_index_url())
+    distro_file = rosdistro.get_cached_distribution(index_file, env['ROS_DISTRO'])
+    pkg_names_released = []
+    for pkg_name in pkg_names:
+        try:
+            if distro_file.get_release_package_xml(pkg_name):
```
dirk-thomas: Why does it need to fetch the `package.xml` content?

ros_buildfarm/templates/devel/devel_task.Dockerfile.em
j-rivero: agree. Changed in b181de8

scripts/devel/build_and_install.py
j-rivero: d5df983

scripts/devel/build_and_install.py
j-rivero: It does not need to fetch the package xml. It does need to check if a given package name has been released. What would be the best way of doing this?

ros_buildfarm/templates/devel/devel_task.Dockerfile.em
```diff
@@ -81,7 +81,7 @@ RUN python3 -u /tmp/wrapper_scripts/apt.py update-install-clean -q -y ccache
 ))@
 
 @[if run_abichecker]@
-RUN python3 -u /tmp/wrapper_scripts/apt.py update-install-clean -q -y python3 abi-compliance-checker python3-catkin-pkg-modules
+RUN python3 -u /tmp/wrapper_scripts/apt.py update-install-clean -q -y python3 python3-pip abi-compliance-checker python3-catkin-pkg-modules
```
dirk-thomas: Please insert in alphabetical order.

scripts/devel/build_and_install.py
```diff
@@ -40,10 +42,25 @@ def call_abi_checker(workspace_root, ros_version, env):
     pkg_names = [pkg.name for pkg in pkgs.values()]
     assert pkg_names, 'No packages found in the workspace'
 
+    # Filter packages in source space that has been released
+    index_file = rosdistro.get_index(rosdistro.get_index_url())
+    distro_file = rosdistro.get_cached_distribution(index_file, env['ROS_DISTRO'])
```
dirk-thomas: The function doesn't return a `DistributionFile` but a `Distribution`. Therefore I suggest to rename the variable to just `dist`.

scripts/devel/build_and_install.py
dirk-thomas: A subset of what the function is doing internally (and you don't need a full `Distribution` for that anymore but only a `DistributionFile` so the comment above might not apply anymore): https://github.com/ros-infrastructure/rosdistro/blob/1bd3c008404d54df0299361afe5a8a4ebf3e3da7/src/rosdistro/distribution.py#L66-L73

scripts/devel/build_and_install.py
j-rivero: be06268

ros_buildfarm/templates/devel/devel_task.Dockerfile.em
j-rivero: 9f96596

scripts/devel/build_and_install.py
j-rivero: Thanks for the explanation. I simplified the logic in 9f96596 to just get the distribution file and try to access to the package using `dist_file.release_packages[pkg_name]`. Good enough?

scripts/devel/build_and_install.py
```diff
@@ -40,10 +42,27 @@ def call_abi_checker(workspace_root, ros_version, env):
...
+
+    pkg_names_released = []
+    for pkg_name in pkg_names:
+        try:
+            # Check in released packages for pkg_name (KeyError if unreleased)
+            dist_file.release_packages[pkg_name]
+            pkg_names_released.append(pkg_name)
+        except KeyError:
+            # skip unreleased packages, nothing to do with ABI checking
+            pass
```
dirk-thomas: Please rewrite the `try` / `except` logic to be more condensed like:

```
if pkg_name in dist_file.release_packages:
    pkg_names_released.append(pkg_name)
```

dirk-thomas: Forget my last comment. The whole loop can be condensed to:

```
pkg_names_released = [pkg_name for pkg_name in pkg_names if pkg_name in dist_file.release_packages]
```

j-rivero: I like it e4eee53

dirk-thomas:
Please address the reported CI failure related to the linter.

Issue #695, opened 2019-12-19
local Prerelease/release jobs fail if ~/.ccache doesnt exist on the host before running the scripts


mikaelarguedas:
Running prerelease job on Bionic (and Xenial) fails if the directory `~/.ccache` doesnt exist on the host machine

When trying to mount the volume the folder is created as root. But the docker container is using a non-root user causing this error to appear.
```
ccache: error: Failed to create directory /home/buildfarm/.ccache/tmp: Permission denied
```

<details><summary>full error message</summary>

```
==> cmake /tmp/ws/src/roscpp_core/cpp_common -DCATKIN_DEVEL_PREFIX=/tmp/ws/devel_isolated/cpp_common -DCMAKE_INSTALL_PREFIX=/tmp/ws/install_isolated -DBUILD_TESTING=0 -DCATKIN_SKIP_TESTING=1 -G Unix Makefiles in '/tmp/ws/build_isolated/cpp_common'
-- The C compiler identification is GNU 7.4.0
-- The CXX compiler identification is GNU 7.4.0
-- Check for working C compiler: /usr/lib/ccache/cc
-- Check for working C compiler: /usr/lib/ccache/cc -- broken
CMake Error at /usr/share/cmake-3.10/Modules/CMakeTestCCompiler.cmake:52 (message):
  The C compiler

    "/usr/lib/ccache/cc"

  is not able to compile a simple test program.

  It fails with the following output:

    Change Dir: /tmp/ws/build_isolated/cpp_common/CMakeFiles/CMakeTmp
    
    Run Build Command:"/usr/bin/make" "cmTC_4192b/fast"
    /usr/bin/make -f CMakeFiles/cmTC_4192b.dir/build.make CMakeFiles/cmTC_4192b.dir/build
    make[1]: Entering directory '/tmp/ws/build_isolated/cpp_common/CMakeFiles/CMakeTmp'
    Building C object CMakeFiles/cmTC_4192b.dir/testCCompiler.c.o
    /usr/lib/ccache/cc    -o CMakeFiles/cmTC_4192b.dir/testCCompiler.c.o   -c /tmp/ws/build_isolated/cpp_common/CMakeFiles/CMakeTmp/testCCompiler.c
    ccache: error: Failed to create directory /home/buildfarm/.ccache/tmp: Permission denied
[...]

PR #696, opened 2019-12-19
create '/$HOME/.ccache' as a user before mounting it

mikaelarguedas:
Fixes https://github.com/ros-infrastructure/ros_buildfarm/issues/695 locally
Also replaced `~/.ccache` with `$HOME/.ccache` to match the other scripts of this package

Hasn't been tested on a buildfarm


dirk-thomas:
Thanks for the patch.

PR #700, opened 2020-01-06
Add buster to os code name mapping

sloretz:
`DB` for Debian Buster since `B` is taken by Ubuntu Bionic.

### DIFF
--- a/ros_buildfarm/common.py
+++ b/ros_buildfarm/common.py
@@ -280,7 +280,7 @@ def get_short_os_code_name(os_code_name):
     os_code_name_mappings = {
         'artful': 'A',
         'bionic': 'B',
-        'buster': 'DB',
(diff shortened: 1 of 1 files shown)


ros_buildfarm/common.py
```diff
@@ -280,6 +280,7 @@ def get_short_os_code_name(os_code_name):
     os_code_name_mappings = {
         'artful': 'A',
         'bionic': 'B',
+        'buster': 'DB',
```
dirk-thomas: This should be just `B`. Similar to Saucy / Stretch the code name mapping alone isn't unique but it always appears the a character identifying for example Debian  / D.

sloretz: Just B in f22b00f

PR #702, opened 2020-01-14
Added num_build option to jenkin plot plugin

ahcorde:
According with the [Jenkins plot plugin documentation](https://wiki.jenkins.io/display/JENKINS/Plot+Plugin) `numBuild` value must be not zero, I set up the default value to to 10 which means we will be able to visualize plots of the 10 days before.

### DIFF
--- a/doc/configuration_options.rst
+++ b/doc/configuration_options.rst
@@ -474,7 +474,7 @@ The following options are valid in version ``1`` (beside the generic options):
   * ``y_axis_maximum``: Maximum y-axis value.
   * ``num_builds``: Number of builds back to show on this plot.
     An empty string or not defined mean all builds.
-    Must not be "0", the default is 10.
+    Must not be "0", the default value is an empty string.
   * ``data_series``: a list of data series definitions comprised of:
 
     * ``data_file``: a path pattern relative to the workspace root to a file
--- a/doc/configuration_options.rst
+++ b/doc/configuration_options.rst
@@ -472,8 +472,9 @@ The following options are valid in version ``1`` (beside the generic options):
     implicit zero value from the y-axis.
   * ``y_axis_minimum``: Minimum y-axis value.
   * ``y_axis_maximum``: Maximum y-axis value.
-  * ``num_build``: Number of builds back to show on this plot.
-    An empty string means all builds. Must not be "0".
+  * ``num_builds``: Number of builds back to show on this plot.
+    An empty string or not defined mean all builds.
+    Must not be "0", the default is 10.
   * ``data_series``: a list of data series definitions comprised of:
 
     * ``data_file``: a path pattern relative to the workspace root to a file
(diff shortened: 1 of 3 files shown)


ros_buildfarm/config/plot_config.py
```diff
@@ -45,6 +45,7 @@ def __init__(self, name, data):
 
         self.y_axis_minimum = data.get('y_axis_minimum', None)
         self.y_axis_maximum = data.get('y_axis_maximum', None)
+        self.num_build = int(data.get('num_build', 10))
```
dirk-thomas: This logic doesn't work for an empty string input.

doc/configuration_options.rst
```diff
@@ -472,6 +472,8 @@ The following options are valid in version ``1`` (beside the generic options):
     implicit zero value from the y-axis.
   * ``y_axis_minimum``: Minimum y-axis value.
   * ``y_axis_maximum``: Maximum y-axis value.
+  * ``num_build``: Number of builds back to show on this plot. An empty string
```
dirk-thomas: Nitpick: start each sentence on a new line.

doc/configuration_options.rst
```diff
@@ -472,6 +472,8 @@ The following options are valid in version ``1`` (beside the generic options):
     implicit zero value from the y-axis.
   * ``y_axis_minimum``: Minimum y-axis value.
   * ``y_axis_maximum``: Maximum y-axis value.
+  * ``num_build``: Number of builds back to show on this plot.
+  An empty string means all builds. Must not be "0".
```
dirk-thomas: Needs to be indented two spaces more. Second sentence still needs to be wrapped. Document the default value.

ros_buildfarm/config/plot_config.py
```diff
@@ -46,6 +46,10 @@ def __init__(self, name, data):
         self.y_axis_minimum = data.get('y_axis_minimum', None)
         self.y_axis_maximum = data.get('y_axis_maximum', None)
 
+        self.num_build = 10
+        if 'num_build' in data:
+            self.num_build = data['num_build']
```
dirk-thomas: It might be good to check the type of the value to ensure it will work as a plugin config option.

ahcorde: I included an `assert` maybe is it better to cast the value to `int`?

doc/configuration_options.rst
dirk-thomas: The second part of my comment still applies.

ros_buildfarm/config/plot_config.py
dirk-thomas: The documentation states that you can pass an empty string which means to keep all builds. That is still not reflected here.

doc/configuration_options.rst
```diff
@@ -472,8 +472,9 @@ The following options are valid in version ``1`` (beside the generic options):
     implicit zero value from the y-axis.
   * ``y_axis_minimum``: Minimum y-axis value.
   * ``y_axis_maximum``: Maximum y-axis value.
-  * ``num_build``: Number of builds back to show on this plot.
-    An empty string means all builds. Must not be "0".
+  * ``num_builds``: Number of builds back to show on this plot.
+    An empty string or not defined mean all builds.
+    Must not be "0", the default is 10.
```
dirk-thomas: > not defined mean all builds
> the default is 10

These two directly contradict each other.

PR #723, opened 2020-01-28
use GoogleTest / JUnit test result parser based on file pattern in ROS 2

dirk-thomas:
For ROS 1 the logic keeps using only `GoogleTest` for all XML results.

The change matches the logic of ci.ros2.org except that there is no CTest type configured since the build directory is wiped before test results are collected. This will make Jenkins render better error messages for failed test: e.g. [bad](http://build.ros2.org/view/Eci/job/Eci__nightly-fastrtps_ubuntu_bionic_amd64/216/testReport/ros2launch.test/test_flake8/test_flake8/) vs. [good](https://ci.ros2.org/view/nightly/job/nightly_linux_debug/1438/testReport/rqt_reconfigure.test/test_flake8/test_flake8/).

* http://build.ros2.org/job/Ddev_reconfigure-jobs/494/ used this branch for a dry run and successfully reconfigured `devel` jobs (with no changes) while their template wasn't update to the modified arguments of the `publisher_junit` snippet: https://github.com/ros-infrastructure/ros_buildfarm/commit/9139aa9068b3f2e991d69766daa238041e0b7ad9#diff-5cf979d1d9687b7cad0d95e630cc87c6R269. This ensured the backward compatibility of the changed snippet arguments works.

* http://build.ros.org/job/Nci_reconfigure-jobs/186/ used this branch for a dry run and successfully reconfigured `ci` jobs (with no changes): https://github.com/ros-infrastructure/ros_buildfarm/commit/9139aa9068b3f2e991d69766daa238041e0b7ad9#diff-7b020c516b04814ffc3fc28ed8fac0deR597. This ensures ROS 1 stays untouched.

* http://build.ros2.org/job/Fci_reconfigure-jobs/5/ (dry run) reconfigured ROS 2 CI jobs to use the more specific xunit types bases on the various patterns.

### DIFF
--- a/ros_buildfarm/ci_job.py
+++ b/ros_buildfarm/ci_job.py
@@ -23,6 +23,7 @@ from ros_buildfarm.common import get_default_node_label
 from ros_buildfarm.common import get_node_label
 from ros_buildfarm.common \
     import get_repositories_and_script_generating_key_files
+from ros_buildfarm.common import get_xunit_publisher_types_and_patterns
 from ros_buildfarm.common import JobValidationError
 from ros_buildfarm.common import write_groovy_script_and_configs
 from ros_buildfarm.config import get_ci_build_files
@@ -293,6 +294,9 @@ def _get_ci_job_config(
 
         'show_images': build_file.show_images,
         'show_plots': build_file.show_plots,
+
+        'xunit_publisher_types': get_xunit_publisher_types_and_patterns(
+            ros_version),
     }
     job_config = expand_template(template_name, job_data)
     return job_config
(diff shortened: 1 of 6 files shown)


tfoote:
The CI is having an issue with "names" in the template. Is this the empy bug you found?

dirk-thomas:
> The CI is having an issue with "names" in the template. Is this the empy bug you found?

Correct. I already reported it upstream. But even if upstream fixed the problem and released a new version I think we have to work around the problem by not using the problematic logic in `empy`. (Not yet sure how...)

PR #737, opened 2020-02-04
Add per-project authorization for CI builds

nuclearsandwich:
Adds the ability to specify per-job (per-project in Jenkins parlance) XML configuration for access control.

I've added config field `project_authorization_xml` to the base BuildFile class but only wired it through for CI jobs where we currently want it.

In order for this configuration to be honored Jenkins Authorization must be configured to use the Project-based Matrix Authorization Strategy whereas the current buildfarm deployment default is Matrix-based security without per-project settings.

### DIFF
--- a/doc/configuration_options.rst
+++ b/doc/configuration_options.rst
@@ -80,6 +80,7 @@ A set of options which can be used in any build file.
   ``<hudson.security.AuthorizationMatrixProperty>`` in job builds.
   This property is definied for all build files but as of `#737`_ is only
   implemented for **CI jobs**.
+
   .. _#737: https://github.com/ros-infrastructure/ros_buildfarm/pull/737
 
 * ``repositories``: additional repositories to fetch packages from.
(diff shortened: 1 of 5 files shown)


dirk-thomas:
Please add the new config option to the docs.

nuclearsandwich:
> Please add the new config option to the docs.

Added the documentation. I could see the logic in moving the block for this option into the CI build files section but I started with it in the general options since it's defined in the base build file.

doc/configuration_options.rst
```diff
@@ -76,6 +76,13 @@ A set of options which can be used in any build file.
   Note that yaml will turn bare words like ``yes`` into boolean values so it
   is recommended to quote values to avoid interpretation.
 
+* ``project_authorization_xml``: an XML blob which will be nested within a
+  ``<hudson.security.AuthorizationMatrixProperty>`` in job builds.
+  This property is definied for all build files but as of `#737`_ is only
```
gavanderhoorn:
```suggestion
  This property is defined for all build files but as of `#737`_ is only
```

PR #742, opened 2020-03-03
Update Jenkins plugin script-security: from 1.68 to 1.70 

j-rivero:
The version 1.69 implements more methods to the whitelist and improvement in the log (that should not affect in anyway the current config). The version 1.70 blocks annotation in Groovy scripts due to security reasons. I did not find any use of this feature in the repo.

The testing buildfarm works just fine after updating this script.


j-rivero:
> The version 1.70 blocks annotation in Groovy scripts due to security reasons. I did not find any use of this feature in the repo.

Uh not true. "Annotation Grab cannot be used in the sandbox" . `Found in ros_buildfarm/templates/devel/devel_job.xml.em` . I need to fix it.

j-rivero:
I've replaced the `Grab` by using the snakeyaml provided by [Pipeline Utility Steps](https://plugins.jenkins.io/pipeline-utility-steps) in 35ba81c. We need to install that plugin but I checked that the yaml file generated looks good.

dirk-thomas:
Please use squash merge in the future.

PR #743, opened 2020-03-03
Replace warnings plugin by warnings next generation

j-rivero:
The warnings plugin is deprecated. Replace it here by the [Warning next generation](https://jenkins.io/doc/pipeline/steps/warnings-ng/).

Tested it with forced warnings: http://54.162.84.148/job/Mdev__ros_comm__ubuntu_bionic_amd64/2

### DIFF
--- a/ros_buildfarm/templates/snippet/publisher_warnings.xml.em
+++ b/ros_buildfarm/templates/snippet/publisher_warnings.xml.em
@@ -26,7 +26,11 @@
   <minimumSeverity plugin="analysis-model-api@@7.0.2">
     <name>LOW</name>
   </minimumSeverity>
-  <filters/>
+  <filters>
+    <io.jenkins.plugins.analysis.core.filter.ExcludeMessage>
+      <pattern>^Manually-specified variables were not used by the project:$</pattern>
+    </io.jenkins.plugins.analysis.core.filter.ExcludeMessage>
+  </filters>
   <isEnabledForFailure>false</isEnabledForFailure>
   <isAggregatingResults>false</isAggregatingResults>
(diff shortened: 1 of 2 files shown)


ros_buildfarm/templates/snippet/publisher_warnings.xml.em
```diff
@@ -1,49 +1,44 @@
...
-      <useStableBuildAsReference>false</useStableBuildAsReference>
-      <useDeltaValues>false</useDeltaValues>
-      <thresholds plugin="analysis-core@@1.96">
+<io.jenkins.plugins.analysis.core.steps.IssuesRecorder plugin="warnings-ng@@7.3.0">
+  <analysisTools>
+    <!--
+         catkin inject CMake variables that are not
+         always used in build. Be careful if planning
+         to add the CMake parser
+    -->
```
dirk-thomas: The CMake parser was used before. Why not anymore?

j-rivero: Uh. When I saw the warning coming from `Manually-specified variables were not used by the project: BUILD_TESTING` I thought there was no CMake warning parser before. I was wrong it was enabled (thanks for catching it). 

I've backed to enable the CMake parser in f5c9293 and added an exception in cfb8fbc . The plugin detect the message `Manually-specified variables were not used by the project:` but not the specific variable so I think we can not be more precise on the exclusion. Since the buildfarm code control the CMake variables passed to the cmake invocation, I would say that is ok to go with the "manually-specified variables" exception.

dirk-thomas:
Please use squash merge in the future.

PR #751, opened 2020-03-06
Install rosdoc_lite deps based on python version

### LINKED ISSUE #706: Use rosdoc_lite with Python 3
## Problem

It looks like doc jobs are invoking `rosdoc_lite` with Python 2. At least on Ubuntu Focal where Python 2 versions of its dependencies aren't available, `rosdoc_lite` needs to be invoked with Python 3.

Python 2 dependencies are manually installed:
https://github.com/ros-infrastructure/ros_buildfarm/blob/65cb136151d44d62c8050fed705433808338309a/scripts/doc/create_doc_task_generator.py#L522-L529

## Possible solution ?

`rosdoc_lite` appears to be installed from source; however, [it is released into the distro](https://github.com/ros/rosdistro/blob/a5e54b13046dd22d1569fb04a323d860571cbe4d/melodic/distribution.yaml#L8290).
https://github.com/ros-infrastructure/ros_buildfarm/blob/65cb136151d44d62c8050fed705433808338309a/ros_buildfarm/templates/doc/doc_job.xml.em#L89

`catkin` and `genmsg` appear to be installed via debian packages. Can the doc jobs install `rosdoc_lite` via the debian package too?

https://github.com/ros-infrastructure/ros_buildfarm/blob/65cb136151d44d62c8050fed705433808338309a/scripts/doc/create_doc_task_generator.py#L531-L533

If that's possible, it would still require [rosdoc_lite to use conditional dependencies](https://github.com/ros-infrastructure/rosdoc_lite/blob/ec884e5f17f962698c653453ece65ffed03fb21d/package.xml#L24) and be released into ROS Noetic.

sloretz:
Resolves #706
Requires https://github.com/ros-infrastructure/ros_buildfarm_config/pull/163

This installs the Python apt dependencies based on the distribution's ROS_PYTHON_VERSION. It looks like the least effort way to get Noetic doc jobs passing.

### DIFF
--- a/scripts/doc/create_doc_task_generator.py
+++ b/scripts/doc/create_doc_task_generator.py
@@ -536,8 +536,8 @@ def main(argv=sys.argv[1:]):
                 'python3-sphinx',
                 'python3-yaml'])
         else:
-            print('Unknown python version', condition_context)
-            # Assume it's old, and using Python 2
+            if '2' != condition_context['ROS_PYTHON_VERSION']:
+                print('Unknown python version, using Python 2', condition_context)
             # the following are required by rosdoc_lite
             debian_pkg_names.extend([
                 'python-catkin-pkg-modules',
--- a/scripts/doc/create_doc_task_generator.py
+++ b/scripts/doc/create_doc_task_generator.py
@@ -527,16 +527,7 @@ def main(argv=sys.argv[1:]):
             get_os_package_name(args.rosdistro_name, 'genmsg'),
         ]
 
-        if '2' == condition_context['ROS_PYTHON_VERSION']:
-            # the following are required by rosdoc_lite
-            debian_pkg_names.extend([
-                'python-catkin-pkg-modules',
-                'python-epydoc',
-                'python-kitchen',
-                'python-rospkg',
-                'python-sphinx',
-                'python-yaml'])
-        elif '3' == condition_context['ROS_PYTHON_VERSION']:
+        if '3' == condition_context['ROS_PYTHON_VERSION']:
             # the following are required by rosdoc_lite
             debian_pkg_names.extend([
                 'python3-catkin-pkg-modules',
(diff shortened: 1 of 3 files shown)


dirk-thomas:
> Resolves #73

That seems to be the wrong ticket?

scripts/doc/create_doc_task_generator.py
```diff
@@ -521,17 +521,32 @@ def main(argv=sys.argv[1:]):
...
+        if '3' == condition_context['ROS_PYTHON_VERSION']:
+            # the following are required by rosdoc_lite
+            debian_pkg_names.extend([
+                'python3-catkin-pkg-modules',
+                'python3-kitchen',
+                'python3-rospkg-modules',
+                'python3-sphinx',
+                'python3-yaml'])
+        else:
+            print('Unknown python version', condition_context)
```
dirk-thomas: That message isn't necessarily true. The Python version might be known - just not `3`.

sloretz: Only print if python version is not `2` in 949059331f282cd2275002be303c68f342e277fd

sloretz:
Catkin Noetic doc job passed with the current state of this PR: http://build.ros.org/view/Ndoc/job/Ndoc__catkin__ubuntu_focal_amd64/14/

scripts/doc/create_doc_task_generator.py
```diff
@@ -579,6 +594,8 @@ def main(argv=sys.argv[1:]):
                 args.distribution_repository_urls,
                 args.distribution_repository_key_files),
 
+            'environment_variables': ['ROS_PYTHON_VERSION={}'.format(condition_context['ROS_PYTHON_VERSION'])],
```
dirk-thomas: Linter warning: line too long.

sloretz: Wrapped line in 00abc36

PR #789, opened 2020-04-10
Pin pyparsing 2.4.7 for Python 3.4 and 2.7 Travis CI jobs.

hidmic:
Precisely what the title says. Let's see if Travis likes it.


dirk-thomas:
Please add comments why that specific version is being installed.

hidmic:
> Please add comments why that specific version is being installed.

PTAL!

dirk-thomas:
Nitpick: a single link (https://github.com/pyparsing/pyparsing/blob/72f2c5a67b4a26f584104b9ff63e1f272f54c5df/CHANGES#L5-L9) looks sufficient to me instead of two.

hidmic:
@dirk-thomas done

Issue #331, opened 2016-08-18
allow status page jobs to run concurrently


tfoote:
On first inspection they look concurrent safe, but testing them overnight for #327 exposed two different backtraces once they were running in parallel. 

```
# BEGIN SECTION: Run Dockerfile - status page
05:48:55 + rm -fr /home/jenkins-slave/workspace/Irel_release-status-page/debian_repo_cache
05:48:55 + rm -fr /home/jenkins-slave/workspace/Irel_release-status-page/status_page
05:48:55 + mkdir -p /home/jenkins-slave/workspace/Irel_release-status-page/debian_repo_cache
05:48:55 + mkdir -p /home/jenkins-slave/workspace/Irel_release-status-page/status_page
05:48:55 + docker run --rm --cidfile=/home/jenkins-slave/workspace/Irel_release-status-page/docker_generate_status_page/docker.cid --net=host -v /home/jenkins-slave/workspace/Irel_release-status-page/ros_buildfarm:/tmp/ros_buildfarm:ro -v /home/jenkins-slave/workspace/Irel_release-status-page/debian_repo_cache:/tmp/debian_repo_cache -v /home/jenkins-slave/workspace/Irel_release-status-page/status_page:/tmp/status_page status_page_generation
05:48:56 The build file contains the following targets:
05:48:56   - xenial source
05:48:56   - xenial armhf
05:48:59 Traceback (most recent call last):
05:48:59   File "/tmp/ros_buildfarm/scripts/status/build_release_status_page.py", line 48, in <module>
05:48:59     main()
05:48:59   File "/tmp/ros_buildfarm/scripts/status/build_release_status_page.py", line 44, in main
05:48:59     args.cache_dir, args.output_dir, copy_resources=args.copy_resources)
05:48:59   File "/tmp/ros_buildfarm/ros_buildfarm/status_page.py", line 76, in build_release_status_page
05:48:59     building_repo_url, targets, cache_dir)
05:48:59   File "/tmp/ros_buildfarm/ros_buildfarm/debian_repo.py", line 31, in get_debian_repo_data
05:48:59     debian_repository_baseurl, target, cache_dir)
05:48:59   File "/tmp/ros_buildfarm/ros_buildfarm/debian_repo.py", line 47, in get_debian_repo_index
05:48:59     fetch_gzip_url(url, cache_filename)
[...]

PR #799, opened 2020-04-25
Allow status page jobs to run on any agent.

tfoote:
The status pages jobs query the public api of the repo for status and upload to a specific non-overlapping scope of files. The uploads are managed by the ssh publisher with the credentials on the master already so they will work on any agent.

With the number of status jobs and reconfigure jobs we are actually almost at 100% duty cycle for the agent_on_master instance. Allowing these to parallelize will lower the load on the master instance. It will also allow more prompt performance of the status and reconfigure jobs.

This is a rework of #327 That was closed due to incorrect logic in https://github.com/ros-infrastructure/ros_buildfarm/issues/331#issuecomment-251174703 the sync jobs run on the repository not the master (or even agent_on_master) so single threading this does not provide any protections.

Time and resources on the agents are cheaper. If anything I might even propose removing the agent_on_master to leave it available for the reconfigure and trigger jobs. These jobs run quickly so would not starve any of the other builds. And if anything will allow the master to have less resource contention which might actually increase throughput as jobs can be dispatched more effectively without having one process dedicated to generating the status pages. 

Here's a successful build on `buildagent` http://build.ros.org/job/Krel_release-status-page/23650/console

### DIFF
--- a/ros_buildfarm/templates/status/blocked_releases_page_job.xml.em
+++ b/ros_buildfarm/templates/status/blocked_releases_page_job.xml.em
@@ -24,7 +24,7 @@
     refspec=None,
 ))@
   <scmCheckoutRetryCount>2</scmCheckoutRetryCount>
-  <assignedNode>agent_on_master</assignedNode>
+  <assignedNode>agent_on_master||buildagent</assignedNode>
   <canRoam>false</canRoam>
   <disabled>false</disabled>
   <blockBuildWhenDownstreamBuilding>false</blockBuildWhenDownstreamBuilding>
--- a/ros_buildfarm/templates/status/blocked_source_entries_page_job.xml.em
+++ b/ros_buildfarm/templates/status/blocked_source_entries_page_job.xml.em
@@ -24,7 +24,7 @@
     refspec=None,
 ))@
   <scmCheckoutRetryCount>2</scmCheckoutRetryCount>
-  <assignedNode>agent_on_master</assignedNode>
+  <assignedNode>agent_on_master||buildagent</assignedNode>
   <canRoam>false</canRoam>
   <disabled>false</disabled>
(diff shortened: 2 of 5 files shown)


tfoote:
I'm looking at this because the master was unresponsive while debugging https://github.com/ros-infrastructure/ros_buildfarm_config/pull/167

The master was giving 502 responses to configuration updates. 

It recovered but after the recovery the temporary buildagent I'm using for rosindex had gone offline: http://build.ros.org/computer/agent-eff13541/

There were also some unusual changes in executor availabilty that likely is correlated. As well as a precipitous failure rate.

Screenshot from 2020-04-24 18-34-02

nuclearsandwich:
Do any of these status jobs overlap such that parallel builds may interfere with eachother?

Are the jobs properly configured not to run two instances of the same status job at the same time?

tfoote:
> Do any of these status jobs overlap such that parallel builds may interfere with eachother?

They are all read only on the repositories. They will push one unique html file and one unique yaml file. Each class copies up the css and js into a shared place. It currently overwrites the same content every time. The file system can handle the multiple incoming copies without corrupting the files.

> Are the jobs properly configured not to run two instances of the same status job at the same time?

Yes that's on by default for non-parameterized jobs.

nuclearsandwich:
Just approved but also thought of one last question which is where do these jobs land priority-wise? I don't think they'll ever be starved out since they can always run on the agent-on-master but we should probably ensure they don't shut out other buildagent jobs when there's a low agent count.

tfoote:
They only take ~20 seconds each. With 16 of them on the farm running every 15 minutes they will take up approximately 1/3 of a single executor. Or on our standard build machines with 4 executors ~8.3% of a lone machine.

PR #800, opened 2020-04-28
make docker image names unique for status pages

tfoote:
Follow up to #799 to make the docker images unique instead of overwriting the docker image and causing a potential race condition on the same host.


dirk-thomas:
@tfoote I don't know how you merged this PR but the commit on the target branch doesn't reference the PR. Please make sure that is not the case next time.

PR #801, opened 2020-04-29
Avoid peak loads

tfoote:
This will will distribute our periodic jobs out over the same period so that we don't have as much spike loading. This will not change the frequency with which they run, but will provide a more even execution.

We historically didn't do this as we wanted to know when things would run. But as we're scaling up the peak loading makes it harder for our servers. And we're still subject to unknown caching delays with github hosting and if they all run at once they don't execute exactly when scheduled because they all get into a queue anyway. 

I also noticed that we're running the repository status every 15 minutes but the release status which is our primary tool every 20 minutes because the load was too high. Slowing down the repo status page to hourly.

### DIFF
--- a/ros_buildfarm/templates/misc/dashboard_job.xml.em
+++ b/ros_buildfarm/templates/misc/dashboard_job.xml.em
@@ -27,7 +27,7 @@
   <triggers>
 @(SNIPPET(
     'trigger_timer',
-    spec='0 */6 * * *',
+    spec='0 H/6 * * *',
 ))@
   </triggers>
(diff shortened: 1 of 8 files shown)


tfoote:
The beginning has the spiky behavior, after it calms down.
Screenshot from 2020-04-29 15-09-33

Waiting for it to go back down to idle. 
Screenshot from 2020-04-30 15-32-37

dirk-thomas:
@tfoote I don't know how you merged this PR but the commits on the target branch don't reference the PR. Please make sure that is not the case next time.

PR #813, opened 2020-06-17
only pass catkin specific CMake definition for ROS 1 jobs

dirk-thomas:
This avoid the following type of CMake warning in ROS 2 jobs:

> Manually-specified variables were not used by the project:

### DIFF
--- a/ros_buildfarm/templates/devel/devel_task.Dockerfile.em
+++ b/ros_buildfarm/templates/devel/devel_task.Dockerfile.em
@@ -115,7 +115,8 @@ if not testing:
 else:
     cmd += \
         ' /tmp/ros_buildfarm/scripts/devel/build_and_test.py' + \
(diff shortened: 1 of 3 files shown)


nuclearsandwich:
This will eliminate warnings from variables used only for ROS 1 but CI jobs may also need something like [this](https://github.com/ros2/ci/blob/8c758cf359ec5328f652f5f068654aa6101b44a7/ros2_batch_job/__main__.py#L350) since some variables like `INSTALL_EXAMPLES=OFF` and `SECURITY=ON` are only used for specific projects.

dirk-thomas:
> CI jobs may also need something like this since some variables like `INSTALL_EXAMPLES=OFF` and `SECURITY=ON` are only used for specific projects.

Good point. I will check for those in another round.

PR #814, opened 2020-06-19
clear log directory before / after a build

dirk-thomas:
Follow up of #812 

When parsing the `stdout_stderr.log` files to extract warnings it needs to be made that no log files from previous builds are still in the workspace.


dirk-thomas:
Since the current regression leads to incorrect results (e.g. http://build.ros2.org/view/Rdev/job/Rdev__rclpy__ubuntu_focal_amd64/4/) I will go ahead and merge this change without waiting for a review. I am happy to address any comments after the fact.

PR #815, opened 2020-06-23
Add a new rosdistro audit tool for release managers

tfoote:
This tool will help them audit all missing packages in the rosdistro to eliminate any missing packages before a sync. 

It will say what's failing to build on all platforms to be removed from the rosdistro. And say what's failing on specific build configs and should be blacklisted. 

Still pending is an additional capability to link into jenkins to get build status and links for easier debugging. It also could help with providing urls for ticketing the issues upstream and possibly opening blacklist and rosdistro delisting tickets automatically or at least providing a template for the manager to copy and paste.

Basic usage `./scripts/release/audit_rosdistro.py https://raw.githubusercontent.com/ros-infrastructure/ros_buildfarm_config/production/index.yaml melodic`

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,3 +15,13 @@ to configure the jobs being generated (e.g.
 If you are going to use any of the provided infrastructure please consider
 watching the [buildfarm Discourse category](https://discourse.ros.org/c/buildfarm)
 in order to receive notifications e.g. about any upcoming changes.
+
+For quick reference to run scripts:
+
+ ## Check Sync Criteria
+
+    ./scripts/release/check_sync_criteria.py https://raw.githubusercontent.com/ros-infrastructure/ros_buildfarm_config/production/index.yaml melodic default ubuntu bionic amd64
+
+## Audit rosdistro
+
+    ./scripts/release/audit_rosdistro.py --cache-dir /tmp/rosdistrocache https://raw.githubusercontent.com/ros-infrastructure/ros_buildfarm_config/production/index.yaml noetic
--- a/doc/ongoing_operations.rst
+++ b/doc/ongoing_operations.rst
@@ -40,14 +40,30 @@ The sync to the ``main`` repository affects all architectures.
 Guidelines for gating a sync to main
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-When preparing for a sync it is recommended to review the
-*release status page*.
-There are filters for viewing regressions, and what packages will sync if the
-sync to main is run as well as the ability to search.
+There is a manual process to go from testing to main.
+We refer to this as a sync.
 
-Note: A version downgrade is also marked as a regression since if users have
-already installed the previous package with the higher version number apt will
-not install the newer package with a lower version automatically.
+When you are preparing for a sync:
+
+* Run the ``rosdistro_audit.py`` script.
+  It will output text which is appropriate for posting to Discourse as a triage list.
+  The ``rosdistro_audit.py`` will report all packages which are failing to build for a whole buildfile.
+
+  * If it's failing on specific architectures, ticket it upstream and blacklist it in the config with a cross reference.
+  * If it's failing on all platforms, rollback or remove the release from the distro.
+* Check the list of packages comes from the status page, e.g. http://repositories.ros.org/status_page/ros_melodic_default.html
+  Click “REGRESSION” to find the list of regressions, click “SYNC” to see how many packages there are to sync.
+  Make sure there are no major regressions and ticket any with the maintainers.
+
+  Note: A version downgrade is also marked as a regression since apt will not automatically install a newer package with a lower version.
+
+* Announce the intent to sync with the above information.
+  Example: https://discourse.ros.org/t/preparing-for-kinetic-sync-2020-08-20/16002/5
+
+* After a period of time for resolving the identified issues above run the sync on Jenkins.
+* Once sync is done: announce on discourse. The details of the sync is in the console output of the "sync to main" job (make sure to not take the ones that generate debug packages otherwise the package count is off by a factor 2)
+* Tag the rosdistro with the format ``ROSDISTRO/YYYY-MM-DD``
+
 
 Importing new upstream packages
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
--- a/ros_buildfarm/argument.py
+++ b/ros_buildfarm/argument.py
@@ -474,6 +474,14 @@ def add_argument_testing(parser):
              'them.')
 
 
+def add_argument_return_zero(parser):
+    parser.add_argument(
+        '--return-zero', action='store_true',
+        help='Always return zero, success, even if there are errors. '
+             'This is used for testing, and scripting. It is not generally'
+             ' recommended for use.')
+
+
 def check_len_action(minargs, maxargs):
     class CheckLength(argparse.Action):
 
--- a/ros_buildfarm/common.py
+++ b/ros_buildfarm/common.py
@@ -649,14 +649,26 @@ def get_downstream_package_names(pkg_names, dependencies):
     return downstream_pkg_names
 
 
-def get_implicitly_ignored_package_names(cached_pkgs, explicitly_ignored_pkg_names):
-    pkg_names = set(cached_pkgs.keys())
+def get_package_manifests(dist):
+    cached_pkgs = {}
+    for pkg_name in dist.release_packages.keys():
+        pkg_xml = dist.get_release_package_xml(pkg_name)
+        if pkg_xml is not None:
+            from catkin_pkg.package import InvalidPackage, parse_package_string
+            try:
+                pkg_manifest = parse_package_string(pkg_xml)
+            except InvalidPackage:
+                continue
+            cached_pkgs[pkg_name] = pkg_manifest
+    return cached_pkgs
 
+
+def get_implicitly_ignored_package_names(cached_pkgs, explicitly_ignored_pkg_names):
     # get direct dependencies from distro cache for each package
     direct_dependencies = {}
-    for pkg_name in pkg_names:
+    for pkg_name in cached_pkgs.keys():
         direct_dependencies[pkg_name] = get_direct_dependencies(
-            pkg_name, cached_pkgs, pkg_names) or set([])
+            pkg_name, cached_pkgs, cached_pkgs) or set([])
 
     # find recursive downstream deps for all explicitly ignored packages
     ignored_pkg_names = set(explicitly_ignored_pkg_names)
@@ -669,6 +681,46 @@ def get_implicitly_ignored_package_names(cached_pkgs, explicitly_ignored_pkg_nam
         break
 
     return ignored_pkg_names.difference(explicitly_ignored_pkg_names)
+
+
+def filter_blocked_dependent_package_names(cached_pkgs, failed_pkg_names):
+    """Return the list of packages that are missing and not blocked.
+
+    Return the list of packages that are missing that are not depending
+    on other missing packages.
+    """
+    # get direct dependencies from distro cache for each package
+    direct_dependencies = {}
+    for pkg_name in cached_pkgs:
+        direct_dependencies[pkg_name] = get_direct_dependencies(
+            pkg_name, cached_pkgs, cached_pkgs) or set([])
+
+    # find recursive downstream deps for all explicitly ignored packages
+    all_deps = get_downstream_package_names(
+            failed_pkg_names, direct_dependencies)
+    while True:
+        blocked_pkgs = get_downstream_package_names(
+            all_deps, direct_dependencies)
+        if blocked_pkgs - all_deps:
+            all_deps |= blocked_pkgs
+            continue
+        break
+
+    return failed_pkg_names.difference(all_deps)
+
+
+def filter_buildfile_packages_recursively(package_names, buildfile, rosdistro_name):
+    """Filter packages based on the build including recursively blocked packages.
+
+    Filter a list of packages based on a build file's blacklist and whitelist
+    including implicit blacklisting of dependent packages for a specific rosdistro.
+    """
+    res = buildfile.filter_packages(package_names)
+    cached_pkgs = get_package_manifests(rosdistro_name)
+    implicitly_ignored = get_implicitly_ignored_package_names(
+        cached_pkgs, buildfile.package_blacklist)
+    res -= implicitly_ignored
+    return res
 
 
 def get_package_condition_context(index, rosdistro_name):
--- a/ros_buildfarm/release_job.py
+++ b/ros_buildfarm/release_job.py
@@ -1,4 +1,4 @@
-# Copyright 2014-2016 Open Source Robotics Foundation, Inc.
+# Copyright 2014-2020 Open Source Robotics Foundation, Inc.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -17,6 +17,8 @@ from __future__ import print_function
 from collections import OrderedDict
 import sys
 
+from ros_buildfarm.common import filter_blocked_dependent_package_names
+from ros_buildfarm.common import filter_buildfile_packages_recursively
 from ros_buildfarm.common import get_binarydeb_job_name
 from ros_buildfarm.common import get_default_node_label
 from ros_buildfarm.common import get_direct_dependencies
@@ -25,6 +27,7 @@ from ros_buildfarm.common import get_implicitly_ignored_package_names
 from ros_buildfarm.common import get_node_label
 from ros_buildfarm.common import get_os_package_name
 from ros_buildfarm.common import get_package_condition_context
+from ros_buildfarm.common import get_package_manifests
 from ros_buildfarm.common import get_release_binary_view_name
 from ros_buildfarm.common import get_release_job_prefix
 from ros_buildfarm.common import get_release_source_view_name
@@ -40,8 +43,11 @@ from ros_buildfarm.config import get_distribution_file
 from ros_buildfarm.config import get_index as get_config_index
 from ros_buildfarm.config import get_release_build_files
 from ros_buildfarm.git import get_repository
+from ros_buildfarm.package_repo import get_package_repo_data
 from ros_buildfarm.templates import expand_template
+from rosdistro import get_cached_distribution
 from rosdistro import get_distribution_cache
+from rosdistro import get_distribution_file as rosdistro_get_distribution_file
 from rosdistro import get_index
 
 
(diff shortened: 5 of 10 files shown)


scripts/release/check_sync_criteria.py
```diff
@@ -54,7 +54,7 @@ def main(argv=sys.argv[1:]):
     return 0 if success else 1
 
 
-def check_sync_criteria(
+def get_missing_packages(
```
nuclearsandwich, line 62: This function is called `get_missing_packages` but has a multiple return value that include `binary_packages`. That seems worth a docblock or a revisited name to me.

scripts/release/audit_rosdistro.py
```diff
@@ -0,0 +1,129 @@
...
+            else:
+                missing_packages[bf_name]['all'] = missing_pkgs
+
+            if 'all' in missing_packages:
+                missing_packages['all'] &= missing_pkgs
+            else:
+                missing_packages['all'] = missing_pkgs
+
+
+    recommended_actions = len(missing_packages['all'])
```
nuclearsandwich, line 77: A bit esoteric for a first review but if I don't mention it now I'll forget.
If there are exactly 256 missing packages (or 512, 768 or anything congruent to 0 modulo 255) this will have an exit code of 0 due to the overflow. What's the intent for using a semantically significant exit code?

ros_buildfarm/config/release_build_file.py
```diff
@@ -100,9 +101,16 @@ def __init__(self, name, data):  # noqa: D107
         assert 'upload_credential_id' in data
         self.upload_credential_id = data['upload_credential_id']
 
-    def filter_packages(self, package_names):
+    def filter_packages(self, package_names, remove_implicit_distro=None):
+        # TODO(tfoote) get the distro from the constructor instead
```
dirk-thomas, line 105: Are you planning to resolve this TODO before converting the PR from a draft?

ros_buildfarm/config/release_build_file.py
```diff
@@ -100,9 +101,16 @@ def __init__(self, name, data):  # noqa: D107
...
+        # of passing it in and just have a bool flag
         res = set(package_names)
         if self.package_whitelist:
             res &= set(self.package_whitelist)
         res -= set(self.package_blacklist)
+        if remove_implicit_distro:
+            cached_pkgs = get_cached_package_manifests(remove_implicit_distro)
+            implicitly_ignored = get_implicitly_ignored_package_names(
+                cached_pkgs, self.package_blacklist)
+            res -= implicitly_ignored
```
dirk-thomas, line 115: Until now callers of this method are doing the implicit filtering on their own. I would suggest to either do the same for the new case or instead create a function for this (outside of the build file which is only supposed to be a simple data class) and also update existing code to use that new function.

ros_buildfarm/common.py
```diff
@@ -634,6 +634,21 @@ def get_downstream_package_names(pkg_names, dependencies):
     return downstream_pkg_names
 
 
+def get_cached_package_manifests(dist):
```
dirk-thomas, line 637: The function name might be misleading. The manifests don't seem to be fetched from the cache but queried through the manifest providers, no?

scripts/release/audit_rosdistro.py
```diff
@@ -0,0 +1,152 @@
+#!/usr/bin/env python3
+
+# Copyright 2014-2020 Open Source Robotics Foundation, Inc.
```
dirk-thomas, line 3: Nitpick: was code for this new module copied from somewhere> If yes, from where? If no, the year should probably just be 2020.

scripts/release/audit_rosdistro.py
```diff
@@ -0,0 +1,152 @@
...
+from ros_buildfarm.status_page import get_jenkins_job_urls
+
+from rosdistro import get_distribution_file
+from rosdistro import get_cached_distribution
+from rosdistro import get_index
+
+
+def main(argv=sys.argv[1:]):
+    parser = argparse.ArgumentParser(
+        description='Check if the sync criteria are matched to sync ' +
```
dirk-thomas, line 45: The description seems to be copy-n-pasted from somewhere else.

Nitpick: no need for `+`.

scripts/release/audit_rosdistro.py
```diff
@@ -0,0 +1,152 @@
...
+            print('   jenkins_urls:')
+            for target, ju in jenkins_urls.items():
+                print('    - [%s](%s)' % (' '.join([x for x in target]), ju.format(pkg = rb)))
+                # TODO(tfoote) embed build status when buildfarm has https
+                # print('    - %s [![Build Status](%s)](%s)' % (' '.join([x for x in target]), ju.format(pkg = rb) + '/badge/icon', ju.format(pkg = rb)))
+
+    return recommended_actions
+
+
+# TODO(tfoote) deduplicate with copy in check_sync_criteria.py
```
dirk-thomas, line 104: I assume this will be addressed before being considered for merging.

ros_buildfarm/config/release_build_file.py
tfoote: I extended this method to support filtering on either the explicit list or what's often going to be more valuable to users the explicit + implicit list. For example we could easily leverage the implicit list on the status page. 

I can turn it into a second method that calls the first method. This is already a class with methods and not a simple data class. I can go ahead and make it into a separate method but it doesn't feel like it's significantly different from the current one, it's just a slightly different filter.

tfoote:
It's now creating copy and pastable content for opening tickets.

ros_buildfarm/config/release_build_file.py
dirk-thomas: Please move this into a separate function (in a different file) since until now this class acts as a data structure which is only concerned about the information coming from that build file and not importing any other modules. And the goal is to keep it that way to not mix other responsibilities.

ros_buildfarm/release_job.py
```diff
@@ -17,8 +17,10 @@
 from collections import OrderedDict
 import sys
 
+from ros_buildfarm.common import filter_blocked_dependent_package_names
 from ros_buildfarm.common import get_binarydeb_job_name
 from ros_buildfarm.common import get_default_node_label
+from ros_buildfarm.common import get_package_manifests
```
dirk-thomas, line 23: Nitpick: alpha order.

Same below.

scripts/release/audit_rosdistro.py
```diff
@@ -0,0 +1,130 @@
...
+            else:
+                missing_packages[bf_name]['all'] = missing_pkgs
+
+            if 'all' in missing_packages:
+                missing_packages['all'] &= missing_pkgs
+            else:
+                missing_packages['all'] = missing_pkgs
+
+    recommended_actions = len(missing_packages['all'])
+    print("# Sync preparation report for %s " % rosdistro_name)
```
dirk-thomas, line 72: Please use single quote were possible without escaping. The message also has a trailing space.

Same below.

scripts/release/audit_rosdistro.py
```diff
@@ -0,0 +1,130 @@
...
+        recommended_actions += len(recommended_blacklists)
+        for rb in recommended_blacklists:
+            print(' - %s:' % rb)
+            jenkins_urls = get_jenkins_job_urls(
+                rosdistro_name,
+                config.jenkins_url,
+                bf_name,
+                build_files[bf_name].get_targets_list())
+            url = get_package_repository_link(dist_file, rb)
+            print(f'   - Suggested ticket location [{url}]({url})')
```
dirk-thomas, line 104: While the script has a Python 3 shebang all scripts in this package currently support Python 2.7 as well as minor version of Python 3 which don't support f-strings. So please use normal string instead.

That being said it would be good to add an invocation of this new script to the CI file to ensure it generates the output as desired.

scripts/release/audit_rosdistro.py
tfoote: I changed it to return 1 if not zero recommendations.

ros_buildfarm/common.py
```diff
@@ -656,6 +671,31 @@ def get_implicitly_ignored_package_names(cached_pkgs, explicitly_ignored_pkg_nam
     return ignored_pkg_names.difference(explicitly_ignored_pkg_names)
 
 
+def filter_blocked_dependent_package_names(cached_pkgs, failed_pkg_names):
+    """Return the list of packages that are missing that are not depending
+    on other missing packages."""
+    pkg_names = set(cached_pkgs.keys())
```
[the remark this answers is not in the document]
dirk-thomas: Still pending.

scripts/release/check_sync_criteria.py
```diff
@@ -48,69 +43,12 @@ def main(argv=sys.argv[1:]):
...
-    success = check_sync_criteria(
+    target = Target(args.os_name, args.os_code_name, args.arch)
+    success = partition_packages(
         args.config_url, args.rosdistro_name, args.release_build_name,
-        args.os_name, args.os_code_name, args.arch, args.cache_dir)
+        target, args.cache_dir)
     return 0 if success else 1
 
 
-def check_sync_criteria(
```
dirk-thomas, line 57: Where have the checks of this function been moved to? I assume we still want to check `sync_packages` / `sync_package_count`.

tfoote: I was going to make it more strict with this requiring no regressions but we can do this one step at a time and not enforce that yet.

dirk-thomas: This patch doesn't update any of the existing documentation and I don't think that kind of change (by default) is even desirable. These sync criteria are what is defined in the config and should continue to work as is. Therefore I would suggest to add a new option to optionally perform additional checks.

scripts/release/audit_rosdistro.py
```diff
@@ -0,0 +1,135 @@
...
+    print('# Sync preparation report for %s' % rosdistro_name)
+    print('Prepared for configuration: %s' % config_url)
+    print('Prepared for rosdistro index: %s' % config.rosdistro_index_url)
+
+    print('## Packages failing on all platforms\n Release should be rolled back:')
+    for mp in sorted(missing_packages['all']):
+        print(' - %s ' % mp)
+
+    def get_package_repository_link(dist_file, pkg_name):
+        """Return the best guess of the url for filing a ticket against the package."""
```
nuclearsandwich, line 82: It doesn't seem like we necessarily have cached package.xml data at this stage but if we did and the package specified an issue url this could be even more precise.

ros_buildfarm/argument.py
```diff
@@ -474,6 +474,14 @@ def add_argument_testing(parser):
              'them.')
 
 
+def add_argument_return_zero(parser):
```
nuclearsandwich, line 477: I gather that earlier iterations of this PR had the new audit scripts run on the buildfarm and used as an additional guard on sync jobs. But since that is no longer the case is this argument still needed?

doc/ongoing_operations.rst
```diff
@@ -40,14 +40,33 @@ The sync to the ``main`` repository affects all architectures.
...
-
-Note: A version downgrade is also marked as a regression since if users have
-already installed the previous package with the higher version number apt will
-not install the newer package with a lower version automatically.
+There is a manual process to go from testing to main.
+We refer to this as a sync.
+
+When you are preparing for a sync:
+
+* Run the ``rosdistro_audit.py`` script it will output text which is appropriate for posting to Discourse as a triage list.
```
nuclearsandwich, line 48:
```suggestion
* Run the ``rosdistro_audit.py`` script.
  It will output text which is appropriate for posting to Discourse as a triage list.
```
or

```suggestion
* Run the ``rosdistro_audit.py`` script which will output text which is appropriate for posting to Discourse as a triage list.
```

I like the first suggestion as it's simpler English but either would do.

doc/ongoing_operations.rst
```diff
@@ -40,14 +40,33 @@ The sync to the ``main`` repository affects all architectures.
...
+
+  * If it's failing on specific architectures, ticket it upstream and blacklist it in the config with a cross reference.
+  * If it's failing on all platforms, rollback or remove the release from the distro.
+* Check the list of packages comes from the status page, e.g. http://repositories.ros.org/status_page/ros_melodic_default.html
+  Click “REGRESSION” to find the list of regressions, click “SYNC” to see how many packages there are to sync.
+  Make sure there are no major regressions and ticket any with the maintainers.
+
+  Note: A version downgrade is also marked as a regression since if users have
+  already installed the previous package with the higher version number apt will
+  not install the newer package with a lower version automatically.
```
nuclearsandwich, line 59:
```suggestion
  Note: A version downgrade is also marked as a regression since apt will not automatically install a newer package with a lower version.
```

scripts/release/audit_rosdistro.py
```diff
@@ -0,0 +1,135 @@
...
+                target_str = ' '.join([x for x in target])
+                url = ju.format(pkg=rb)
+                print('          - [%s](%s)' % (target_str, url))
+                # TODO(tfoote) embed build status when buildfarm has https
+                # print('    - %s [![Build Status](%s)](%s)' % (' '.join([x for x in target]),
+                #       ju.format(pkg = rb) + '/badge/icon', ju.format(pkg = rb)))
+            print('       This is being filed because this package is about to be blacklisted.'
+                  ' If this ticket is resolved please review whether it can be removed from'
+                  ' the blacklist that should cross reference here.')
+            print('')
```
nuclearsandwich, line 129: Consider moving this padding space up one level or not printing "Packages should be blacklisted:" unless there are some. The Dashing result has several stanzas that appear like this:
```markdown
## Packages failing for buildfile default

Packages should be blacklisted:
## Packages failing for buildfile ubv8

```
which doesn't affect the rendered markdown but made looking at the plaintext harder to follow.

ros_buildfarm/argument.py
tfoote: This argument was added for being able to run on Travis and differentiate between a completed report vs a crash.

ros_buildfarm/common.py
```diff
@@ -656,6 +668,46 @@ def get_implicitly_ignored_package_names(cached_pkgs, explicitly_ignored_pkg_nam
...
+    # find recursive downstream deps for all explicitly ignored packages
+    all_deps = get_downstream_package_names(
+            failed_pkg_names, direct_dependencies)
+    while True:
+        blocked_pkgs = get_downstream_package_names(
+            all_deps, direct_dependencies)
+        if blocked_pkgs - all_deps:
+            all_deps |= blocked_pkgs
+            continue
+        break
```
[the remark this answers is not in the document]
dirk-thomas: The logic within this loop should certainly ensure that it terminates. The implicitly assumption is that each iteration `all_deps` is increasing and therefore at some point reaching `blocked_pkgs` and `break`. That should be explicitly check with an `assert` to ensure this doesn't loop indefinitely.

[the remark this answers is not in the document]
tfoote: This logic is guaranteed to terminate. 

Without going into the details of `get_downstream_package_names`, it returns a subset of the packages in the dependency tree defined in `direct_dependencies` which does not change throughout the loop. Making no assumptions about a valid structure of the dependency tree. It will return zero or more packages from the dependency tree.

If it returns zero packages, the loop breaks. 

The results of each iteration is compared to a list of previously detected dependencies, and will only continue to loop if it's returning one or more new packages to add to the list. If it returns a set of packages which are fully contained in the known list it will break out. Thus the worst case is that it returns 1 new package every time, and in that case it will only run for N cycles where N is the number of packages in the dependency tree. And that worse case will only happen if there's a purely linear dependency with N layers.

The case of a malformed dependency tree, such as one with loops, it will still exit because the dependencies returned will be in the existing list and thus terminate the iteration as soon as the loop cycle starts to repeat because there are no new packages.

In reality this will iterate N times where N is the depth of the dependency tree from the initial packages.

scripts/release/audit_rosdistro.py
tfoote: I improved the wording in 39089a8cfff8e9650eb2f3ce4373b06589750a69

ros_buildfarm/common.py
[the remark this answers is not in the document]
dirk-thomas: Thanks for describing the logic in more detail. I still think add an assertion to ensure this is the case is the defensive way to write this loop.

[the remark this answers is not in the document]
tfoote: What are you asking to be asserted?

You mention this above:

> The implicit assumption is that in each iteration all_deps is growing and therefore at some point reaching blocked_pkgs and break the loop. That should be explicitly checked with an assert to ensure this doesn't loop indefinitely.

But this is not an implicit assumption that should be asserted about. It is explicitly checked in the if statement. If there are no new dependencies then the loop breaks.

scripts/release/check_sync_criteria.py
```diff
@@ -108,7 +110,6 @@ def check_sync_criteria(
         print('%d binary packages available ' % binary_package_count +
               '(more or equal then the configured sync limit of %d)' %
               build_file.sync_package_count)
-
```
dirk-thomas, line 111: Nitpick: please revert unrelated changes like this.

README.md
```diff
@@ -15,3 +15,15 @@ to configure the jobs being generated (e.g.
 If you are going to use any of the provided infrastructure please consider
 watching the [buildfarm Discourse category](https://discourse.ros.org/c/buildfarm)
 in order to receive notifications e.g. about any upcoming changes.
+
+
```
dirk-thomas, line 19: Nitpick: please remove double / repeated empty lines.

Same below.

ros_buildfarm/common.py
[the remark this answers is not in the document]
dirk-thomas: > If it returns zero packages, the loop breaks.

I guess I didn't see that the condition `blocked_pkgs - all_deps` is doing both at the same time:
* it ensures that if `blocked_pkgs` is empty that it `breaks` as well as
* it ensures that `blocked_pkgs` later added to `all_deps` contains at least one element not already in `all_deps` which ensured progress.

ros_buildfarm/argument.py
dirk-thomas: Instead of adding an argument for this I would argue the script should be invoked in a way which ignores the return code (e.g. `... || true`).

tfoote: I initially thought of using `|| true` when adding it to travis. However, I realized that would also mask any crashes and make it not useful to run the test on travis at all since it would never fail. Thus I took the extra time to implement the additional argument which will return zero if the program completes successfully but had a failing report which is distinct from what the unit test cares about which is crashing and failing to complete which will still generate a non-zero return code.

dirk-thomas (approved):

PR #816, opened 2020-06-23
Add Suite3 and Python2-Depends-Name configuration for stdeb releases.

nuclearsandwich:
This PR includes two changes:

#### Adds `Suite3` for releasing to Ubuntu Focal

The Suite3 option introduced in stdeb 0.9.0 and allows specifying
different release suites when building a python3-only deb.

#### Sets Python2-Depends-Name for releasing from Ubuntu Focal

Requires stdeb >= 0.9.1 and makes the created python2 package uninstallable on Ubuntu Focal and later where the `python` package is called `python2` instead.

---

If there is some reason why we are not planning to release ros_buildfarm into Ubuntu Focal at this time the first commit could be dropped during review.


dirk-thomas:
@nuclearsandwich I retriggered the failed CI builds. With them passing green is there anything holding this back?

PR #817, opened 2020-07-01
Install abi-compliance-checker from source to avoid Focal package bug

j-rivero:
The `abi-compliance-checker` package in Ubuntu Focal has a bug introduced by a patch done in Debian. [The bug](https://github.com/lvc/abi-compliance-checker/pull/80#issuecomment-652521014) prevents the tool from exit when being used from python scripts.

I'm modifying the code in the buildfarm to install the 2.3 version of the tool from upstream sources. Code is perl, it does not need too much apart from the own perl and make.

### DIFF
--- a/ros_buildfarm/templates/devel/devel_task.Dockerfile.em
+++ b/ros_buildfarm/templates/devel/devel_task.Dockerfile.em
@@ -71,7 +71,7 @@ RUN python3 -u /tmp/wrapper_scripts/apt.py update-install-clean -q -y ros-@(rosd
 @[end if]@
 RUN python3 -u /tmp/wrapper_scripts/apt.py update-install-clean -q -y ccache
 
-RUN python3 -u /tmp/wrapper_scripts/apt.py update-install-clean -q -y python3-catkin-pkg-modules
+RUN python3 -u /tmp/wrapper_scripts/apt.py update-install-clean -q -y python3-catkin-pkg-modules python3-pip
 @[if run_abichecker]@
 @{ubuntu_before_focal=('xenial', 'bionic')}@
 @[if os_name == 'ubuntu' and os_code_name not in ubuntu_before_focal]@
--- a/ros_buildfarm/templates/devel/devel_task.Dockerfile.em
+++ b/ros_buildfarm/templates/devel/devel_task.Dockerfile.em
@@ -71,21 +71,6 @@ RUN python3 -u /tmp/wrapper_scripts/apt.py update-install-clean -q -y ros-@(rosd
 @[end if]@
 RUN python3 -u /tmp/wrapper_scripts/apt.py update-install-clean -q -y ccache
 
-@(TEMPLATE(
-    'snippet/set_environment_variables.Dockerfile.em',
-    environment_variables=build_environment_variables,
-))@
-
-@(TEMPLATE(
-    'snippet/install_dependencies.Dockerfile.em',
-    dependencies=dependencies,
-    dependency_versions=dependency_versions,
-))@
-
-@(TEMPLATE(
-    'snippet/install_dependencies_from_file.Dockerfile.em',
-    install_lists=install_lists,
-))@
 RUN python3 -u /tmp/wrapper_scripts/apt.py update-install-clean -q -y python3 python3-catkin-pkg-modules python3-pip
 @[if run_abichecker]@
 @{ubuntu_before_focal=('xenial', 'bionic')}@
@@ -102,6 +87,22 @@ RUN python3 -u /tmp/wrapper_scripts/apt.py update-install-clean -q -y abi-compli
 @[end if]@
 RUN pip3 install -U auto_abi_checker
 @[end if]@
+
+@(TEMPLATE(
+    'snippet/set_environment_variables.Dockerfile.em',
+    environment_variables=build_environment_variables,
+))@
+
+@(TEMPLATE(
+    'snippet/install_dependencies.Dockerfile.em',
+    dependencies=dependencies,
+    dependency_versions=dependency_versions,
+))@
+
+@(TEMPLATE(
+    'snippet/install_dependencies_from_file.Dockerfile.em',
+    install_lists=install_lists,
+))@
 
 # After all dependencies are installed, update ccache symlinks.
 # This command is supposed to be invoked whenever a new compiler is installed
--- a/ros_buildfarm/templates/devel/devel_task.Dockerfile.em
+++ b/ros_buildfarm/templates/devel/devel_task.Dockerfile.em
@@ -71,8 +71,8 @@ RUN python3 -u /tmp/wrapper_scripts/apt.py update-install-clean -q -y ros-@(rosd
 @[end if]@
 RUN python3 -u /tmp/wrapper_scripts/apt.py update-install-clean -q -y ccache
 
-RUN python3 -u /tmp/wrapper_scripts/apt.py update-install-clean -q -y python3-catkin-pkg-modules python3-pip
 @[if run_abichecker]@
+RUN python3 -u /tmp/wrapper_scripts/apt.py update-install-clean -q -y python3-catkin-pkg-modules python3-pip
 @{ubuntu_before_focal=('xenial', 'bionic')}@
 @[if os_name == 'ubuntu' and os_code_name not in ubuntu_before_focal]@
 # Focal/Groovy abi-compliance-checker package has a bug that breaks python invocation
--- a/ros_buildfarm/templates/devel/devel_task.Dockerfile.em
+++ b/ros_buildfarm/templates/devel/devel_task.Dockerfile.em
@@ -73,8 +73,7 @@ RUN python3 -u /tmp/wrapper_scripts/apt.py update-install-clean -q -y ccache
 
 @[if run_abichecker]@
 RUN python3 -u /tmp/wrapper_scripts/apt.py update-install-clean -q -y python3-catkin-pkg-modules python3-pip
-@{ubuntu_before_focal=('xenial', 'bionic')}@
-@[if os_name == 'ubuntu' and os_code_name not in ubuntu_before_focal]@
+@[if os_name == 'ubuntu' and os_code_name not in ('xenial', 'bionic')]@
 # Focal/Groovy abi-compliance-checker package has a bug that breaks python invocation
 # See: https://github.com/lvc/abi-compliance-checker/pull/80#issuecomment-652521014
 # Install 2.3 version from source, needs perl
--- a/ros_buildfarm/templates/devel/devel_task.Dockerfile.em
+++ b/ros_buildfarm/templates/devel/devel_task.Dockerfile.em
@@ -100,7 +100,7 @@ RUN curl -sL https://github.com/lvc/abi-compliance-checker/archive/2.3.tar.gz | 
     make install prefix=/usr -C /tmp/abi-compliance-checker-2.3 && \
     rm -fr /tmp/abi-compliance
 @[else]@
-RUN python3 -u /tmp/wrapper_scripts/apt.py update-install-clean -q -y abi-compliance-checker @(common_deps)
+RUN python3 -u /tmp/wrapper_scripts/apt.py update-install-clean -q -y abi-compliance-checker
 @[end if]@
 RUN python3 -u /tmp/wrapper_scripts/apt.py update-install-clean -q -y @(common_deps)
 RUN pip3 install -U auto_abi_checker
--- a/ros_buildfarm/templates/devel/devel_task.Dockerfile.em
+++ b/ros_buildfarm/templates/devel/devel_task.Dockerfile.em
@@ -88,13 +88,18 @@ RUN python3 -u /tmp/wrapper_scripts/apt.py update-install-clean -q -y ccache
 ))@
 
 @[if run_abichecker]@
+@{common_deps='python3 python3-catkin-pkg-modules python3-pip'}
+@[if os_name == 'ubuntu' and os_code_name == 'focal' ]@
 # Focal abi-compliance-checker package has a bug that breaks python invocation
 # See: https://github.com/lvc/abi-compliance-checker/pull/80#issuecomment-652521014
 # Install 2.3 version from source, needs perl
-RUN python3 -u /tmp/wrapper_scripts/apt.py update-install-clean -q -y curl make perl python3 python3-catkin-pkg-modules python3-pip
+RUN python3 -u /tmp/wrapper_scripts/apt.py update-install-clean -q -y curl make perl
 RUN curl -sL https://github.com/lvc/abi-compliance-checker/archive/2.3.tar.gz | tar xvz -C /tmp
 RUN make install prefix=/usr -C /tmp/abi-compliance-checker-2.3
-RUN pip3 install -U auto_abi_checker
+@[else]@
+RUN python3 -u /tmp/wrapper_scripts/apt.py update-install-clean -q -y abi-compliance-checker @(common_deps)
+@[end if]@
+RUN pip3 install -U auto_abi_checker @(common_deps)
 @[end if]@
 
 # After all dependencies are installed, update ccache symlinks.
--- a/ros_buildfarm/templates/devel/devel_task.Dockerfile.em
+++ b/ros_buildfarm/templates/devel/devel_task.Dockerfile.em
@@ -88,8 +88,10 @@ RUN python3 -u /tmp/wrapper_scripts/apt.py update-install-clean -q -y ccache
 ))@
 
 @[if run_abichecker]@
-@{common_deps='python3 python3-catkin-pkg-modules python3-pip'}
-@[if os_name == 'ubuntu' and short_os_code_name >= 'F']@
+@{common_deps='python3 python3-catkin-pkg-modules python3-pip'
+ubuntu_before_focal=('xenial', 'bionic')
+}@
+@[os_name == 'ubuntu' and os_code_name not in ubuntu_before_focal]@
 # Focal/Groovy abi-compliance-checker package has a bug that breaks python invocation
 # See: https://github.com/lvc/abi-compliance-checker/pull/80#issuecomment-652521014
 # Install 2.3 version from source, needs perl
--- a/ros_buildfarm/templates/devel/devel_task.Dockerfile.em
+++ b/ros_buildfarm/templates/devel/devel_task.Dockerfile.em
@@ -89,8 +89,8 @@ RUN python3 -u /tmp/wrapper_scripts/apt.py update-install-clean -q -y ccache
 
 @[if run_abichecker]@
 @{common_deps='python3 python3-catkin-pkg-modules python3-pip'}
-@[if os_name == 'ubuntu' and os_code_name == 'focal' ]@
-# Focal abi-compliance-checker package has a bug that breaks python invocation
+@[if os_name == 'ubuntu' and short_os_code_name >= 'F']@
+# Focal/Groovy abi-compliance-checker package has a bug that breaks python invocation
 # See: https://github.com/lvc/abi-compliance-checker/pull/80#issuecomment-652521014
 # Install 2.3 version from source, needs perl
 RUN python3 -u /tmp/wrapper_scripts/apt.py update-install-clean -q -y curl make perl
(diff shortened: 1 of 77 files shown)


ros_buildfarm/templates/devel/devel_task.Dockerfile.em
```diff
@@ -87,7 +87,12 @@ RUN python3 -u /tmp/wrapper_scripts/apt.py update-install-clean -q -y ccache
 ))@
 
 @[if run_abichecker]@
-RUN python3 -u /tmp/wrapper_scripts/apt.py update-install-clean -q -y abi-compliance-checker python3 python3-catkin-pkg-modules python3-pip 
+# Focal abi-compliance-checker package has a bug that breaks python invocation
```
dirk-thomas, line 90: Shouldn't the logic be conditional on `focal` then?

ros_buildfarm/templates/devel/devel_task.Dockerfile.em
```diff
@@ -87,7 +87,12 @@ RUN python3 -u /tmp/wrapper_scripts/apt.py update-install-clean -q -y ccache
...
+# See: https://github.com/lvc/abi-compliance-checker/pull/80#issuecomment-652521014
+# Install 2.3 version from source, needs perl
+RUN python3 -u /tmp/wrapper_scripts/apt.py update-install-clean -q -y curl make perl python3 python3-catkin-pkg-modules python3-pip
+RUN curl -sL https://github.com/lvc/abi-compliance-checker/archive/2.3.tar.gz | tar xvz -C /tmp
+RUN make install prefix=/usr -C /tmp/abi-compliance-checker-2.3
```
dirk-thomas, line 95: Should the last two `RUN` lines be collapsed into a single line to avoid the first being cached but the second failing (due to an incorrect download)? The line could also delete `/tmp/abi-compliance-checker-2.3` afterwards to keep the image size smaller.

dirk-thomas:
@j-rivero Any update on this?

dirk-thomas:
@j-rivero Any update on this?

ros_buildfarm/templates/devel/devel_task.Dockerfile.em
ahcorde, line 95:
```suggestion
curl -sL https://github.com/lvc/abi-compliance-checker/archive/2.3.tar.gz | tar xvz -C /tmp && make install prefix=/usr -C /tmp/abi-compliance-checker-2.3 && rm -r /tmp/abi-compliance-checker-2.3
```

ros_buildfarm/templates/devel/devel_task.Dockerfile.em
j-rivero: Yes 8441c39

ros_buildfarm/templates/devel/devel_task.Dockerfile.em
j-rivero: sounds correct a18d87bc7b7ff7be56209ec4adbb7a99c7c04d85

j-rivero: ouch sorry I saw your suggestion after editing the files.

j-rivero:
Tested locally using:
```bash
./create_devel_task_generator.py --rosdistro-name dashing --os-name ubuntu --os-code-name=bionic --arch amd64 --build-tool colcon --ros-version 2 --run-abichecker --workspace-root ~/code/ros2/w/ > Dockerfile
```
Dockerfile relevant snippet:
```Dockerfile
  RUN python3 -u /tmp/wrapper_scripts/apt.py update-install-clean -q -y abi-compliance-checker python3 python3-catkin-pkg-modules python3-pip
  RUN pip3 install -U auto_abi_checker python3 python3-catkin-pkg-modules python3-pip
```
With focal:
```bash
./create_devel_task_generator.py --rosdistro-name foxy --os-name ubuntu --os-code-name=focal --arch amd64 --build-tool colcon --ros-version 2 --run-abichecker --workspace-root ~/code/ros2/ws/ > Dockerfile
```
```Dockerfile
  # Focal abi-compliance-checker package has a bug that breaks python invocation
  # See: https://github.com/lvc/abi-compliance-checker/pull/80#issuecomment-652521014
  # Install 2.3 version from source, needs perl
  RUN python3 -u /tmp/wrapper_scripts/apt.py update-install-clean -q -y curl make perl
  RUN curl -sL https://github.com/lvc/abi-compliance-checker/archive/2.3.tar.gz | tar xvz -C /tmp && \
      make install prefix=/usr -C /tmp/abi-compliance-checker-2.3 && \
      rm -fr /tmp/abi-compliance
  RUN pip3 install -U auto_abi_checker python3 python3-catkin-pkg-modules python3-pip
```

ros_buildfarm/templates/devel/devel_task.Dockerfile.em
dirk-thomas: If the same problem applies to newer Ubuntu distros the condition should be changed to ensure this workaround is being used in the future (until a fix has landed upstream and it can be removed).

j-rivero: The same patch is shipped in Ubuntu Groovy. Changed in 027bb48

dirk-thomas: This kind of check doesn't work for e.g. Xenial. Please use the same approach used elsewhere in this repo: enumate all past distros for which the workaround *shouldn't* be applied.

ros_buildfarm/templates/devel/devel_task.Dockerfile.em
```diff
@@ -88,8 +88,19 @@ RUN python3 -u /tmp/wrapper_scripts/apt.py update-install-clean -q -y ccache
 ))@
 
 @[if run_abichecker]@
-RUN python3 -u /tmp/wrapper_scripts/apt.py update-install-clean -q -y abi-compliance-checker python3 python3-catkin-pkg-modules python3-pip 
-RUN pip3 install -U auto_abi_checker
+@{common_deps='python3 python3-catkin-pkg-modules python3-pip'}
+@[if os_name == 'ubuntu' and short_os_code_name >= 'F']@
```
ahcorde, line 92:
```suggestion
@{common_deps='python3 python3-catkin-pkg-modules python3-pip'
ubuntu_before_focal = (
    'precise', 'quantal', 'raring', 'saucy',
    'trusty', 'utopic', 'vivid', 'wily',
    'xenial', 'yakkety', 'zesty', 'artful', 'bionic')
}@
@[os_name == 'ubuntu' and os_code_name not in ubuntu_before_focal]@
```

scripts/ci/build_task_generator.py
```diff
@@ -110,6 +111,7 @@ def main(argv=sys.argv[1:]):
     data = {
         'os_name': args.os_name,
         'os_code_name': args.os_code_name,
+        'short_os_code_name': get_short_os_code_name(args.os_code_name),
```
ahcorde, line 114:
```suggestion
```

scripts/ci/build_task_generator.py
```diff
@@ -39,6 +39,7 @@
 from ros_buildfarm.argument import extract_multiple_remainders
 from ros_buildfarm.common import get_binary_package_versions
 from ros_buildfarm.common import get_distribution_repository_keys
+from ros_buildfarm.common import get_short_os_code_name
```
ahcorde, line 42:
```suggestion
```

scripts/devel/create_devel_task_generator.py
```diff
@@ -34,6 +34,7 @@
 from ros_buildfarm.common import get_binary_package_versions
 from ros_buildfarm.common import get_distribution_repository_keys
 from ros_buildfarm.common import get_packages_in_workspaces
+from ros_buildfarm.common import get_short_os_code_name
```
ahcorde, line 37:
```suggestion

```

scripts/devel/create_devel_task_generator.py
```diff
@@ -166,6 +167,7 @@ def main(argv=sys.argv[1:]):
     data = {
         'os_name': args.os_name,
         'os_code_name': args.os_code_name,
+        'short_os_code_name': get_short_os_code_name(args.os_code_name),
```
ahcorde, line 170:
```suggestion
```

ahcorde (changes requested):
@j-rivero, please review my suggestions

ros_buildfarm/templates/devel/devel_task.Dockerfile.em
j-rivero: Looks like a cleaner approach. I just reduced the list of distributions to the ones not EOL (Xenial, Bionic) and condensed the suggestions in one commit 88d5520972c47b3d816bd9a8fea1f5adfe62617e

ros_buildfarm/templates/devel/devel_task.Dockerfile.em
```diff
@@ -88,8 +88,21 @@ RUN python3 -u /tmp/wrapper_scripts/apt.py update-install-clean -q -y ccache
...
+# See: https://github.com/lvc/abi-compliance-checker/pull/80#issuecomment-652521014
+# Install 2.3 version from source, needs perl
+RUN python3 -u /tmp/wrapper_scripts/apt.py update-install-clean -q -y curl make perl
+RUN curl -sL https://github.com/lvc/abi-compliance-checker/archive/2.3.tar.gz | tar xvz -C /tmp && \
+    make install prefix=/usr -C /tmp/abi-compliance-checker-2.3 && \
+    rm -fr /tmp/abi-compliance
+@[else]@
+RUN python3 -u /tmp/wrapper_scripts/apt.py update-install-clean -q -y abi-compliance-checker @(common_deps)
+@[end if]@
+RUN pip3 install -U auto_abi_checker @(common_deps)
```
dirk-thomas, line 105: Doesn't this try to `pip3 install` the Debian package names `python3`, `python3-catkin-pkg-modules`, `python3-pip`?

ros_buildfarm/templates/devel/devel_task.Dockerfile.em
```diff
@@ -88,8 +88,21 @@ RUN python3 -u /tmp/wrapper_scripts/apt.py update-install-clean -q -y ccache
 ))@
 
 @[if run_abichecker]@
-RUN python3 -u /tmp/wrapper_scripts/apt.py update-install-clean -q -y abi-compliance-checker python3 python3-catkin-pkg-modules python3-pip 
-RUN pip3 install -U auto_abi_checker
+@{common_deps='python3 python3-catkin-pkg-modules python3-pip'
+ubuntu_before_focal=('xenial', 'bionic')
+}@
+@[os_name == 'ubuntu' and os_code_name not in ubuntu_before_focal]@
```
ahcorde, line 94:
```suggestion
@[if os_name == 'ubuntu' and os_code_name not in ubuntu_before_focal]@
```

ros_buildfarm/templates/devel/devel_task.Dockerfile.em
ahcorde, line 105:
```suggestion
RUN python3 -u /tmp/wrapper_scripts/apt.py update-install-clean -q -y abi-compliance-checker @(common_deps)
@[end if]@
RUN python3 -u /tmp/wrapper_scripts/apt.py update-install-clean -q -y @(common_deps)
RUN pip3 install -U auto_abi_checker
```

ahcorde (changes requested):

ros_buildfarm/templates/devel/devel_task.Dockerfile.em
```diff
@@ -88,7 +88,21 @@ RUN python3 -u /tmp/wrapper_scripts/apt.py update-install-clean -q -y ccache
...
+# See: https://github.com/lvc/abi-compliance-checker/pull/80#issuecomment-652521014
+# Install 2.3 version from source, needs perl
+RUN python3 -u /tmp/wrapper_scripts/apt.py update-install-clean -q -y curl make perl
+RUN curl -sL https://github.com/lvc/abi-compliance-checker/archive/2.3.tar.gz | tar xvz -C /tmp && \
+    make install prefix=/usr -C /tmp/abi-compliance-checker-2.3 && \
+    rm -fr /tmp/abi-compliance
+@[else]@
+RUN python3 -u /tmp/wrapper_scripts/apt.py update-install-clean -q -y abi-compliance-checker @(common_deps)
+@[end if]@
+RUN python3 -u /tmp/wrapper_scripts/apt.py update-install-clean -q -y @(common_deps)
```
dirk-thomas, line 105: This seems to be installing `common_deps` \twice (in case the above condition was `False` and the `else` block was executed).

ros_buildfarm/templates/devel/devel_task.Dockerfile.em
```diff
@@ -88,7 +88,21 @@ RUN python3 -u /tmp/wrapper_scripts/apt.py update-install-clean -q -y ccache
...
+@[if os_name == 'ubuntu' and os_code_name not in ubuntu_before_focal]@
+# Focal/Groovy abi-compliance-checker package has a bug that breaks python invocation
+# See: https://github.com/lvc/abi-compliance-checker/pull/80#issuecomment-652521014
+# Install 2.3 version from source, needs perl
+RUN python3 -u /tmp/wrapper_scripts/apt.py update-install-clean -q -y curl make perl
+RUN curl -sL https://github.com/lvc/abi-compliance-checker/archive/2.3.tar.gz | tar xvz -C /tmp && \
+    make install prefix=/usr -C /tmp/abi-compliance-checker-2.3 && \
+    rm -fr /tmp/abi-compliance
+@[else]@
+RUN python3 -u /tmp/wrapper_scripts/apt.py update-install-clean -q -y abi-compliance-checker @(common_deps)
```
ahcorde, line 103:
```suggestion
RUN python3 -u /tmp/wrapper_scripts/apt.py update-install-clean -q -y abi-compliance-checker
```

ros_buildfarm/templates/devel/devel_task.Dockerfile.em
```diff
@@ -88,7 +88,21 @@ RUN python3 -u /tmp/wrapper_scripts/apt.py update-install-clean -q -y ccache
...
+# See: https://github.com/lvc/abi-compliance-checker/pull/80#issuecomment-652521014
+# Install 2.3 version from source, needs perl
+RUN python3 -u /tmp/wrapper_scripts/apt.py update-install-clean -q -y curl make perl
+RUN curl -sL https://github.com/lvc/abi-compliance-checker/archive/2.3.tar.gz | tar xvz -C /tmp && \
+    make install prefix=/usr -C /tmp/abi-compliance-checker-2.3 && \
+    rm -fr /tmp/abi-compliance
+@[else]@
+RUN python3 -u /tmp/wrapper_scripts/apt.py update-install-clean -q -y abi-compliance-checker
+@[end if]@
+RUN python3 -u /tmp/wrapper_scripts/apt.py update-install-clean -q -y @(common_deps)
```
dirk-thomas, line 105: Nitpick: moving this `RUN` line above the conditional block will allow reusing the same layer across different Docker images.

ros_buildfarm/templates/devel/devel_task.Dockerfile.em
```diff
@@ -88,7 +88,21 @@ RUN python3 -u /tmp/wrapper_scripts/apt.py update-install-clean -q -y ccache
 ))@
```
ahcorde, line 89:
```suggestion
RUN python3 -u /tmp/wrapper_scripts/apt.py update-install-clean -q -y python3 python3-catkin-pkg-modules python3-pip
```

ros_buildfarm/templates/devel/devel_task.Dockerfile.em
ahcorde, line 105:
```suggestion
@[end if]@
```

ros_buildfarm/templates/devel/devel_task.Dockerfile.em
```diff
@@ -88,7 +88,21 @@ RUN python3 -u /tmp/wrapper_scripts/apt.py update-install-clean -q -y ccache
...
 
 @[if run_abichecker]@
-RUN python3 -u /tmp/wrapper_scripts/apt.py update-install-clean -q -y abi-compliance-checker python3 python3-catkin-pkg-modules python3-pip 
+@{common_deps='python3 python3-catkin-pkg-modules python3-pip'
+ubuntu_before_focal=('xenial', 'bionic')
+}@
```
ahcorde, line 93:
```suggestion
@{ubuntu_before_focal=('xenial', 'bionic')}@
```

ahcorde (changes requested):

dirk-thomas:
Actually should the whole ABI checker block be moved about the dependency installation? Otherwise the custom compilation will happen over and over again for packages which different dependencies.

ahcorde:
@dirk-thomas I'm not familiar with this repo. Can you give a little bit more of context?  

I can see here some dependencies installed in the same way:

https://github.com/ros-infrastructure/ros_buildfarm/blob/b13f25c55136a185b4218135f6fef52b47ce7e6d/ros_buildfarm/templates/devel/devel_task.Dockerfile.em#L61

Where is the right place to do it ?

dirk-thomas:
> I'm not familiar with this repo. Can you give a little bit more of context?

This isn't specific to this repository. It is about the order of `RUN` lines in a `Dockerfile`.

Basically each line creates a new layer in the Docker image. If the lines are different between two Dockerfiles they basically fork the image into two separate branches. So even if the different lines are followed by several identical lines they still different images / layers. Therefore the goal is to have as many identical lines from the beginning of Dockerfiles as possible.

> I can see here some dependencies installed in the same way:

The referenced line is static / doesn't change and therefore is identical for all Dockerfiles.

> Where is the right place to do it ?

Th following lines install the job specific dependencies and are therefore likely different for each Dockerfile. Therefore I think it makes sense to install the ABI checker above those lines: https://github.com/ros-infrastructure/ros_buildfarm/blob/b13f25c55136a185b4218135f6fef52b47ce7e6d/ros_buildfarm/templates/devel/devel_task.Dockerfile.em#L79-L88

ros_buildfarm/templates/devel/devel_task.Dockerfile.em
```diff
@@ -71,6 +71,23 @@ RUN python3 -u /tmp/wrapper_scripts/apt.py update-install-clean -q -y ros-@(rosd
 @[end if]@
 RUN python3 -u /tmp/wrapper_scripts/apt.py update-install-clean -q -y ccache
 
+RUN python3 -u /tmp/wrapper_scripts/apt.py update-install-clean -q -y python3-catkin-pkg-modules
```
dirk-thomas, line 74: Why is this now installed unconditionally?

ros_buildfarm/templates/devel/devel_task.Dockerfile.em
```diff
@@ -87,11 +104,6 @@ RUN python3 -u /tmp/wrapper_scripts/apt.py update-install-clean -q -y ccache
     install_lists=install_lists,
 ))@
 
-@[if run_abichecker]@
-RUN python3 -u /tmp/wrapper_scripts/apt.py update-install-clean -q -y abi-compliance-checker python3 python3-catkin-pkg-modules python3-pip
```
dirk-thomas, line 91: Why is `python3-pip` not needed anymore?

ahcorde: it's installed here https://github.com/ros-infrastructure/ros_buildfarm/blob/20a4a62433d7806705e7184aef4520bd7caf0028/ros_buildfarm/templates/devel/devel_task.Dockerfile.em#L64

dirk-thomas: > it's installed here

Only conditionally - see the line directly above:

https://github.com/ros-infrastructure/ros_buildfarm/blob/20a4a62433d7806705e7184aef4520bd7caf0028/ros_buildfarm/templates/devel/devel_task.Dockerfile.em#L63

ahcorde: my bad, but I think we can remove `python3` because the command uses `python3` to launch the script.

ahcorde: Restored https://github.com/ros-infrastructure/ros_buildfarm/pull/817/commits/469e8cad5ec41d0f9314d0713b45bf71d2162ae2

dirk-thomas (approved):

ahcorde:
@dirk-thomas feel free to merge it when CI has finished and it's green. I don't have permissions in the repo.

PR #821, opened 2020-07-22
quote test args containing spaces

dirk-thomas:
Follow up of #793 which didn't correctly preserve quoted arguments containing spaces.

Necessary for the CI configuration from ros2/ros_buildfarm_config#114.

Before: [Build Status](http://build.ros2.org/view/Rci/job/Rci__nightly-fastrtps_ubuntu_focal_amd64/21/)
After: [Build Status](http://build.ros2.org/view/Rci/job/Rci__nightly-fastrtps_ubuntu_focal_amd64/24/)


dirk-thomas:
Temporarily includes #822.

PR #826, opened 2020-07-29
Update Jenkins plugin versions used by ros_buildfarm.

nuclearsandwich:
In addition to the version updates there have also been a few additions or structural changes in the configurations.

The broadest is that the rebuild plugin now adds a configuration element to every job.
I've extracted that configuration element into a snippet and hard-coded the default value of false for both auto rebuilding and disabling rebuilding. If, in the future we want to change these properties for a job we can parameterize the snippet but I don't see a reason to do so pre-emptively.

When updating CI jobs I also noticed the addition of an scm element specifying a null scm property since the job itself is not tied to any scm repository.

There now appears to be a `<shallow>` tag in the Git SCM configuration.

### DIFF
--- a/ros_buildfarm/templates/misc/check_agents_job.xml.em
+++ b/ros_buildfarm/templates/misc/check_agents_job.xml.em
@@ -16,9 +16,6 @@
     'property_rebuild-settings',
 ))@
 @(SNIPPET(
-    'property_rebuild-settings',
-))@
-@(SNIPPET(
     'property_requeue-job',
 ))@
   </properties>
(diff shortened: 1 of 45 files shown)


ros_buildfarm/templates/misc/check_agents_job.xml.em
```diff
@@ -12,6 +12,12 @@
     'property_job-priority',
     priority=20,
 ))@
+@(SNIPPET(
+    'property_rebuild-settings',
+))@
+@(SNIPPET(
+    'property_rebuild-settings',
+))@
```
dirk-thomas, line 20: Twice?

nuclearsandwich: Fixed in 8feb2ffea452ae1b434837d7ab469d9a8d1bdfed

PR #827, opened 2020-08-06
Show additional context on Jenkins job diffs.

clalancette:
When trying to debug the differences between an old and new
job, a single line diff may not give enough context as to
what actually changed.  There's no exactly correct number to
use here, but I've increased it to 5 lines before and after
as a compromise between being too verbose and giving enough
context.

@nuclearsandwich FYI, this allowed me to also debug the ci.ros2.org problem I was talking to you about earlier.
@dirk-thomas FYI as the original author of this piece of code.

### DIFF
--- a/ros_buildfarm/jenkins.py
+++ b/ros_buildfarm/jenkins.py
@@ -278,7 +278,7 @@ def _diff_configs(remote_config, new_config):
     lines2 = xml2.splitlines()
 
     return difflib.unified_diff(
-        lines1, lines2, 'remote config', 'new config', n=0)
+        lines1, lines2, 'remote config', 'new config', n=5)
 
 
 def remove_jobs(jenkins, job_prefix, excluded_job_names, dry_run=False):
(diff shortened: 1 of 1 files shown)


dirk-thomas:
I always found no context line sufficient to determine if the diff is desired / expected or not. Since the output can already be extremely verbose without any context I would rather not add 10 lines to each diff part by default. How about making this a command line option to choose an arbitrary higher context count when desired?

clalancette:
> I always found no context line sufficient to determine if the diff is desired / expected or not. Since the output can already be extremely verbose without any context I would rather not add 10 lines to each diff part by default. How about making this a command line option to choose an arbitrary higher context count when desired?

That seems reasonable.  Let me take a look at doing that.

clalancette:
All right, 9885132 makes diff_lines a kwarg that can be passed in.  I'll open up a separate PR to https://github.com/ros2/ci to add a command-line argument to create_jenkins_job.py and pass that number in.

dirk-thomas:
LGTM.

One nitpick: I would suggest to name the argument `context_lines` instead.

clalancette:
> One nitpick: I would suggest to name the argument `context_lines` instead.

Sounds good, I've switched to that now.

PR #828, opened 2020-08-14
use distinct xunit pattern for ROS 2

dirk-thomas:
This redoes #723 and #724 (which were reverted in #725) and also adds a specific pattern for the xUnit type `CTestType` since CTest result files are available now (see colcon/colcon-cmake#89).

Without this change Jenkins tries to interpret the newly copied CTest result files as GoogleTest results files and chokes, e.g.:
* [Build Status](http://build.ros2.org/view/Rdev/job/Rdev__launch__ubuntu_focal_amd64/22/)
* [Build Status](http://build.ros2.org/job/Rpr__rosidl_typesupport__ubuntu_focal_amd64/20/)
* [Build Status](http://build.ros2.org/job/Fpr__message_filters__ubuntu_focal_amd64/3/)

With this patch:
* [Build Status](http://build.ros2.org/view/Rdev/job/Rdev__launch__ubuntu_focal_amd64/23/)
* [Build Status](http://build.ros2.org/job/Rpr__rosidl_typesupport__ubuntu_focal_amd64/23/)
* [Build Status](http://build.ros2.org/job/Fpr__message_filters__ubuntu_focal_amd64/4/)

I assume this works now because both buildfarms have been updated to the latest xUnit plugin version.

### DIFF
--- a/ros_buildfarm/ci_job.py
+++ b/ros_buildfarm/ci_job.py
@@ -303,7 +303,8 @@ def _get_ci_job_config(
 
         # only Ubuntu Focal has a new enough pytest version which generates
         # JUnit compliant result files
-        'xunit_publisher_types': get_xunit_publisher_types_and_patterns(),
+        'xunit_publisher_types': get_xunit_publisher_types_and_patterns(
+            ros_version, os_name == 'ubuntu' and os_code_name != 'bionic'),
     }
     job_config = expand_template(template_name, job_data)
     return job_config
--- a/ros_buildfarm/common.py
+++ b/ros_buildfarm/common.py
@@ -600,9 +600,21 @@ def get_packages_in_workspaces(workspace_roots, condition_context=None):
     return pkgs
 
 
-def get_xunit_publisher_types_and_patterns():
+def get_xunit_publisher_types_and_patterns(
+    ros_version, pytest_junit_compliant
+):
     types = []
-    types.append(('GoogleTestType', 'ws/test_results/**/*.xml'))
+    if ros_version == 1:
+        types.append(('GoogleTestType', 'ws/test_results/**/*.xml'))
+    elif ros_version == 2:
+        types.append(('CTestType', 'ws/test_results/*/Testing/*/Test.xml'))
+        types.append(('GoogleTestType', 'ws/test_results/**/*.gtest.xml'))
+        types.append((
+            'JUnitType' if pytest_junit_compliant else 'GoogleTestType',
+            'ws/test_results/*/pytest.xml'))
+        types.append(('JUnitType', 'ws/test_results/**/*.xunit.xml'))
+    else:
+        assert False, 'Unsupported ROS version: ' + str(ros_version)
     return types
 
 
--- a/ros_buildfarm/devel_job.py
+++ b/ros_buildfarm/devel_job.py
@@ -439,7 +439,8 @@ def _get_devel_job_config(
 
         # only Ubuntu Focal has a new enough pytest version which generates
         # JUnit compliant result files
-        'xunit_publisher_types': get_xunit_publisher_types_and_patterns(),
+        'xunit_publisher_types': get_xunit_publisher_types_and_patterns(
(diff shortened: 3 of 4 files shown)


ros_buildfarm/ci_job.py
```diff
@@ -303,7 +303,8 @@ def _get_ci_job_config(
 
         # only Ubuntu Focal has a new enough pytest version which generates
         # JUnit compliant result files
-        'xunit_publisher_types': get_xunit_publisher_types_and_patterns(),
+        'xunit_publisher_types': get_xunit_publisher_types_and_patterns(
+            ros_version, os_name == 'ubuntu' and os_code_name != 'bionic'),
```
j-rivero, line 307: to be junit compliant, does it depends on a given platform or depends on a certain pytest version?

ros_buildfarm/ci_job.py
```diff
@@ -303,7 +303,8 @@ def _get_ci_job_config(
 
         # only Ubuntu Focal has a new enough pytest version which generates
```
j-rivero, line 304: this comment probably needs to be updated given that the logic for `pytest_junit_compliant` involves bionic.

ros_buildfarm/ci_job.py
dirk-thomas: As the comment above this line states:

> only Ubuntu Focal has a new enough pytest version which generates JUnit compliant result files

ros_buildfarm/ci_job.py
dirk-thomas: I don't see how the comment should change? The `pytest` version as of Focal is new enough and since we don't want to keep enumerating newer distros as the appear the condition checks for the opposite - not being Bionic - which is the only older Ubuntu distro used and older than Focal.

j-rivero: the main point here that changes the whole thing is that somehow the code seems to affect only the following platforms: Bionic, Focal and new releases after Focal. I was not able to find where this context occur in the code.

dirk-thomas: As far as I can tell the condition exactly implements what the comment describes. If there is no concrete proposal what to change I will go ahead and merge this as-is.

dirk-thomas:
I assume the current patch resolves the issue for Rolling and Foxy.

For Eloquent this needs more work: e.g. [Build Status](http://build.ros2.org/job/Epr__rclcpp__ubuntu_bionic_amd64/183/)

ros_buildfarm/common.py
```diff
@@ -600,9 +600,21 @@ def get_packages_in_workspaces(workspace_roots, condition_context=None):
...
+):
     types = []
-    types.append(('GoogleTestType', 'ws/test_results/**/*.xml'))
+    if ros_version == 1:
+        types.append(('GoogleTestType', 'ws/test_results/**/*.xml'))
+    elif ros_version == 2:
+        types.append(('CTestType', 'ws/test_results/*/Testing/*/Test.xml'))
+        types.append(('GoogleTestType', 'ws/test_results/**/*.gtest.xml'))
+        types.append((
+            'JUnitType' if pytest_junit_compliant else 'GoogleTestType',
```
jacobperron, line 613: It's not super clear in the [pytest documentation](https://docs.pytest.org/en/latest/deprecations.html#junit-family-default-value-change-to-xunit2) or in the deprecation warnings produced, but testing locally I've observed that using pytest version >= 6 does **not** imply that the junit_family type will be xunit2. In fact, if you don't explicitly set the junit_family (either with a `pytest.ini` file or a pytest `-o` option), then we get the legacy behavior. This is what's causing CI to fail for Foxy (https://github.com/ros2/ros2/issues/1016).

I think this only happened to work for Rolling because of all of the backports explicitly setting the junit_family in pytest.ini files (https://github.com/ros2/ros2/issues/951).

jacobperron: See https://github.com/ros-infrastructure/ros_buildfarm/pull/836 for a proposed fix.

PR #836, opened 2020-08-27
Set junit_family=xunit2 for pytest results in Foxy and older distros

jacobperron:
This is to satisfy the Jenkins xunit plugin.

Connects to https://github.com/ros2/ros2/issues/1016.

Similar to what we've done for ci.ros2.org in https://github.com/ros2/ci/pull/503.

### DIFF
--- a/ros_buildfarm/workspace.py
+++ b/ros_buildfarm/workspace.py
@@ -86,7 +86,9 @@ def call_build_tool(
         if colcon_verb == 'test':
             cmd += [
                 '--event-handlers', 'console_direct+',
-                '--executor', 'sequential']
+                '--executor', 'sequential',
+                # Force junit_family to make Jenkins xunit plugin 2.x happy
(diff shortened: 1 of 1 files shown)


ros_buildfarm/workspace.py
```diff
@@ -86,7 +86,9 @@ def call_build_tool(
         if colcon_verb == 'test':
             cmd += [
                 '--event-handlers', 'console_direct+',
-                '--executor', 'sequential']
+                '--executor', 'sequential',
+                # Force junit_family to make Jenkins xunit plugin 2.x happy
+                '--pytest-args', '-o', 'junit_family=xunit2']
```
dirk-thomas, line 91: I don't think we should set this unconditionally. This should mimick the logic from ros2/ci#503 imo.

jacobperron: Good point, see 9a69528

PR #842, opened 2020-10-29
Use python3 when invoking reprepro-updater

nuclearsandwich:
reprepro-updater's #! lines default to /usr/bin/env python and the build templates invoke all reprepro-updater scripts with a Python 2 interpreter. The first draft of this change changes this and runs reprepro-updater scripts with python3.

This worked fine on a test server until reaching the sync jobs where there is currently some code that isn't python3 compatible. I am leaving this PR in draft until the `refactor` branch of reprepro-updater is compatible with both python and python3 so we can continue to use it for the existing build farms and the new 20.04-backed ones.


nuclearsandwich:
with https://github.com/ros-infrastructure/reprepro-updater/pull/84 and https://github.com/ros-infrastructure/reprepro-updater/pull/85 reprepro-updater is now working with python3. I don't think there is anything stopping existing buildfarm deployments from taking this change as long as they're using the latest `refactor` branch of reprepro-updater alongside the latest ros_buildfarm.

cottsay (approved):
Does `ros_buildfarm` always target the tip of `reprepro-updater`, or do we need to update a pinned ref somewhere?

nuclearsandwich:
> Does `ros_buildfarm` always target the tip of `reprepro-updater`, or do we need to update a pinned ref somewhere?

Yeah it always pulls the latest.
Forgive me for not taking the last step of linking the lines on GitHub but the refactor branch is hard-coded in the job templates and fetched with the git wrapper script for each job:

rg -F refactor
ros_buildfarm/templates/release/deb/sync_packages_to_main_job.xml.em
28:    branch_name='refactor',

ros_buildfarm/templates/release/deb/sync_packages_to_testing_job.xml.em
99:        'python3 -u $WORKSPACE/ros_buildfarm/scripts/wrapper/git.py clone --depth 1 -b refactor https://github.com/ros-infrastructure/reprepro-updater.git reprepro-updater',

ros_buildfarm/templates/release/deb/import_package_job.xml.em
41:    branch_name='refactor',

ros_buildfarm/templates/release/deb/import_upstream_job.xml.em
49:    branch_name='refactor',

ros-discourse:

https://discourse.ros.org/t/changes-to-ros-buildfarm-affecting-repository-import-jobs/17047/1

PR #846, opened 2020-12-07
Make the doc_independent_docker_build a little more generic.

clalancette:
In particular:

1.  Allow the user to specify the branch to checkout on the
source repository.
2.  Allow the doc_independent_docker_build to upload either
via the publish-over-git publisher, or via an rsync script.

This functionality is needed for uploading ros2_documentation to docs.ros.org.  I've tested this functionality up to the upload sequence; the test buildfarm I am working on doesn't actually have the correct credentials, so I can't do an end-to-end test.  I'll point out a few things of interest inline.

### DIFF
--- a/ros_buildfarm/config/doc_build_file.py
+++ b/ros_buildfarm/config/doc_build_file.py
@@ -179,6 +179,12 @@ class DocBuildFile(BuildFile):
 
             self.doc_repository_branch = data.get('doc_repository_branch', None)
 
+        # If neither upload location is specified, fall back to a default of
+        # an upload_host of 'repo', which is valid for the publish-over-ssh
+        # configuration.
+        if self.upload_host is None and self.upload_repository_url is None:
+            self.upload_host = 'repo'
+
         # Ensure that we have one, and only one, place to upload to
         assert (self.upload_host is not None) ^ (self.upload_repository_url is not None)
 
--- a/ros_buildfarm/config/doc_build_file.py
+++ b/ros_buildfarm/config/doc_build_file.py
@@ -167,14 +167,21 @@ class DocBuildFile(BuildFile):
         # user host and docroot have default of uploading to the repo machine
         # next to the debs
         self.upload_user = data.get('upload_user', 'jenkins-agent')
-        self.upload_host = data.get('upload_host', 'repo')
+        self.upload_host = data.get('upload_host', None)
         self.upload_root = data.get('upload_root', '/var/repos/docs')
+
+        self.upload_repository_url = None
         if self.documentation_type == DOC_TYPE_DOCKER:
-            assert 'upload_repository_url' in data
-            self.upload_repository_url = data['upload_repository_url']
+            self.upload_repository_url = data.get('upload_repository_url', None)
             self.upload_repository_branch = data.get(
                 'upload_repository_branch', 'gh-pages'
             )
+
+            self.doc_repository_branch = data.get('doc_repository_branch', None)
+
+        # Ensure that we have one, and only one, place to upload to
+        assert (self.upload_host is not None) ^ (self.upload_repository_url is not None)
+
         assert 'upload_credential_id' in data
         self.upload_credential_id = data['upload_credential_id']
 
--- a/ros_buildfarm/config/doc_build_file.py
+++ b/ros_buildfarm/config/doc_build_file.py
@@ -74,10 +74,22 @@ class DocBuildFile(BuildFile):
             self.canonical_base_url = data['canonical_base_url']
             assert not self.canonical_base_url or is_rosdoc_type
 
-        self.doc_repositories = []
+        self.doc_repositories = {}
         if 'doc_repositories' in data:
-            self.doc_repositories = data['doc_repositories']
-            assert isinstance(self.doc_repositories, list)
+            doc_repos = data['doc_repositories']
+            if isinstance(doc_repos, list):
+                for url in doc_repos:
+                    self.doc_repositories[url] = ''
+            elif isinstance(doc_repos, dict):
+                for repo in doc_repos:
+                    assert 'url' in doc_repos[repo]
+                    url = doc_repos[repo]['url']
+                    if 'branch' in doc_repos[repo]:
+                        self.doc_repositories[url] = doc_repos[repo]['branch']
+                    else:
+                        self.doc_repositories[url] = ''
+            else:
+                assert isinstance(doc_repos, list) or isinstance(doc_repos, dict)
 
         # doc_repositories can only be used with make_target and docker_build
         # doc types
@@ -176,8 +188,6 @@ class DocBuildFile(BuildFile):
             self.upload_repository_branch = data.get(
                 'upload_repository_branch', 'gh-pages'
             )
-
-            self.doc_repository_branch = data.get('doc_repository_branch', None)
 
         # If neither upload location is specified, fall back to a default of
         # an upload_host of 'repo', which is valid for the publish-over-ssh
--- a/ros_buildfarm/doc_job.py
+++ b/ros_buildfarm/doc_job.py
@@ -427,6 +427,12 @@ def _get_doc_independent_job_config(
         'notify_emails': build_file.notify_emails,
 
         'timeout_minutes': build_file.jenkins_job_timeout,
+
+        'upload_user': build_file.upload_user,
+        'upload_host': build_file.upload_host,
+        'upload_root': build_file.upload_root,
+
+        'credential_id': build_file.upload_credential_id
     }
 
     if build_file.documentation_type == 'make_target':
@@ -434,17 +440,13 @@ def _get_doc_independent_job_config(
         job_data.update({
             'install_apt_packages': build_file.install_apt_packages,
             'install_pip_packages': build_file.install_pip_packages,
-            'upload_user': build_file.upload_user,
-            'upload_host': build_file.upload_host,
-            'upload_root': build_file.upload_root,
-            'credential_id': build_file.upload_credential_id
         })
     elif build_file.documentation_type == 'docker_build':
         template_name = 'doc/doc_independent_docker_job.xml.em'
         job_data.update({
+            'doc_repository_branch': build_file.doc_repository_branch,
             'upload_repository_url': build_file.upload_repository_url,
             'upload_repository_branch': build_file.upload_repository_branch,
-            'upload_credential_id': build_file.upload_credential_id,
         })
     else:
         raise JobValidationError(
--- a/ros_buildfarm/doc_job.py
+++ b/ros_buildfarm/doc_job.py
@@ -444,7 +444,6 @@ def _get_doc_independent_job_config(
     elif build_file.documentation_type == 'docker_build':
         template_name = 'doc/doc_independent_docker_job.xml.em'
         job_data.update({
-            'doc_repository_branch': build_file.doc_repository_branch,
             'upload_repository_url': build_file.upload_repository_url,
             'upload_repository_branch': build_file.upload_repository_branch,
         })
--- a/ros_buildfarm/doc_job.py
+++ b/ros_buildfarm/doc_job.py
@@ -432,7 +432,7 @@ def _get_doc_independent_job_config(
         'upload_host': build_file.upload_host,
         'upload_root': build_file.upload_root,
 
-        'credential_id': build_file.upload_credential_id
+        'upload_credential_id': build_file.upload_credential_id
     }
 
     if build_file.documentation_type == 'make_target':
--- a/ros_buildfarm/templates/doc/doc_independent_docker_job.xml.em
+++ b/ros_buildfarm/templates/doc/doc_independent_docker_job.xml.em
@@ -89,15 +89,15 @@ import os
 doc_repository_name = os.path.splitext(os.path.basename(doc_repository_url))[0]
 
 if doc_repository_branch is None:
-  doc_repository_branch = ''
+  repo_branch_arg = ''
 else:
-  doc_repository_branch = '--no-single-branch -b ' + doc_repository_branch
+  repo_branch_arg = '--no-single-branch -b ' + doc_repository_branch
 }@
 @(SNIPPET(
     'builder_shell',
     script='\n'.join([
         'echo "# BEGIN SECTION: Clone %s"' % doc_repository_name,
-        'python3 -u $WORKSPACE/ros_buildfarm/scripts/wrapper/git.py clone --depth 1 %s %s $WORKSPACE/repositories/%s' % (doc_repository_url, doc_repository_branch, doc_repository_name),
+        'python3 -u $WORKSPACE/ros_buildfarm/scripts/wrapper/git.py clone --depth 1 %s %s $WORKSPACE/repositories/%s' % (doc_repository_url, repo_branch_arg, doc_repository_name),
         'git -C $WORKSPACE/repositories/%s log -n 1' % doc_repository_name,
         'echo "# END SECTION"',
     ]),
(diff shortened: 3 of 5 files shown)


ros_buildfarm/config/doc_build_file.py
```diff
@@ -167,14 +167,21 @@ def __init__(self, name, data):  # noqa: D107
         # user host and docroot have default of uploading to the repo machine
         # next to the debs
         self.upload_user = data.get('upload_user', 'jenkins-agent')
-        self.upload_host = data.get('upload_host', 'repo')
+        self.upload_host = data.get('upload_host', None)
```
clalancette, line 170: Note that I'm changing from a default of `repo` (which was bogus anyway), to a default of `None`.

ros_buildfarm/config/doc_build_file.py
```diff
@@ -167,14 +167,21 @@ def __init__(self, name, data):  # noqa: D107
...
+        self.upload_repository_url = None
         if self.documentation_type == DOC_TYPE_DOCKER:
-            assert 'upload_repository_url' in data
-            self.upload_repository_url = data['upload_repository_url']
+            self.upload_repository_url = data.get('upload_repository_url', None)
             self.upload_repository_branch = data.get(
                 'upload_repository_branch', 'gh-pages'
             )
+
+            self.doc_repository_branch = data.get('doc_repository_branch', None)
```
clalancette, line 180: This is a new user-visible configuration parameter.  It allows the user to specify what branch to checkout on the source documentation repository.  I know I'll need to document this, but I wanted to make sure that adding something like this is OK before going further.

ros_buildfarm/config/doc_build_file.py
```diff
@@ -167,14 +167,21 @@ def __init__(self, name, data):  # noqa: D107
...
+
+        # Ensure that we have one, and only one, place to upload to
+        assert (self.upload_host is not None) ^ (self.upload_repository_url is not None)
```
clalancette, line 183: We assert that only one of `upload_host` or `upload_repository_url` is specified.

ros_buildfarm/templates/doc/doc_independent_docker_job.xml.em
```diff
@@ -81,14 +87,18 @@
 @{
 import os
 doc_repository_name = os.path.splitext(os.path.basename(doc_repository_url))[0]
+
+if doc_repository_branch is None:
+  doc_repository_branch = ''
+else:
+  doc_repository_branch = '--no-single-branch -b ' + doc_repository_branch
```
clalancette, line 94: For the current work on ros2_documentation, it uses the `sphinx-multiversion` project.  It works by taking a list of branches from `conf.py`, checking out each of the branches in turn and running sphinx against each one.  So we need *all* of the branches available for this to work.  This is also why I removed the `rm -fr .git` below.

ros_buildfarm/config/doc_build_file.py
nuclearsandwich: `repo` is a valid SSH host for the Jenkins publish-over-ssh configuration and corresponds to the host running the building_repository agent. Since this agent hosts an http server it's used as the default docs host for custom build farms.

I expect that this default is consistent across most of the doc jobs and would suggest either not changing it or changing it everywhere.

ros_buildfarm/config/doc_build_file.py
nuclearsandwich: My primary concern is that doc_repositories is a list of repositories while the proposed branch option is singular.

The naive solution would be to follow the pattern of repository urls and keys and make `doc_repository_branches` an array where `url[0]` corresponds to `branch[0]` but I don't care much for the obfuscated connection there and would much rather support a compound data structure such as `doc_repositories:{url:,branch:} or even adopting the full `doc_repositories:{type:,url:,version:}` structure for repositories that is used in rosdistro distribution.yaml. I'd value @cottsay's opinion on the matter as well since the fact that this peeves me may mean it weighs more than it should in my review.

ros_buildfarm/doc_job.py
```diff
@@ -427,24 +427,26 @@ def _get_doc_independent_job_config(
         'notify_emails': build_file.notify_emails,
 
         'timeout_minutes': build_file.jenkins_job_timeout,
+
+        'upload_user': build_file.upload_user,
+        'upload_host': build_file.upload_host,
+        'upload_root': build_file.upload_root,
+
+        'credential_id': build_file.upload_credential_id
```
nuclearsandwich, line 435: I haven't done a commit history dive to check whether accident or intent had the credential variable inserted based on specific job types but if we're going to normalize it I would prefer to keep `upload_credential_id` so that a grep through the codebase turns up more direct hits.

If updating the make_target job is too cumbersome for this patch I would be willing to write one which does so separately.

ros_buildfarm/templates/doc/doc_independent_docker_job.xml.em
```diff
@@ -81,14 +87,18 @@
...
 }@
 @(SNIPPET(
     'builder_shell',
     script='\n'.join([
         'echo "# BEGIN SECTION: Clone %s"' % doc_repository_name,
-        'python3 -u $WORKSPACE/ros_buildfarm/scripts/wrapper/git.py clone --depth 1 %s $WORKSPACE/repositories/%s' % (doc_repository_url, doc_repository_name),
+        'python3 -u $WORKSPACE/ros_buildfarm/scripts/wrapper/git.py clone --depth 1 %s %s $WORKSPACE/repositories/%s' % (doc_repository_url, doc_repository_branch, doc_repository_name),
```
nuclearsandwich, line 100: Don't you need a `-b` flag before the branch name?

ros_buildfarm/templates/doc/doc_independent_docker_job.xml.em
```diff
@@ -123,6 +133,22 @@ doc_repository_name = os.path.splitext(os.path.basename(doc_repository_url))[0]
...
+        '  ssh %s@%s "mkdir -p %s"' %
+          (upload_user, upload_host, os.path.join(upload_root, 'build', 'html')),
+        '  cd $WORKSPACE/repositories/%s/build' % (doc_repository_name),
+        '  rsync -e ssh --stats -r --delete html %s@%s:%s' % \
+          (upload_user, upload_host, os.path.join(upload_root, 'build/html')),
+        '  echo "# END SECTION"',
+        'fi',
+    ]),
+))@
+@[end if]@
```
nuclearsandwich, line 151: This is all new here because this job previously only implemented support for `upload_repository_url` right?

ros_buildfarm/config/doc_build_file.py
clalancette: Oh, OK.  Sorry, that is not clear at all here.

Well, the main reason I changed it is so that I can do error checking down below to make sure only one of `upload_host` and `upload_repository_url` was specified per-job.  I guess I can change the logic here so that if *neither* is specified, it falls back to an `upload_host` of "repo".  Does that make sense?

ros_buildfarm/config/doc_build_file.py
clalancette: Yes, that actually annoyed me too.  And your concern about the obfuscated connection there is exactly why I didn't go that route.

I'd be for making a compound structure like you mention above, but I couldn't quite see how to do it in a backwards-compatible way.  But that may just be my unfamiliarity with YAML.  If someone could suggest a structure that would be backwards-compatible, I'd be happy to implement it that way.

ros_buildfarm/doc_job.py
clalancette: I arbitrarily chose `credential_id`, but I'm happy to switch it to `upload_credential_id`.  It is straightforward to do.  I'll go ahead and do that.

ros_buildfarm/templates/doc/doc_independent_docker_job.xml.em
clalancette: That gets added in the code above.  That is, if `doc_repository_branch` is *not* specified, then the `doc_repository_branch` string is empty and we get the default branch.  If `doc_repository_branch` *is* specified, then the `doc_repository_branch` string is `--no-single-branch -b <branch>`, and we get the requested branch.

ros_buildfarm/templates/doc/doc_independent_docker_job.xml.em
clalancette: Exactly.

ros_buildfarm/config/doc_build_file.py
clalancette: I went ahead and did this in 4c374cf12ffbf265b248771fc9d9fc0c2dc38e0e

ros_buildfarm/config/doc_build_file.py
clalancette: Looking at this closer, I could do the following.  If the `doc_repositories` section looks like:

```
doc_repositories:
- https://github.com/clalancette/ros2_documentation.git
```

Then the Python type will be list.  If it looks more like:

```
doc_repositories:
  ros2_doc:
    url: https://github.com/clalancette/ros2_documentation.git
    branch: clalancette/rolling-add-multiversion-target
```

Then the Python type will be dict.  So I can do something like the following in the Python code:

```
repos = {}
if isinstance(data['doc_repositories'], list):
    for url in data['doc_repositories']:
        repos[url] = ''
elif isinstance(data['doc_repositories'], dict):
    for repo in data['doc_repositories']:
        assert 'url' in data['doc_repositories'][repo]
        url = data['doc_repositories'][repo]['url']
        if 'branch' in data['doc_repositories'][repo]:
            repos[url] = data['doc_repositories'][repo]['branch']
        else:
            repos[url] = ''
```

Which would always get me a dictionary from 'url' -> 'branch' (and where the branch name may be empty, indicating use the default branch).  What do you think about that?

ros_buildfarm/templates/doc/doc_independent_docker_job.xml.em
nuclearsandwich: I'm somewhat leery of overwriting the `doc_repository_branch` as it means that once we've modified it to create this argument we no longer have access to the "raw" branch name should we need it. But consider this a nitpick.

clalancette: Yeah, that's fine.  I can rename the variable to something else.

cottsay (commented):
I don't have any real concerns here. Nothing that would need a staged rollout as long as the config change already being discussed gets implemented.

ros_buildfarm/templates/doc/doc_independent_docker_job.xml.em
```diff
@@ -81,14 +87,18 @@
...
+  doc_repository_branch = '--no-single-branch -b ' + doc_repository_branch
 }@
 @(SNIPPET(
     'builder_shell',
     script='\n'.join([
         'echo "# BEGIN SECTION: Clone %s"' % doc_repository_name,
-        'python3 -u $WORKSPACE/ros_buildfarm/scripts/wrapper/git.py clone --depth 1 %s $WORKSPACE/repositories/%s' % (doc_repository_url, doc_repository_name),
+        'python3 -u $WORKSPACE/ros_buildfarm/scripts/wrapper/git.py clone --depth 1 %s %s $WORKSPACE/repositories/%s' % (doc_repository_url, doc_repository_branch, doc_repository_name),
         'git -C $WORKSPACE/repositories/%s log -n 1' % doc_repository_name,
-        'rm -fr $WORKSPACE/repositories/%s/.git' % doc_repository_name,
```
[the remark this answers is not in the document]
clalancette: This has been there since the doc_independent_docker_job was added.  I can only guess that this was a copy from the doc_independent_job, which does the same thing.  Maybe it is an attempt to limit the number of Docker layers?  I don't really know.

ros_buildfarm/templates/doc/doc_independent_docker_job.xml.em
```diff
@@ -123,6 +133,22 @@ doc_repository_name = os.path.splitext(os.path.basename(doc_repository_url))[0]
         'echo "# END SECTION"',
     ]),
 ))@
+@[if upload_host is not None]@
+@(SNIPPET(
+    'builder_shell',
+    script='\n'.join([
+        'if [ -d "$WORKSPACE/repositories/%s/build/html" ]; then' % (doc_repository_name),
```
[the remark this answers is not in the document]
clalancette: That's just habit for me; I always use the tuple form, that way it is easier to add new arguments later.  But I'll make it consistent with the rest of the code in here.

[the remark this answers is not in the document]
clalancette: Fixed in 1b81fcd

ros_buildfarm/templates/doc/doc_independent_docker_job.xml.em
clalancette: Fixed in 1b81fcd

ros_buildfarm/config/doc_build_file.py
clalancette: Just to be clear; I don't think this would break the API, so much as *add* to the API.

That is, if we go this way, then what worked in the YAML before (a list of doc_repositories) would still work.  It's just that we would additionally support having a dictionary of values in a particular way in the YAML.

I'm going to go ahead and prototype this and see what it ends up looking like.

clalancette: All right, I've coded this up and tested it on the test farm, and things seem to be good.  I've also documented it now.  So I'm going to mark this conversation resolved, and change this to a "real" PR.

clalancette:
All right, I think I've addressed all of the open comments.  You can see the job that results from using this code along with the configuration at https://github.com/ros-staging/ros_buildfarm_config/blob/staging/ros2-documentation-build.yaml here: https://build.test.ros2.org/job/doc_ros2-documentation/

@nuclearsandwich @cottsay please take another look when you get a chance.  Thanks.

ros_buildfarm/templates/doc/doc_independent_docker_job.xml.em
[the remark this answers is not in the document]
nuclearsandwich: It does occur to me that some of the tools which recurse over projects may stray into .git metadata if left unchecked but I'm okay giving this a shot and monitoring in production. https://github.com/ros-infrastructure/ros_buildfarm/commit/c204239b08fdec35c2e6237fdde4bba7504f382e is the commit which introduced this for the doc_independent job and appears to be part of a generic cleanup approach. The commit message does not indicate a motivation.

nuclearsandwich (approved):

ros_buildfarm/templates/doc/doc_independent_docker_job.xml.em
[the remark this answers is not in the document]
clalancette: That's true, I think we'll have to monitor it.  The good news is that this job is used for very few jobs currently (only index at the moment, I believe), so any breakage will be limited in scope.

ros_buildfarm/templates/doc/doc_independent_docker_job.xml.em
```diff
@@ -132,10 +158,12 @@ doc_repository_name = os.path.splitext(os.path.basename(doc_repository_url))[0]
     dynamic_recipients=[],
     send_to_individuals=False,
 ))@
+@[if upload_repository_url is not None]@
```
cottsay, line 161: Just to make sure, should this be `upload_repository_branch`?

clalancette: Yes and no.  The code in https://github.com/ros-infrastructure/ros_buildfarm/pull/846/files#diff-2f3d0521af79f10d17016a49a17b0007a0cd3510e58bbc7f2de273ec30ca2e5cR187 ensures that if there is an `upload_repository_url`, there is also guaranteed to be an `upload_repository_branch`.  In order to match this snippet with the snippet above at https://github.com/ros-infrastructure/ros_buildfarm/pull/846/files#diff-f5faccbba5d43185cbbd3c046bad114c25aa1782f9a0b11c07dd6ac6939bb6e4R27, I decided to use the same conditional for both.

Essentially, it is a trade-off in readability: we can either make the two snippets match (and hence make it easier to discover that they "go together"), or we can check the variables we use in a particular snippet (making it a bit safer, but also losing the property that these go together).

cottsay (approved):

clalancette:
All right, going ahead and merging this.  Thanks for the reviews.

PR #854, opened 2021-02-09
Fix a few minor issues in the doc_independent_docker_job.

clalancette:
* Don't create a build/html on the destination; that's the
source directory structure.
* Upload just the contents of the local html (essentially,
add a / to the rsync source entry)
* Make sure to setup the credentials properly when we are
setting up for an upload.

I've tested all of these by configuring https://build.ros.org/job/doc_ros2doc/ "by hand" and then testing out the resulting build.


tfoote (approved):

clalancette:
Travis failed because of docker pull-rate-limiting.  So I'm going to go ahead and merge and deploy this, thanks for the revew.

PR #856, opened 2021-02-19
add graphviz as a dependency necessary for sphinx in rosdoc_lite

tfoote:
Working to resolve: https://github.com/ros/geometry2/issues/498

I think the dependency used to come in for free from something indirectly.


MatthijsBurgh:
Yes it is used by rosdoc_lite for inheritance diagrams in doxygen generated docs. I should definetly be added to rosdoc_lite as dependency. I am not sure, it still requires to be added here, I am don't know how dependencies are installed in the buildfarm.

cottsay (approved):
This looks fine to me if it is still needed. Thanks @tfoote.

PR #876, opened 2021-05-04
Add --shared-ccache argument to generate_prerelease_script.py

ipa-mdl:
This PR will set `shared_ccache` from the command line in generate_prerelease_script.py.

Since this script is supposed to be run locally by users, the build file setting will always be overwritten.


ipa-mdl:
Open issues:

- Is it okay to always overwrite the build file setting?
- Should I patch all *_scripts.py?
- Should the ccache directory (`$HOME/.ccache`) be configurable? (as a replacement for #582)

PR #881, opened 2021-05-13
remove old directory structure that was a reference to an unused repo

tfoote:
I'm going to archive the repo too. https://github.com/ros-infrastructure/rosindex_containers


nuclearsandwich (approved):

nuclearsandwich:
Amusingly enough, this change did not deploy to the rosdoc2 job until three years later, and when it was, it introduced a breakage since there is no longer any cleanup of the cidfile which was previously removed when recursively removing the docker_containers directory.

PR #883, opened 2021-05-17
Update version of dashboard-view plugin.

nuclearsandwich:
Companion to ros-infrastructure/cookbook-ros-buildfarm#94. The other plugins are not referenced directly in our configuration templates.


j-rivero (approved):
mainly cosmetic changes here, if I'm not wrong.

Issue #878, opened 2021-05-11
13 jobs still use EOL xenial base images


tfoote:
Following #877 

## Already Failing
* [x] doc_independent #877 
* [ ] doc_rosindex
  * [x] https://github.com/ros2/ros2_documentation/pull/1518
  * [x] https://github.com/ros-infrastructure/rosindex/pull/268
  * [ ] https://github.com/hidmic/sphinx-tabs/pull/1

## Not yet failing
* [x] ci_create_reconfigure_task https://github.com/ros-infrastructure/ros_buildfarm/pull/906
* [x] devel_create_reconfigure_task https://github.com/ros-infrastructure/ros_buildfarm/pull/906
* [x] doc_create_reconfigure_task https://github.com/ros-infrastructure/ros_buildfarm/pull/884
* [x] doc_metadata_task https://github.com/ros-infrastructure/ros_buildfarm/pull/884
* [x] rosdistro_cache_task https://github.com/ros-infrastructure/ros_buildfarm/pull/886
* [x] release_check_sync_criteria_task https://github.com/ros-infrastructure/ros_buildfarm/pull/886
* [x] release_create_reconfigure_task https://github.com/ros-infrastructure/ros_buildfarm/pull/886
* [x] release_create_trigger_task https://github.com/ros-infrastructure/ros_buildfarm/pull/886
* [x] blocked_releases_page_task https://github.com/ros-infrastructure/ros_buildfarm/pull/885
* [x] blocked_source_entries_page_task https://github.com/ros-infrastructure/ros_buildfarm/pull/885
* [x] release_compare_page_task https://github.com/ros-infrastructure/ros_buildfarm/pull/885
* [x] release_status_page_task https://github.com/ros-infrastructure/ros_buildfarm/pull/885

nuclearsandwich:
@tfoote Propose deprecation note hidmic/sphinx-tabs#1 is the only outstanding TODO on this issue.
Since the rosindex job is running (now no longer on Xenial) is that work still required before we can close this one out?

tfoote:
Yeah we can close this out now. The sphinx-tabs us removed now: https://github.com/ros-infrastructure/rosindex/pull/268 so the  deprecation note is valuable but not necessary for our purposes now.

PR #884, opened 2021-05-17
Update container image base for doc-related container templates.

nuclearsandwich:
Partially addressing #878

doc_create_reconfigure_task
---------------------------

This container is used to bootstrap the creation of doc reconfigure
jobs. Changes here aren't likely to be high impact since ros_buildfarm
is tested on Focal's python3 version by CI and our local workstations.

doc_independent_task
--------------------

Container for running doc independent jobs. There's a high likelihood
for churn here but the task is currently failing due to changes in pip
dependencies.

doc_metadata_task
-----------------

It does not appear that we have any current doc_metadata jobs on the
build farm. Nor does this job appear tested so this change will not
break anything immediately but it may break if doc_metadata jobs are
reintroduced.

### DIFF
--- a/ros_buildfarm/templates/doc/doc_create_reconfigure_task.Dockerfile.em
+++ b/ros_buildfarm/templates/doc/doc_create_reconfigure_task.Dockerfile.em
@@ -1,6 +1,6 @@
 # generated from @template_name
 
-FROM ubuntu:xenial
+FROM ubuntu:focal
 
 VOLUME ["/var/cache/apt/archives"]
 
--- a/ros_buildfarm/templates/doc/doc_create_reconfigure_task.Dockerfile.em
+++ b/ros_buildfarm/templates/doc/doc_create_reconfigure_task.Dockerfile.em
@@ -1,6 +1,7 @@
 # generated from @template_name
 
-FROM ubuntu:focal
+@{os_code_name = 'focal'}
+FROM ubuntu:@os_code_name
 
 VOLUME ["/var/cache/apt/archives"]
 
@@ -18,7 +19,7 @@ RUN useradd -u @uid -l -m buildfarm
     distribution_repository_keys=distribution_repository_keys,
     distribution_repository_urls=distribution_repository_urls,
     os_name='ubuntu',
-    os_code_name='xenial',
+    os_code_name=os_code_name,
     add_source=False,
 ))@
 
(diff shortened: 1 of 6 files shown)


nuclearsandwich:
### Requesting feedback from other maintainers:

In one file I used a template variable to set the OS distribution once and use it everywhere in the template. 

1. Is that preferred compared to hard-coding the value repeatedly or is the hard-coding clearer due to the lack of magic? [(example)](https://github.com/ros-infrastructure/ros_buildfarm/pull/884/files#diff-c59a06d7639623ce9b1d29597ed567bfc1653eb969aea77d0add28453cb43ec0)
2. Is it still preferred to do even when there's only one instance of the hard-coded distribution name in the template?  [(example)](https://github.com/ros-infrastructure/ros_buildfarm/pull/884/files#diff-f5faccbba5d43185cbbd3c046bad114c25aa1782f9a0b11c07dd6ac6939bb6e4)
3. In files that aren't Dockerfiles with a `FROM ...` line does it make sense to set the variable just before first use or at the top of the template?

Please let me know what you think!

tfoote:
Using the template makes sense. I think that it should be used everywhere. It's a little bit of magic, but if we use it the same way everwhere it should be relatively clear. 

I thought about suggesting using Docker ARG to be less "magic" but I think the consistency makes more sense.

nuclearsandwich:
> Using the template makes sense. I think that it should be used everywhere. It's a little bit of magic, but if we use it the same way everwhere it should be relatively clear.

Thanks for the feedback. I'll go ahead and propagate the use of the template variable and pull this out of draft.

nuclearsandwich:
As a result of this change doc jobs will now be run using Sphinx 4.1.2 rather than 1.1.3 which has a significantly updated theme and presentation. An example of the updated documentation is currently on the staging farm here http://repo.test.ros2.org/docs/independent/api/catkin_pkg/html/index.html.

In addition to the content changes sphinx is also building these files under an html subdirectory instead of directly within the package name.

nuclearsandwich:
After discussing the changes to documentation in today's meeting offline we reached a consensus to move packages to the rtd theme. Since beginning that work I've realized that we're not very consistent _at all_ with respect to documentation theming within these packages and as a result I think we can de-couple updating the documentation themes and updating the base images. https://github.com/ros-infrastructure/ros_buildfarm_config/pull/211 updates the doc-independent configuration to account for changes between Xenial and Focal and I think they're the only two inter-dependent pull requests related to this change.

nuclearsandwich:
[Build Status](https://build.test.ros2.org/job/doc_independent/14/) is a staging build of doc_independent running with the updated apt and pip dependencies in https://github.com/ros-infrastructure/ros_buildfarm_config/pull/211

cottsay (approved):

tfoote (approved):

PR #886, opened 2021-05-19
Update base container images for release-related jobs.

nuclearsandwich:
Partially addresses #878

Run jobs in focal containers rather than xenial containers and use a template variable so that future updates to the base image are made in a single location per job type.

Affected jobs

* _rosdistro-cache
* _release-check
* _rel-sync-packages-
* _rel-reconfigure-jobs
* _rel-trigger-broken-with-non-broken-upstream
* _rel-trigger-jobs
* _rel-trigger-missed-jobs

## rosdistro_cache_task

Used to run the rosdistro_build_cache script. There doesn't appear to be a material difference between cache files generated on Xenial or Focal.

## release_check_sync_criteria_task

An issue was caused by not realizing that os_code_name was already in use as a parameter for the script target so my creation of an in-template os_code_name variable shadowed that value and hard-coded it to `focal`. 7d5af5e renames the in-template variable to task_os_code_name to distinguish it and allow the target os_code_name to pass through correctly to the command line invocation.

## create_reconfigure_task

There does not appear to be a difference between jobs run on Xenial or Focal.

## create_trigger_task

There does not appear to be a difference between jobs run on Xenial or Focal.

### DIFF
--- a/ros_buildfarm/templates/release/release_create_trigger_task.Dockerfile.em
+++ b/ros_buildfarm/templates/release/release_create_trigger_task.Dockerfile.em
@@ -1,6 +1,7 @@
 # generated from @template_name
 
-FROM ubuntu:xenial
+@{os_code_name = 'focal'}@
+FROM ubuntu:focal
 
 VOLUME ["/var/cache/apt/archives"]
 
@@ -18,7 +19,7 @@ RUN useradd -u @uid -l -m buildfarm
     distribution_repository_keys=distribution_repository_keys,
     distribution_repository_urls=distribution_repository_urls,
     os_name='ubuntu',
-    os_code_name='xenial',
+    os_code_name=os_code_name,
     add_source=False,
 ))@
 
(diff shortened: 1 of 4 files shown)


cottsay (approved):

ros_buildfarm/templates/release/release_create_trigger_task.Dockerfile.em
```diff
@@ -1,6 +1,7 @@
 # generated from @template_name
 
-FROM ubuntu:xenial
+@{os_code_name = 'focal'}@
+FROM ubuntu:focal
```
jacobperron, line 4: Did you mean?
```suggestion
FROM ubuntu:@os_code_name
```

nuclearsandwich: Hah. Ayup I did. My first pass at these updates involved just changing the hard-coded value and it was only later I created the in-template variable to make future changes easier. I clearly missed the fact that I'd already "touched" this line as I was scanning to replace `xenial` with the variable.

jacobperron (approved):

PR #936, opened 2022-01-27
Use agent_on_jenkins label for jobs which should run on that node.

nuclearsandwich:
The agent_on_master was renamed to agent_on_jenkins in the migration to
cookbook-ros-buildfarm and Ubuntu 20.04. To maintain compatibility the
cookbook still applies the label agent_on_master to the agent_on_jenkins
but since Jenkins upstream is updating the name of the built-in agent
this seems like a good time to update the labels in ROS build farm.

A future release of ROS Buildfarm can remove the old names and simplify
the label expressions again.

### DIFF
--- a/ros_buildfarm/templates/status/blocked_releases_page_job.xml.em
+++ b/ros_buildfarm/templates/status/blocked_releases_page_job.xml.em
@@ -30,7 +30,7 @@
     refspec=None,
 ))@
   <scmCheckoutRetryCount>2</scmCheckoutRetryCount>
-  <assignedNode>agent_on_master||buildagent</assignedNode>
+  <assignedNode>agent_on_master || buildagent</assignedNode>
   <canRoam>false</canRoam>
   <disabled>false</disabled>
   <blockBuildWhenDownstreamBuilding>false</blockBuildWhenDownstreamBuilding>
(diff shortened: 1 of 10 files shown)


ros_buildfarm/templates/status/blocked_releases_page_job.xml.em
```diff
@@ -30,7 +30,7 @@
     refspec=None,
 ))@
   <scmCheckoutRetryCount>2</scmCheckoutRetryCount>
-  <assignedNode>agent_on_master||buildagent</assignedNode>
+  <assignedNode>agent_on_jenkins || agent_on_master || buildagent</assignedNode>
```
tfoote, line 33: Have you confirmed that this whitespace is stripped before the string evaluation?

nuclearsandwich: What string evaluation are you referring to? The empy template process, Jenkins label expressions, or something else?

tfoote: The agent matching was what I was referring to.

tfoote (approved):
This is good to help us move forward on the rename.

PR #1010, opened 2023-09-28
Add display of cpu info

tfoote:
I had originally catted /proc/cpuinfo but it was very verbose

lscpu is in [util-linux](https://packages.ubuntu.com/focal/util-linux) which is in [essential] so should be available on systems.


nuclearsandwich (approved):
One tiny nitpick, I think it makes sense to put cpu info _before_ docker info rather than after but I'm willing to yield to other reviewers or a strong justification. My rationale is showing hardware info (less variable) before software info (more variable).

Other than that, no objections!

tfoote:
Moved to the front. It looks like CI is failing for all the python 2.7 builds

PR #1011, opened 2023-09-28
Remove python 2.7, melodic, and foxy, due to being unsupported anymore

tfoote:
Python 2.7 has been removed as an option from github actions as a base image causing all of these to fail. As all of our distros are on python 3 default platforms now we don't need to test this anymore.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -38,8 +38,8 @@ jobs:
       - name: Run job
         uses: ./.github/actions/audit
         with:
-          ros_distro: melodic
-          os_code_name: bionic
+          ros_distro: noetic
+          os_code_name: focal
 
   ros1_config:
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yaml
```diff
@@ -260,8 +260,8 @@ jobs:
         uses: ./.github/actions/audit
         with:
           config_url: https://raw.githubusercontent.com/ros2/ros_buildfarm_config/ros2/index.yaml
-          ros_distro: foxy
-          os_code_name: focal
+          ros_distro: humble
+          os_code_name: jammy
```
clalancette, line 264: Should we update the `runs-on` stanza for these to be `ubuntu-22.04`?

tfoote:
Replaced by #1015

PR #1021, opened 2024-02-06
Do not install ccache unless shared_ccache is used in binarydeb

j-rivero:
Currently the sharing of compilation artifacts through ccache is conditional after #844, which enable the support for sharing with the host using the `shared_ccache` option in the build files. The ccache binaries are currently being installed unconditionally thought.

The PR changes the behavior to install the ccache binaries only when the `shared_ccache` option is being used.

### DIFF
--- a/ros_buildfarm/scripts/release/run_binarydeb_job.py
+++ b/ros_buildfarm/scripts/release/run_binarydeb_job.py
@@ -16,7 +16,7 @@ import argparse
 import copy
 import sys
 
-from ros_buildfarm.argument import add_argument_append_timestamp
+from ros_buildfarm.argument import add_argument_append_timestamp, add_argument_install_ccache
 from ros_buildfarm.argument import add_argument_arch
 from ros_buildfarm.argument import add_argument_binarypkg_dir
 from ros_buildfarm.argument import \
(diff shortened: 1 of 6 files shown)


ros_buildfarm/scripts/release/run_binarydeb_job.py
```diff
@@ -16,7 +16,7 @@
 import copy
 import sys
 
-from ros_buildfarm.argument import add_argument_append_timestamp
+from ros_buildfarm.argument import add_argument_append_timestamp, add_argument_install_ccache
```
tfoote, line 19: We're doing one per line for all the others. We should do that here too

tfoote, line 19: We're doing one per line for all the others. We should do that here too. Note that will move it down to keep in alphabetical order.

tfoote (commented):
This generally looks good. One nit about the import ordering.

ros_buildfarm/scripts/release/run_binarydeb_job.py
j-rivero: [64df76d](https://github.com/ros-infrastructure/ros_buildfarm/pull/1021/commits/64df76d7545f4f08e8d15523eff5ea3869ac6b50)

tfoote (approved):
lgtm after the fixup

PR #1026, opened 2024-03-13
Adding break-system-packages pip option for noble in devel_task

marcoag:
Rpr jobs are failing due to the `externally-managed-environment` pip error, i.e.:

https://build.ros2.org/job/Rpr__rosbag2__ubuntu_noble_amd64/4/console

This PR adds a conditional to use `--break-system-packages`  when installing `setuptools` with `pip3`.

Maybe not affecting Rpr jobs but I was wondering if some other templates installing pip packages should also be updated, thoughts?


nuclearsandwich:
I can't see from the commit log why we started installing setuptools from pip. This _seems_ to me like something we'd done for a specific reason that has been lost to time. Is it worth switching this to use the system provided setuptools for noble and phase out this pip override?

PR #1028, opened 2024-03-15
Use --break-system-packages when installing setuptools on noble.

nuclearsandwich:
Reverts ros-infrastructure/ros_buildfarm#1026 which started out with this approach and then was altered (by me) to remove the pip install entirely. This is an alternative fix to #1027 while we determine what is actually needed to rely fully on apt.


clalancette (approved):

nuclearsandwich:
I'm closing this as we've been able to recover the dev jobs by bumping our colcon-core packages so that the upstream packages no longer create installation issues. This is a "temporary" fix until ROS repository configuration manages apt pinning or we add our own explicitly to buildfarm scripts but that temporary time will hold us at least until Ubuntu updates their colcon packages in a distribution we support.

If I eat these words in 2026 I will buy cottsay a meal of his choosing.

Issue #1029, opened 2024-03-15
ROS 2 Doc is currently broken


tfoote:
I've gotten everything back to green on CI in https://github.com/ros-infrastructure/ros_buildfarm/pull/1015 except ROS 2 Doc

The error is related to the python build process and needs to be resolved. 

https://github.com/ros-infrastructure/ros_buildfarm/pull/1015#issuecomment-1874653709

nuclearsandwich:
This is fixed by #1031.

PR #1030, opened 2024-03-16
Copy rosdoc2 sources into docker image instead of bind-mounting them read-only.

nuclearsandwich:
This is being explored as a fix for #1029.
The deprecated option that we were relying on to allow rosdoc2 to be built from a read-only source directory has been removed and I did not find a suitable PEP517 builder with an out-of-tree option.

I probably could have just cut `:ro` from the docker run invocation and called it a day and I may open that PR to see how it compares. I was trying to avoid modifying the system host's rosdoc2 clone, but given that most of this occurs in the bootstrap task I don't think that these portions of the workspace are preserved at all between job runs so that concern may be groundless.

For this solution, I copied in the source directory first to the docker build context and then into the image itself before installing it with `--break-system-packages` which seems to be required even when installing into a local user directory.

One thing I _don't_ like about this approach and the mixing of pip and break-system-packages is the fact that dependencies unsatisfied when `pip install . ...` is invoked will be pulled in from pip along with our local project. I would like to mitigate this with the installation of dependencies from apt before pip is invoked with an option not to also install dependencies. Whether I do this with a full rosdep setup or by eye is TBD.

### DIFF
--- a/ros_buildfarm/scripts/doc/build_rosdoc2.py
+++ b/ros_buildfarm/scripts/doc/build_rosdoc2.py
@@ -49,6 +49,7 @@ def main(argv=sys.argv[1:]):
                                   '-m',
                                   'pip',
                                   'install',
+                                  '--break-system-packages',
                                   '.'],
                                  cwd=args.rosdoc2_dir)
         if pip_rc:
--- a/ros_buildfarm/scripts/doc/build_rosdoc2.py
+++ b/ros_buildfarm/scripts/doc/build_rosdoc2.py
@@ -49,8 +49,6 @@ def main(argv=sys.argv[1:]):
                                   '-m',
                                   'pip',
                                   'install',
-                                  '--no-warn-script-location',
-                                  '--use-deprecated=out-of-tree-build',
                                   '.'],
                                  cwd=args.rosdoc2_dir)
         if pip_rc:
--- a/ros_buildfarm/scripts/doc/build_rosdoc2.py
+++ b/ros_buildfarm/scripts/doc/build_rosdoc2.py
@@ -63,7 +63,7 @@ def main(argv=sys.argv[1:]):
             env['PATH'] = ''
         else:
             env['PATH'] += ':'
-        env['PATH'] += '/home/buildfarm/.local/bin'
+        env['PATH'] += os.path.join(env['HOME'], '.local/bin')
 
         source_space = os.path.join(args.workspace_root, 'src')
         print("Crawling for packages in workspace '%s'" % (source_space))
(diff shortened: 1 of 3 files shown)


ros_buildfarm/scripts/doc/build_rosdoc2.py
```diff
@@ -64,7 +63,7 @@ def main(argv=sys.argv[1:]):
             env['PATH'] = ''
         else:
             env['PATH'] += ':'
-        env['PATH'] += '/home/buildfarm/.local/bin'
+        env['PATH'] += os.path.join(env['HOME'], '.local/bin')
```
nuclearsandwich, line 66: This was creating problems for my nuclearsandwich/uid-override patches which are required if your host user id is `1000` as the ubuntu:noble image now has an `ubuntu` user there.

ros_buildfarm/templates/doc/rosdoc2_job.xml.em
```diff
@@ -178,7 +180,6 @@ else:
         ' --rm ' +
         ' --cidfile=$WORKSPACE/docker_doc/docker.cid' +
         ' -v $WORKSPACE/ros_buildfarm:/tmp/ros_buildfarm:ro' +
-        ' -v $WORKSPACE/rosdoc2:/tmp/rosdoc2:ro' +
```
nuclearsandwich, line 181: This mount is no longer useful as we're copying the contents into the image.

ros_buildfarm/templates/doc/rosdoc2_job.xml.em
```diff
@@ -167,6 +167,8 @@ else:
         'sleep 1',
         '',
         'echo "# BEGIN SECTION: Build Dockerfile - doc"',
+	'# copy rosdoc2 into image build context directory',
+	'cp -r $WORKSPACE/rosdoc2 $WORKSPACE/docker_doc/rosdoc2',
```
nuclearsandwich, line 171: In order to be available in the docker context to be copied into the image the `rosdoc2` source directory needs to be in the docker context subdirectory.

nuclearsandwich:
It does indeed fix the test. Now I have to decide if an alternative is preferred.

tfoote (commented):
I think that it would make sense to just do the regular mount and drop the read only. 

In line 94 it's rmed before cloning so there shouldn't be a risk of persistence between runs
```
        'echo "# BEGIN SECTION: Clone rosdoc2"',
        'rm -fr rosdoc2',
        'python3 -u $WORKSPACE/ros_buildfarm/scripts/wrapper/git.py clone --depth 1 https://github.com/ros-infrastructure/rosdoc2.git rosdoc2',
```

Also I noticed there's another instance of mounting rosdoc2 into a container. that would also want to be synchronized either way we go.

clalancette (commented):
I don't have a strong opinion either way whether we bind mount r/w or take this solution where we copy.  I'll let you and @tfoote decide.

Otherwise, thanks for looking into this!  It is a huge help.

nuclearsandwich:
Closing in favor of #1031.

PR #1037, opened 2024-03-29
Add jammy rosdoc2 jobs as well.

nuclearsandwich:
I couldn't figure out if there was a better solution than just doubling the existing job. I marked the previous one as for Noble and the one I just added for Humble as "jammy" I guess that when we introduce new platforms we'd roll these forward or back based on what is currently supported.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -297,8 +297,8 @@ jobs:
           os_code_name: jammy
           repo: rcutils
 
-  ros2_doc:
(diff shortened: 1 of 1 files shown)


tfoote (approved):
We could do some fancy matrix mechanisms but it would take quite a bit of effort with the complexity for now this is enough to protect us from the regressions and won't need to be updated often.

cottsay (approved):
:bullettrain_front:

PR #1047, opened 2024-05-17
Clean up any dangling cidfiles (Container ID files) in workspace.

nuclearsandwich:
We previously built container images in a `docker_containers` subdirectory, but that directory and its cleanup were removed in #881 which results in the cidfile left in workspace between runs.


clalancette (approved):
:tada:

PR #1055, opened 2024-06-06
Add support for custom_rosdep_urls in build files

j-rivero:
In addition to the existing support of the parameter `custom_rosdep_urls` for the doc and source files, the PR adds the support for the build files so it can be used for things like nightly jobs.

I tested it here https://citest.build.osrfoundation.org/view/Rci/job/Rci__nightly-release_ubuntu_noble_amd64/26/ which builds gz-rendering8 (outside of ROS) using the packages.osrfoundation.org repository and an external rosdep rule to resolve `libogre-next-2.3-dev` to the package of the same name hosted in packages.o.o.

### DIFF
--- a/doc/configuration_options.rst
+++ b/doc/configuration_options.rst
@@ -597,3 +597,15 @@ The following options are valid in version ``1`` (beside the generic options):
   to, if desired.
   By default, the resulting archives are only available to other jobs within
   Jenkins.
+
+The following options are valid as keys in the ``_config`` dict under
+``targets``:
+
+* ``custom_rosdep_urls``: a list of URLs containing rosdep sources.list.d entry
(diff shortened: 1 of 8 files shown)


cottsay (commented):
Thanks for adding documentation.

The block you added the documentation to is specific to CI jobs, but the file you added the actual configuration support to is global to many job types. I think it might be more appropriate to add the config to `ci_build_file.py` rather than the global `build_file.py`.

Alternatively, we could keep the config change in `build_file.py` and move the documentation block, but in that case I think we would need to remove the configuration and documentation in the other specific jobs that already have it, and now inherit support from `build_file.py`. That said, I'm not sure that every job type that inherits from `build_file.py` can use rosdep to begin with, so I'm a little wary of this approach.

cottsay (approved):
Looks good, thanks!

PR #1084, opened 2025-03-04
24.04 migration

Crola1702:
With these changes I can run the buildfarm in jenkins LTS and 24.04:

* Use `importlib` instead of `imp`. This is suggested by python docs
* Use `UnifiedDiffUtils`


cottsay (approved):
I don't have a lot of feelings about the Groovy change, but the Python change looks good.

nuclearsandwich (approved):
The Groovy changes look fine. API drift makes sense given the time and changes.

PR #1093, opened 2025-04-22
Update ros_buildfarm templates plugins

Crola1702:
I used [this plugins versions](https://github.com/ros-infrastructure/cookbook-ros-buildfarm/blob/1faf7877d93ed9b51e04ed42f3c61e2fb9ff21b3/attributes/plugins.rb)

<details>
<summary>
and this script:
</summary>

```bash
#!/bin/bash

if [[ ! -f "$1" ]]; then
  echo "Usage: $0 <plugins.rb>"
  exit 1
fi

grep -r 'plugin="[^"]*@@[^"]*"' | sed -n 's/.*plugin="\([^@]*\)@\([^"]*\)".*/\1@\2/p' > current_versions.txt

while read -r line; do
  plugin_name="${line%@@*}"
  old_version="${line#*@@}"
  new_version=$(sed -n "/^ *\"$plugin_name\" => \"\([^\"]*\)\".*/ s//\1/p" $1)
  if [[ -n "$new_version" ]]; then
    echo "$plugin_name@@$old_version $plugin_name@@$new_version"
  fi
done < current_versions.txt > versions_to_replace.txt

while read -r old new; do
  old_escaped=$(sed 's/[\/&]/\\&/g; s/\./\\\./g' <<< "$old")
  new_escaped=$(sed 's/[\/&]/\\&/g' <<< "$new")
  grep -rl "$old" | xargs sed -i "s/plugin=\"$old_escaped\"/plugin=\"$new_escaped\"/g"
done < versions_to_replace.txt

rm current_versions.txt

echo "Plugin versions updated."
```

</details>

### DIFF
--- a/ros_buildfarm/templates/dashboard_view_all_jobs.xml.em
+++ b/ros_buildfarm/templates/dashboard_view_all_jobs.xml.em
@@ -1,4 +1,4 @@
-<hudson.plugins.view.dashboard.Dashboard plugin="dashboard-view@@2.16">
+<hudson.plugins.view.dashboard.Dashboard plugin="dashboard-view@@2.521.v339b_a_f4d8da_8">
   <name>@view_name</name>
   <description>Generated at @ESCAPE(now_str) from template '@ESCAPE(template_name)'</description>
   <filterExecutors>false</filterExecutors>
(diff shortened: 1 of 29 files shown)


nuclearsandwich (approved):
I think this is a good start. Whether the plan is to check for structural config changes in a follow-up PR or add additional commits to this PR is up to you.

I don't think "right versions, outdated structure" is materially worse than "outdated versions, outdated structure" on `master`.

PR #1100, opened 2025-09-25
Update the documentation for ROS 2.

clalancette:
Now that ROS 1 is EOL, the examples in the documentation here seem more suited to point to ROS 2 examples.

While I was in here, I also removed the bits talking about how to configure for Travis, now that Travis is no longer a thing.

### DIFF
--- a/doc/jobs/ci_jobs.rst
+++ b/doc/jobs/ci_jobs.rst
@@ -130,6 +130,15 @@ The return code of the invocation of ``catkin_tests_results`` /
 ``colcon test-result`` is then available in the environment variable
 ``test_result_RC``.
 
+Run the *CI* job on external CI
+-------------------------------
+
+Since it is easy to run a *CI* job locally it can also be run on an external CI
+runner (like GitHub actions) to either test every commit or pull request.
+The setup and invocation is the same as locally.
+An example can be found in the `.github/actions/ci/action.yaml <https://github.com/ros-infrastructure/ros_buildfarm/blob/master/.github/actions/ci/action.yaml>`_
+file of the *ros_buildfarm* repository.
+
 Run for "custom" repositories
 -----------------------------
 
--- a/doc/jobs/devel_jobs.rst
+++ b/doc/jobs/devel_jobs.rst
@@ -189,6 +189,14 @@ The return code of the invocation of ``catkin_tests_results`` /
 ``colcon test-result``is then available in the environment variable
 ``test_result_RC``.
 
+Run the *devel* job on external CI
+-------------------------------
+
+Since it is easy to run a *devel* job locally it can also be run on an external CI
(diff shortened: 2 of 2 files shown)


doc/jobs/ci_jobs.rst
```diff
@@ -124,69 +124,12 @@ Instead of invoking the generated script it can also be *sourced*:
...
 .. code:: sh
 
-  . ci_job_crystal.sh
+  . ci_job_rolling_ament_cmake_ros.sh
 
 The return code of the invocation of ``catkin_tests_results`` /
 ``colcon test-result`` is then available in the environment variable
 ``test_result_RC``.
 
-Run the *CI* job on Travis
```
nuclearsandwich, line 133: This section could be preserved / adapted to run via GitHub Actions. There are jobs in this very repo that use this.

doc/jobs/devel_jobs.rst
```diff
@@ -183,68 +183,12 @@ Instead of invoking the generated script it can also be *sourced*:
...
 .. code:: sh
 
-  . devel_job_indigo_roscpp_core.sh
+  . devel_job_rolling_rclcpp.sh
 
 The return code of the invocation of ``catkin_tests_results`` /
 ``colcon test-result``is then available in the environment variable
 ``test_result_RC``.
 
-Run the *devel* job on Travis
```
nuclearsandwich, line 192: This section could be adapted to GitHub Actions and there are examples in this very repo.

doc/jobs/prerelease_jobs.rst
```diff
@@ -115,13 +115,6 @@ scripts which can be invoked manually in the same order:
   ``colcon test-result --all`` to output a summary of all tests in the overlay
   workspace.
 
-Run the *prerelease* job on Travis
```
nuclearsandwich, line 118: I'm not sure if suggesting pre-release jobs is as worthy of preserving. I know they work in theory but I think test suite quality makes these pretty difficult to parse.

nuclearsandwich (commented):
Thanks, nice to see that effectively not much has changed but it's worth referring to the ROS 2 counterparts. I've left a few suggestions for how to handle the Travis sections as they're illustrative of how to run these jobs under any general CI system as well.

doc/jobs/devel_jobs.rst
clalancette: Ditto.

doc/jobs/prerelease_jobs.rst
clalancette: I left this one out, but let me know if you change your mind and I can easily do the same thing here I did for the above ones.

Crola1702 (commented):
Looks good to me! Just one comment here

PR #1102, opened 2025-10-10
Make doc job git clone recursive

InvincibleRMC:
See more info over there https://github.com/ros2/ros2_documentation/pull/5928


cottsay (approved):
I think this is how Bloom behaves, so it makes sense to adopt it here too.

PR #1104, opened 2025-11-11
Bump GH Actions to 22.04 and 24.04, Python 3.7

cottsay:
The Ubuntu 20.04 runners are no longer available, and Python 3.7 is the oldest supported version on 22.04 runners.


cottsay:
We'll need to address the pytest regressions independently, but this is now ready for review.

PR #1112, opened 2026-01-05
Use parallel testing by default

Crola1702:
Currently, build.ros2.org jobs use sequential executor for testing. This leads to ~2 hours of testing. Parallelization might save around 1 hour of testing.

This is kind of a cherry pick from: https://github.com/ros2/ci/pull/723


cottsay:
I don't think we want to do this in `ros_buildfarm`. It can be done entirely using config changes. The default behavior for colcon should continue to align with what we saw in `catkin` for consistency.

It should be possible to add `python3-colcon-ros-domain-id-coordinator` to the `install_packages` option and then `--executor parallel` can be added to `build_tool_test_args`.

PR #1119, opened 2026-01-20
Use python standard library to replace bzip2 dependency

j-rivero:
This pull request updates the way install space compression is performed in the CI job template. The main change is switching from a shell-based `tar` command to a Python-based approach using the `tarfile` module. This should save us a dependency on bzip2 for the Jenkins nodes. See https://github.com/ros-infrastructure/ros_buildfarm/issues/1118

Mainly using `arcname` to do the conversion from ws/install_isolated to ros-${version}. Not fully sure if there is a use case that is not be covered by the IA generated regexp.

Local generation:
```diff
   @@ -418,2 +418 @@                                                                                                                                                                                                                                                                                                                  
         -cd $WORKSPACE                                                                                                                                                                                                                                                                                                                     
         -tar -cjf ros2-global-linux-jammy-amd64-ci.tar.bz2 -C ws --transform "s/^install_isolated/ros2-linux/" install_isolated                                                                                                                                                                                                            
         +cd $WORKSPACE &amp;&amp; python3 -c "import tarfile; t = tarfile.open('ros2-global-linux-jammy-amd64-ci.tar.bz2', 'w:bz2'); t.add('ws/install_isolated', arcname='ros2-linux'); t.close()"
```

Tested https://build.osrfoundation.org/job/__test_ci__colcon_any-manual_ubuntu_noble_amd64/3/ . Tar seems like:
```shell
~/Downloads/ros2-global-linux-noble-amd64-ci via 🐏 25GiB/31GiB | 964kiB/8GiB ❯ tree -L 2
.
└── ros2-linux
    ├── COLCON_IGNORE
    ├── gz-cmake
    ├── gz-common
    ├── gz-fuel_tools
    ├── gz-gui
    ├── gz-math
    ├── gz-msgs
    ├── gz-physics
    ├── gz-plugin
    ├── gz-rendering
    ├── gz-sensors
    ├── gz-sim
    ├── gz-tools2
    ├── gz-transport
    ├── gz-utils
    ├── local_setup.sh
    ├── _local_setup_util_sh.py
    ├── sdformat
    └── setup.sh

17 directories, 4 files
```

### DIFF
--- a/ros_buildfarm/scripts/ci/create_workspace_archive.py
+++ b/ros_buildfarm/scripts/ci/create_workspace_archive.py
@@ -1,0 +1,74 @@
+# Copyright 2025 Open Source Robotics Foundation, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import argparse
+import hashlib
+import os
+import sys
+import tarfile
+
+from ros_buildfarm.argument import add_argument_arch
+from ros_buildfarm.argument import add_argument_os_code_name
+from ros_buildfarm.argument import add_argument_rosdistro_name
+from ros_buildfarm.common import Scope
+
+
+def main(argv=sys.argv[1:]):
+    parser = argparse.ArgumentParser(
+        description='Create a compressed archive of the workspace install space.')
+    add_argument_rosdistro_name(parser)
+    add_argument_os_code_name(parser)
+    add_argument_arch(parser)
+    parser.add_argument(
+        '--ros-version',
+        type=int,
+        required=True,
+        help='The ROS version (1 or 2)')
+    parser.add_argument(
+        '--install-dir',
+        required=True,
+        help='The path to the install space to archive')
+    parser.add_argument(
+        '--output-dir',
+        default='.',
+        help='The directory to write the archive and checksum to')
+    args = parser.parse_args(argv)
+
+    rosdistro_name = args.rosdistro_name if args.rosdistro_name else 'global'
+    archive_name = 'ros%d-%s-linux-%s-%s-ci.tar.bz2' % (
+        args.ros_version, rosdistro_name, args.os_code_name, args.arch)
+    archive_path = os.path.join(args.output_dir, archive_name)
+    checksum_path = archive_path.replace('.tar.bz2', '-CHECKSUM')
+    arcname = 'ros%d-linux' % args.ros_version
+
+    with Scope('SUBSECTION', 'create workspace archive'):
+        print("Creating archive '%s' from '%s'" % (archive_path, args.install_dir))
+        with tarfile.open(archive_path, 'w:bz2') as t:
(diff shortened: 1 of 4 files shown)


j-rivero:
We could potentially add the sha256 to Python with something like:
```diff
--- a/ros_buildfarm/templates/ci/ci_job.xml.em
+++ b/ros_buildfarm/templates/ci/ci_job.xml.em
@@ -373,17 +373,18 @@ parameters = [
     script='\n'.join([
         'echo "# BEGIN SECTION: Compress install space"',
         'cd $WORKSPACE && python3 -c "'
+        'import hashlib; '
         'import tarfile; '
-        't = tarfile.open('
-        '\'ros%d-%s-linux-%s-%s-ci.tar.bz2\', '
-        '\'w:bz2\'); '
+        'archive = \'ros%d-%s-linux-%s-%s-ci.tar.bz2\'; '
+        't = tarfile.open(archive, \'w:bz2\'); '
         't.add(\'ws/install_isolated\', arcname=\'ros%d-linux\'); '
-        't.close()"' % (
+        't.close(); '
+        'h = hashlib.sha256(); '
+        'h.update(open(archive, \'rb\').read()); '
+        'open(archive.replace(\'.tar.bz2\', \'-CHECKSUM\'), \'w\').write(h.hexdigest() + \' *\' + archive + \'\\\\n\')"' % (
           ros_version, rosdistro_name or 'global', os_code_name, arch,
           ros_version
         ),
-        'sha256sum -b ros%d-%s-linux-%s-%s-ci.tar.bz2' % (ros_version, rosdistro_name or 'global', os_code_name, arch) +
-        ' > ros%d-%s-linux-%s-%s-ci-CHECKSUM' % (ros_version, rosdistro_name or 'global', os_code_name, arch),
         'cd -',
         'echo "# END SECTION"',
     ]),
~
```
but not sure if that is going to be a performance improvement, python needs to close the file and call that open through hashlib again.

cottsay:
I took a quick look at the Python docs and didn't find what I was hoping to see.

The idea is to open the tar file object and wrap that file object in an object that computes the sha256 as data is written to the file object. The `tarfile.open` function can be passed that file object instead of the file path.

I'm confident that such code can be written, but I'm disappointed that Python doesn't make it easier to combine file streams and buffered readers/writers with hashlib.

In any case, even if we don't optimize the computation as I had in mind, moving the sha256 computation into Python also drops our dependency on the `sha256sum` tool. Assuming that tool is present just like we did with `bzip2` is why we're having this conversation to begin with.

cottsay:
An additional thought is that we might want to move this Python code into a new script in `ros_buildfarm` rather than invoking inline Python. `create_workspace_archive.py` or something.

j-rivero:
Moved to an script in [75c13a2](https://github.com/ros-infrastructure/ros_buildfarm/pull/1119/commits/75c13a2c6d697fd1c9a703e29332386b50f87251) + [2266df1](https://github.com/ros-infrastructure/ros_buildfarm/pull/1119/commits/2266df113328f5ea678cd0419eb99415057806c1).

Avoid loading the whole tarball in memory and process it by chunks in [b2131c6](https://github.com/ros-infrastructure/ros_buildfarm/pull/1119/commits/b2131c663487234e0fdd3d7e3975e8f98014af74)

j-rivero:
Given that the code is not trivial I used the AI to generate a test file in [01db84b](https://github.com/ros-infrastructure/ros_buildfarm/pull/1119/commits/01db84bc0dee49110ab1be8667fc8df921d7b457)

j-rivero:
Testing it [Build Status](https://build.osrfoundation.org/job/ci__colcon_gpu_any-manual_ubuntu_noble_amd64/26/)

cottsay (approved):
This looks good to me, thanks for iterating.

PR #1120, opened 2026-01-30
Native multiarch support with backwards compatibility to osrf images

Crola1702:
# Description

This PR adds multi-arch Docker build support and introduces a configurable base image override path for release jobs. It updates how base images are selected so legacy `osrf/*` images are used only for specific arch/codename combinations, while defaulting to official OS images otherwise.

## Changes

- Added `--docker-base-image-override` argument and propagated it through release job configuration and release scripts.
- Passed the override into binarydeb task generation templates and Dockerfile snippets.
- Updated build templates to use docker build `--platform=linux/<arch>`.
- Pinned doc-independent/doc-metadata Docker builds to linux/amd64.
- Refined base-image resolution logic in shared snippets (builder_check-docker and from_base_image).

## Considerations
- `docker_base_image_override` is treated as image name, so it still appends `:<os_code_name>` tag at the end

## Testing

This behavior was tested in:

| Ubuntu Distribution / Architecture | Amd64 | Arm64 |
| -- | -- | -- |
| **Noble** | `--platform=linux/amd64` and `FROM ubuntu:noble` ([reference build](https://build.test.ros2.org/view/Rbin_unv8_uNv8/job/Rbin_uN64__ament_package__ubuntu_noble_amd64__binary/3/consoleFull)) | `--platform=linux/arm64` and `FROM osrf/ubuntu_arm64:noble` ([reference build](https://build.test.ros2.org/view/Rbin_unv8_uNv8/job/Rbin_unv8_uNv8__ament_package__ubuntu_noble_arm64__binary/6/consoleFull)) |
| **Resolute** | `--platform=linux/amd64` and `FROM ubuntu:resolute` ([reference build](https://build.test.ros2.org/view/Rci/job/Rci__nightly-release_ubuntu_resolute_amd64/39/consoleFull#console-section-9)) | `--platform=linux/arm64` and `FROM ubuntu:resolute` ([reference build](https://build.test.ros2.org/view/Rci/job/Rci__nightly-release_ubuntu_resolute_arm64/4/consoleFull#console-section-9)) |

Closes https://github.com/osrf/infrastructure-private/issues/60


ros_buildfarm/scripts/misc/docker_pull_baseimage.py
```diff
@@ -24,12 +24,17 @@ def main(argv=sys.argv[1:]):
     base_image = get_base_image_from_dockerfile(dockerfile)
     print(base_image)
 
+    platform_flag = None
+    if base_image.startswith('arm64v8/'):
+        platform_flag = 'linux/arm64'
```
claraberendsen: Do we need this flag even if the image is already targeting the correct architecture? 
I believe docker should be able to fetch the correct image based on the host OS. 
If you do `docker run -it ubuntu:noble uname -m` it will automatically download the correct image if we are not doing cross-compilation.

claraberendsen:
For the transition to using the native multi arch docker images the main things we should provide are:

- Use the correct docker image for the base OS when we are not using qemu. That should be as simple as fetching the ubuntu:noble image from the correct agent. 
- Allow backwards compatibility for the older images noble and before. That should be a flag that could be `--legacy-multi-arch` or something as to have a path to what was working. 
- For resolute onward use the native docker `--platform` to signal that we want to emulate the other architecture. If qemu is already installed and configured properly on the host `--platform` should suffice to indicate.

I think is a topic of discussion if we want this to be a flag the user has to set for the job to use emulation or if we want ros_buildfarm to work smarter and detect the arch drift and set the correct docker args.

ros_buildfarm/scripts/release/create_binarydeb_task_generator.py
```diff
@@ -56,6 +56,10 @@ def main(argv=sys.argv[1:]):
     add_argument_dockerfile_dir(parser)
     add_argument_env_vars(parser)
     add_argument_skip_tests(parser)
+    parser.add_argument(
```
claraberendsen: We should invert this argument and set the new way as default with a specific argument to fallback to the previous ways.

ros_buildfarm/templates/snippet/builder_check-docker.xml.em
```diff
@@ -1,7 +1,11 @@
 @{
 # same logic as in from_base_image.Dockerfile.em
-if arch in ['i386', 'armhf', 'arm64']:
+docker_platform_arg = ''
+if arch in ['i386', 'armhf', 'arm64'] and not vars().get('use_official_docker_images', False):
     base_image = 'osrf/%s_%s:%s' % (os_name, arch, os_code_name)
+elif arch == 'arm64' and vars().get('use_official_docker_images', False):
+    base_image = 'arm64v8/%s:%s' % (os_name, os_code_name)
```
claraberendsen: Even though this naming is correct I think it's unnecessary since in modern docker the architecture resolution is done based on the image pull.

Crola1702:
So, with the last changes, I have the following results for amd and arm

<details><summary>amd64 (my workstation)</summary>

Running:
```
python3 -m ros_buildfarm.scripts.release.generate_release_script https://raw.githubusercontent.com/ros2/ros_buildfarm_config/refs/heads/ros2/index.yaml rolling default ament_package ubuntu noble amd64 > amd64_job.sh
```

[amd64_job.sh](https://github.com/user-attachments/files/25188354/amd64_job.sh)
[amd64_build_log.txt](https://github.com/user-attachments/files/25188355/amd64_build_log.txt)

</details>

<details><summary>arm64 (testfarm agent)</summary>

Running:
```
python3 -m ros_buildfarm.scripts.release.generate_release_script https://raw.githubusercontent.com/ros2/ros_buildfarm_config/refs/heads/ros2/index.yaml rolling unv8 ament_package ubuntu noble arm64 > arm64_job.sh
```

[arm64_job.sh](https://github.com/user-attachments/files/25188371/arm64_job.sh)
[arm64_build_log.txt](https://github.com/user-attachments/files/25188372/arm64_build_log.txt)


Note in `job.sh` that ros_buildfarm generates `--platform=linux/arm64` but also `--platform=linux/armv8`. Not completely sure why, but it seemed to work, as you can see in the build_log.

</details>

What I think I'm missing is the possibility of using `osrf/ubuntu_{arch}:noble` images.

I added `--docker-image-prefix`, so we can do something like `osrf/ubuntu:noble` instead. However, I'm not sure how to put the architecture in-between, as using `--docker-image-prefix osrf/ubuntu_arm64:` results in `osrf/ubuntu_arm64:ubuntu:noble` image :thinking:

Looking for another review and ideas now @cottsay @claraberendsen

Crola1702:
@claraberendsen  suggestion:

- `--docker-image-prefix` is `os_code` by default. So `--docker-image-prefix` is actually the whole image.
- To keep in mind: in `buildfarm_tools_config` we should specify `--docker-image-prefix osrf/ubuntu_{arch}` so that we keep backwards compatibility
- We should configure a job in the test farm to check if everything is working correctly

Maybe we can do instead (as we're not actually doing a prefix, but the image without the tag):
```diff
- --docker-image-prefix
+ --docker-image
```

Crola1702:
With the current changes now running:
```
python3 -m ros_buildfarm.scripts.release.generate_release_script https://raw.githubusercontent.com/ros2/ros_buildfarm_config/refs/heads/ros2/index.yaml rolling unv8 ament_package ubuntu noble arm64 --docker-image-prefix osrf/ubuntu_arm64 > job.sh
```

[arm64_job.sh](https://github.com/user-attachments/files/25235668/arm64_job.sh)
[arm64_build_log.txt](https://github.com/user-attachments/files/25235667/arm64_build_log.txt)

ros_buildfarm/scripts/misc/docker_pull_baseimage.py
cottsay, line 1: I don't see any actual changes in this file. Can it be reverted?

ros_buildfarm/scripts/release/create_binarydeb_task_generator.py
```diff
@@ -32,6 +32,7 @@
 from ros_buildfarm.argument import add_argument_rosdistro_index_url
 from ros_buildfarm.argument import add_argument_rosdistro_name
 from ros_buildfarm.argument import add_argument_skip_tests
+from ros_buildfarm.argument import add_argument_docker_image_prefix
```
cottsay: Please alphabetize the imports - same comment in other files.

ros_buildfarm/templates/release/deb/binarypkg_create_task.Dockerfile.em
```diff
@@ -2,10 +2,11 @@
...
     'snippet/from_base_image.Dockerfile.em',
-    os_name=os_name,
-    os_code_name=os_code_name,
-    arch=arch,
-))@
+        os_name=os_name,
+        os_code_name=os_code_name,
+        arch=arch,
+        docker_image_prefix=vars().get('docker_image_prefix'),
+    )) @
```
cottsay:
```suggestion
    os_name=os_name,
    os_code_name=os_code_name,
    arch=arch,
    docker_image_prefix=vars().get('docker_image_prefix'),
)) @
```

ros_buildfarm/templates/release/deb/binarypkg_create_task.Dockerfile.em
```diff
@@ -107,7 +108,8 @@ cmds.append(
     ' --distribution-repository-key-files ' + ' ' .join(['/etc/apt/keyrings/ros-buildfarm-%d.asc' % i for i in range(len(distribution_repository_keys))]) +
     ' --binarypkg-dir ' + binarypkg_dir +
     ' --env-vars ' + ' '.join(build_environment_variables) +
-    ' --dockerfile-dir ' + dockerfile_dir +
-    (' --skip-tests' if skip_tests else ''))
-}@
+        ' --dockerfile-dir ' + dockerfile_dir +
+        (' --skip-tests' if skip_tests else '') +
+        (' --docker-image-prefix ' + docker_image_prefix if vars().get('docker_image_prefix') else ''))
+    } @
```
cottsay:
```suggestion
    ' --dockerfile-dir ' + dockerfile_dir +
    (' --skip-tests' if skip_tests else '') +
    (' --docker-image-prefix ' + docker_image_prefix if vars().get('docker_image_prefix') else ''))
} @
```

ros_buildfarm/templates/snippet/builder_check-docker.xml.em
```diff
@@ -1,9 +1,6 @@
 @{
 # same logic as in from_base_image.Dockerfile.em
-if arch in ['i386', 'armhf', 'arm64']:
-    base_image = 'osrf/%s_%s:%s' % (os_name, arch, os_code_name)
-else:
-    base_image = '%s:%s' % (os_name, os_code_name)
+base_image = '%s:%s' % (vars().get('docker_image_prefix', '%s' % os_name), os_code_name)
```
cottsay: Can we drop this format string?

```suggestion
base_image = '%s:%s' % (vars().get('docker_image_prefix', os_name), os_code_name)
```

Also, if I'm not mistaken, the value of `docker_image_prefix` will be `None` here if it isn't specified at all. Maybe it should be more like this?

```suggestion
base_image = '%s:%s' % (vars().get('docker_image_prefix') or os_name, os_code_name)
```

ALSO also, maybe this is where we should maintain the "current behavior" for backwards compatibility (but allow overriding using `docker_image_prefix`)

```suggestion
base_image = '%s:%s' % (
    vars().get('docker_image_prefix') or (
        'osrf/%s_%s' % (os_name, arch)
        if arch in ('i386', 'armhf', 'arm64')
        else os_name
    ),
    os_code_name,
)
```

ros_buildfarm/templates/snippet/from_base_image.Dockerfile.em
```diff
@@ -2,11 +2,7 @@
 FROM @base_image
 @[else]@
 @# same logic as in builder_check-docker.xml.em
-@[  if arch in ['i386', 'armhf', 'arm64']]@
-FROM osrf/@(os_name)_@arch:@os_code_name
-@[  else]@
-FROM @os_name:@os_code_name
-@[  end if]@
+FROM @docker_image_prefix:@os_code_name
```
cottsay: Whatever logic there is in `builder_check-docker.xml.em` should be mirrored here.

cottsay (changes requested):

ros_buildfarm/scripts/release/create_binarydeb_task_generator.py
Crola1702: Noted. Done in 1c2a44e

ros_buildfarm/templates/snippet/builder_check-docker.xml.em
Crola1702: Suggestion applied in 3877b1b

ros_buildfarm/templates/snippet/from_base_image.Dockerfile.em
Crola1702: Suggestion applied in 3877b1b

ros_buildfarm/templates/release/deb/binarypkg_create_task.Dockerfile.em
Crola1702: Suggestion applied in 3877b1b

ros_buildfarm/templates/release/deb/binarypkg_create_task.Dockerfile.em
Crola1702: Suggestion applied in 3877b1b

ros_buildfarm/templates/snippet/builder_check-docker.xml.em
```diff
@@ -1,9 +1,13 @@
 @{
 # same logic as in from_base_image.Dockerfile.em
-if arch in ['i386', 'armhf', 'arm64']:
-    base_image = 'osrf/%s_%s:%s' % (os_name, arch, os_code_name)
-else:
-    base_image = '%s:%s' % (os_name, os_code_name)
+base_image = '%s:%s' % (
+    vars().get('docker_image_prefix') or (
+        'osrf/%s_%s' %(os_name, arch)
```
cottsay:
```suggestion
        'osrf/%s_%s' % (os_name, arch)
```

ros_buildfarm/argument.py
```diff
@@ -452,6 +452,10 @@ def add_argument_skip_tests(parser):
         '--skip-tests', action='store_true',
         help='Skip execution of tests as part of the package build process.')
 
+def add_argument_docker_image_prefix(parser):
+    parser.add_argument(
+        '--docker-image-prefix',
+        help='Prefix for the docker image (e.g. "osrf/" or "arm64v8/")')
```
cottsay, line 460: To please flake8:

```suggestion


def add_argument_docker_image_prefix(parser):
    parser.add_argument(
        '--docker-image-prefix',
        help='Prefix for the docker image (e.g. "osrf/" or "arm64v8/")')


```

ros_buildfarm/templates/snippet/from_base_image.Dockerfile.em
```diff
@@ -1,13 +1,22 @@
...
-@# same logic as in builder_check-docker.xml.em
-@[  if arch in ['i386', 'armhf', 'arm64']]@
-FROM osrf/@(os_name)_@arch:@os_code_name
-@[  else]@
-FROM @os_name:@os_code_name
-@[  end if]@
+@[if 'base_image' not in locals()]@
+@{
+base_image = '%s:%s' % (
+  vars().get('docker_image_prefix') or (
```
cottsay: This isn't really a prefix anymore, it's more like a `docker_base_image_override` or something.

claraberendsen (approved):
LGTM! Thanks for pushing this through

cottsay (approved):

PR #1122, opened 2026-04-14
Use dynamic RPM branch for RHEL 10+, Fedora 43+

cottsay:
This new generator uses a different branching scheme in Bloom, and does not include the OS name, codename, or version.


claraberendsen (approved):
Simple, short and sweet! Looks good!

PR #1123, opened 2026-04-14
Make rosdep database available to RPM builds

cottsay:
The rosdep database will be used to resolve dependencies during the build of dynamic RPM packages. Rather than updating it inside the pristine build environment, we can mount an up-to-date rosdep cache into the buildroot.

### DIFF
--- a/ros_buildfarm/templates/release/rpm/sourcepkg_task.Dockerfile.em
+++ b/ros_buildfarm/templates/release/rpm/sourcepkg_task.Dockerfile.em
@@ -11,7 +11,9 @@ FROM @(os_name):@(os_code_name)
 
 RUN dnf update --refresh -y
 
-RUN dnf install --refresh -y --setopt=install_weak_deps=False dnf{,-command\(download\)} git mock{,-{core-configs,scm}} python3{,-{catkin_pkg,empy,rosdistro,yaml}}
+RUN dnf install --refresh -y --setopt=install_weak_deps=False dnf{,-command\(download\)} git mock{,-{core-configs,scm}} python3{,-{catkin_pkg,empy,rosdep,rosdistro,yaml}}
+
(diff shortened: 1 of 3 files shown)


ros_buildfarm/templates/release/rpm/sourcepkg_task.Dockerfile.em
```diff
@@ -11,7 +11,9 @@ FROM @(os_name):@(os_code_name)
 
 RUN dnf update --refresh -y
 
-RUN dnf install --refresh -y --setopt=install_weak_deps=False dnf{,-command\(download\)} git mock{,-{core-configs,scm}} python3{,-{catkin_pkg,empy,rosdistro,yaml}}
+RUN dnf install --refresh -y --setopt=install_weak_deps=False dnf{,-command\(download\)} git mock{,-{core-configs,scm}} python3{,-{catkin_pkg,empy,rosdep,rosdistro,yaml}}
+
+RUN rosdep init
```
claraberendsen, line 17: Isn't it necessary to update the rosdep DB here as well?

cottsay: We don't actually need the rosdep database for source package builds. The reason it's installed here like this is so that the Dockerfile for the source and binary jobs are perfectly aligned and share more cache.

cottsay: Someday we should unify a "base image" concept for these jobs to formalize that pattern.

claraberendsen: That makes sense. And 100% on board to have a pre-built base image that workers share.

claraberendsen (approved):

PR #1124, opened 2026-04-14
Plumb package_ignore_list into RPM build environment

cottsay:
By making the global package_ignore_list available within the RPM build environment, we enable the dynamic RPM generator respect the same ignore list that Jenkins is using to build out the job graph.

---

Specifically, this will be critical for RHEL where we need to ignore all references to RTI connext packages.


claraberendsen (approved):
LGTM! 
Good to have this plumbed through for RHEL and other scenarios where we might need to skip chunks of the job graph.

PR #1130, opened 2026-05-11
Remove legacy osrf/* image default override

Crola1702:
# Description
Remove legacy `osrf/*` image in the docker template. This aligns the build process with modern multi-arch official library images and simplifies the template logic.

# Changes
* Base image is now built only with the `os_code_name` and the `os_name` or the docker image override attribute

# Considerations
* With this change, we will use always Docker Hub official images by default
* `docker_base_image_override` variable remains functional for cases where a custom base image is required. It can replicate the previous compatibility if `--docker-base-image-override osrf/ubuntu_arm64` is specified.
* Old distributions might not be compatible with this change

### DIFF
--- a/ros_buildfarm/templates/snippet/builder_check-docker.xml.em
+++ b/ros_buildfarm/templates/snippet/builder_check-docker.xml.em
@@ -1,16 +1,7 @@
 @{
 # same logic as in from_base_image.Dockerfile.em
 base_image = '%s:%s' % (
-    vars().get('docker_base_image_override') or (
-        'osrf/%s_%s' % (os_name, arch)
-        if arch in ('i386', 'armhf', 'arm64') and (
-            os_code_name in (
-                'artful', 'bionic', 'cosmic', 'disco', 'focal',
-                'jammy', 'noble', 'wily', 'xenial', 'yakkety',
-                'zesty')
-        )
-        else os_name
-    ),
+    vars().get('docker_base_image_override') or os_name,
     os_code_name,
 )
 }@
--- a/ros_buildfarm/templates/snippet/from_base_image.Dockerfile.em
+++ b/ros_buildfarm/templates/snippet/from_base_image.Dockerfile.em
@@ -1,16 +1,7 @@
 @[if 'base_image' not in locals()]@
 @{
 base_image = '%s:%s' % (
-  vars().get('docker_base_image_override') or (
-      'osrf/%s_%s' % (os_name, arch)
-      if arch in ('i386', 'armhf', 'arm64') and (
-            os_code_name in (
-                'artful', 'bionic', 'cosmic', 'disco', 'focal',
-                'jammy', 'noble', 'wily', 'xenial', 'yakkety',
-                'zesty')
-      )
-      else os_name
-  ),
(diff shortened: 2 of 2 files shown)


cottsay:
I think you'll need to update `ros_buildfarm/templates/snippet/builder_check-docker.xml.em` as well

christophebedard:
I don't feel very qualified to review this, but it LGTM (with Scott's feedback addressed). Will this be tested first on something like https://build.test.ros2.org/?

Crola1702:
> I think you'll need to update ros_buildfarm/templates/snippet/builder_check-docker.xml.em as well

Good catch. Thanks!

> I don't feel very qualified to review this, but it LGTM (with Scott's feedback addressed). Will this be tested first on something like https://build.test.ros2.org/?

Yes we should. I will trigger a deploy there and then you can run the packages that are failing to test.

Actually, now that I see the testfarm with a lot of agents. @cottsay do you think we should downscale the testfarm? Or are we testing something heavy these days?

Crola1702:
I just deployed these changes in build.test.ros2.org. @christophebedard feel free to test whatever you need there and we can discuss the results here

cottsay:
> do you think we should downscale the testfarm? Or are we testing something heavy these days?

It's just the smoke build for Fedora. Once that finishes, yes, we should scale in.

christophebedard:
I triggered a `Hbin_ujv8_uJv8__ros_workspace__ubuntu_jammy_arm64__binary` job, but it looks like it's still using the `osrf/ubuntu_arm64:jammy` image: https://build.test.ros2.org/job/Hbin_ujv8_uJv8__ros_workspace__ubuntu_jammy_arm64__binary/3/console#console-section-10. Am I missing something?

cottsay:
I don't think it's been fully deployed, we'll need the reconfigure job to run first.

I'm skeptical that the staging farm is in a healthy enough place right now, we may need to take a moment to re-deploy the agents to pick up the new GPG key.

christophebedard:
> I don't think it's been fully deployed, we'll need the reconfigure job to run first.

What do we need to do? Is it something I can do?

cottsay:
Normally a triggering the reconfigure job after deploying it with the custom branch is all that's needed, but even that job can't run until the GPG issue is resolved: https://build.test.ros2.org/job/Rrel_reconfigure-jobs/799/

cottsay:
Specifically, it looks like the existing key for ROS Staging expired last month. I can't get to this until next week, unless @Crola1702 wants to do it.

cottsay (approved):
Alright, this change passed checks on the staging buildfarm. Here's a passing canary build:

https://build.test.ros2.org/job/Hbin_ujv8_uJv8__ament_package__ubuntu_jammy_arm64__binary/4/

The logs show `FROM ubuntu:jammy`, where we previously saw `FROM osrf/ubuntu_arm64:jammy`, indicating that this change is having the desired effect.

christophebedard:
amazing!

christophebedard:
can we merge this and get this deployed on build.ros2.org?

christophebedard (approved):

PR #1131, opened 2026-05-11
Make RPM buildroot search failure a warning

cottsay:
The process for extracting the maintainer E-mail addresses wasn't done very well in the initial RPM build implementation in ros_buildfarm, and this fragile code has bit us in the past.

Modern RPM build has changed the directory layout again and I think it's time to try a different approach here. To unblock Fedora builds, I'm proposing that we make this failure non-fatal. We won't have maintainer E-mails turned on for Fedora anyway, so this shouldn't result in any change in behavior (unless RHEL adopts Fedora's newer RPM version, which I can't see happening until RHEL 11).

Demo build using this change: https://build.test.ros2.org/view/Rbin_fedora_fc4364/job/Rbin_fedora_fc4364__ament_package__fedora_43_x86_64__binary/5/

### DIFF
--- a/ros_buildfarm/binaryrpm_job.py
+++ b/ros_buildfarm/binaryrpm_job.py
@@ -16,6 +16,7 @@ from datetime import datetime
 import glob
 import os
 import subprocess
+import sys
 
 from ros_buildfarm.common import get_os_package_name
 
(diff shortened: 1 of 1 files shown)


claraberendsen (approved):
This looks good to me to unblock the Fedora builds. 
It would be good to ticket the necessary improvements on the maintainer email address extraction to fix it from the root.

cottsay:
Filed as #1132

PR #1134, opened 2026-05-14
Respect modified dependency behaviors when disabling jobs

### LINKED ISSUE #1133: Inconsistent behavior around include_test_dependencies
It seems that this flag correctly dropped the test dependencies from the jobs, but when build of an upstream test dependency is blocked in the config, downstream packages appear to get disabled unnecessarily.

cottsay:
This change plumbs the build file configuration values which augment dependency enumeration into the code which disables jobs downstream of other jobs which have been explicitly disabled in the build config.

Closes #1133

### DIFF
--- a/ros_buildfarm/common.py
+++ b/ros_buildfarm/common.py
@@ -681,13 +681,18 @@ def get_package_manifests(dist):
     return cached_pkgs
 
 
-def get_implicitly_ignored_package_names(cached_pkgs, explicitly_ignored_pkg_names):
+def get_implicitly_ignored_package_names(
+    cached_pkgs, explicitly_ignored_pkg_names,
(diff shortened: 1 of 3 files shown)


cottsay:
This is a bit tricky to test prior to deployment, but I did deploy it to build.test.ros2.org and observed the effects on the Fedora packages there, which use both of the flags plumbed through here. The net effect was that 30 packages got enabled which were previously unnecessarily disabled before this change, and no new packages got disabled.

All 30 of the packages which got enabled built successfully.

As confident as I might be, we should still wait until after Lyrical ships to merge this.

cottsay:
I updated this to include the status pages, but including groups support there will require more work.

claraberendsen (approved):
LGTM!

PR #1142, opened 2026-07-15
Set `artifact_num_to_keep` to 30 for CI jobs config

Crola1702:
# Description

Limit CI artifact retention to 30 builds by reducing Jenkins job artifact storage overhead by retaining only the 30 most recent builds per job.

The artifacts are the heaviest part of a build result. Cleaning this will lead to a ~70% of less size in the CI jobs that leave artifacts


cottsay (approved):
Wow, I thought for sure the retention period was exposed by the build configs, but that's clearly not the case.

Not a today problem.

PR #1143, opened 2026-07-17
Add artifact {Num,Day} ToKeep logic in log-rotator template

Crola1702:
We were missing this to wire #1142

### DIFF
--- a/ros_buildfarm/templates/snippet/property_log-rotator.xml.em
+++ b/ros_buildfarm/templates/snippet/property_log-rotator.xml.em
@@ -2,8 +2,8 @@
       <strategy class="hudson.tasks.LogRotator">
         <daysToKeep>@int(days_to_keep)</daysToKeep>
         <numToKeep>@int(num_to_keep)</numToKeep>
-        <artifactDaysToKeep>-1</artifactDaysToKeep>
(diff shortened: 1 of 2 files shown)


ros_buildfarm/templates/snippet/property_log-rotator.xml.em
```diff
@@ -3,7 +3,7 @@
         <daysToKeep>@int(days_to_keep)</daysToKeep>
         <numToKeep>@int(num_to_keep)</numToKeep>
         <artifactDaysToKeep>-1</artifactDaysToKeep>
-        <artifactNumToKeep>-1</artifactNumToKeep>
+        <artifactNumToKeep>@(artifact_num_to_keep if 'artifact_num_to_keep' in vars() else -1)</artifactNumToKeep>
```
cottsay: We should probably do an implicit `int` conversion to align with the other values.

```suggestion
        <artifactNumToKeep>@int(artifact_num_to_keep if 'artifact_num_to_keep' in vars() else -1)</artifactNumToKeep>
```

Should we consider adding a var for `artifact_days_to_keep` while we're here?

cottsay (approved):

PR #1144, opened 2026-07-20
Vendor crate dependencies in sourcejobs

Blast545:
We ship the source code of the dependencies required for a crate using cargo vendor before building it.

We use a per-package flag to enable it / disable it.

Decision: we leave git sources and private registries out of the vendoring, at least for now

### DIFF
--- a/ros_buildfarm/sourcedeb_job.py
+++ b/ros_buildfarm/sourcedeb_job.py
@@ -12,12 +12,13 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import json
 import os
 import subprocess
 from urllib.error import HTTPError
 from urllib.request import urlretrieve
 
+from ros_buildfarm.cargo import is_vendoring_requested
+from ros_buildfarm.cargo import vendor_cargo_crates
 from ros_buildfarm.common import get_os_package_name
 from ros_buildfarm.release_common import dpkg_parsechangelog
 
@@ -100,8 +101,8 @@ def get_sources(
     # The crates go into debian/ because the quilt patches of a '3.0 (quilt)'
     # source package cannot represent the binary files some crates contain,
     # while anything under debian/ ships in the debian tarball untouched
-    if _is_vendoring_requested(pkg):
-        _vendor_cargo_crates(sources_dir, ['debian', 'vendor'])
+    if is_vendoring_requested(pkg):
+        vendor_cargo_crates(sources_dir, ['debian', 'vendor'])
 
 
 def _get_source_tag(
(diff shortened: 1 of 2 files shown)


ros_buildfarm/sourcedeb_job.py
```diff
@@ -102,7 +102,36 @@ def _get_source_tag(
          pkg_version, os_code_name)
 
 
-def build_sourcedeb(sources_dir, os_name=None, os_code_name=None):
+def _vendor_cargo_crates(sources_dir):
```
cottsay: This should be refactored into a common location so that the RPM builds can invoke it as well.

ros_buildfarm/sourcedeb_job.py
```diff
@@ -102,7 +102,36 @@ def _get_source_tag(
...
+        raise RuntimeError(
+            "Cannot vendor cargo crates: no 'Cargo.lock' in '%s'. "
+            'A lock file committed alongside Cargo.toml is required for '
+            'reproducible dependency resolution.' % sources_dir)
+
+    # vendor into debian/ so that the crates are shipped in the debian
+    # tarball, the quilt patches of a '3.0 (quilt)' source package cannot
+    # represent the binary files some crates contain
+    cmd = [
+        'cargo', 'vendor', '--locked', os.path.join('debian', 'vendor')]
```
cottsay: As discussed, we won't be able to use `--locked` because it would require re-generation of the lock file every time an upstream ROS cargo package version is updated.

```suggestion
    cmd = [
        'cargo', 'vendor', os.path.join('debian', 'vendor')]
```

ros_buildfarm/templates/release/deb/binarypkg_job.xml.em
```diff
@@ -173,12 +173,15 @@ but disabled since the package is blacklisted (or not whitelisted) in the config
...
+        '# cargo packages are built without network access, the task',
+        '# generator drops a marker file for them',
+        'if [ -f "$WORKSPACE/docker_build_binarydeb/isolate_network" ]; then docker_network=none; else docker_network=host; fi',
         'docker run' +
         ' --rm ' +
         ' --cidfile=$WORKSPACE/docker_build_binarydeb/docker.cid' +
         ' -e=HOME=/home/buildfarm' +
         ' -e=TRAVIS=$TRAVIS' +
-        ' --net=host' +
+        ' --net=$docker_network' +
```
cottsay: I think we should keep existing precedent and still allow network access during these builds, but we should update the cargo configuration to disallow network access as part of the Bloom template.

https://doc.rust-lang.org/cargo/reference/config.html#netoffline

cottsay (commented):
I'll add RPM changes to this PR as well.

It would be a good idea to leverage "Multiple Upstream Tarballs" to keep the vendor crates separate. The unpacking story is pretty clean.

We also need to make this process "opt-in", likely based on a specific configuration in the `<export>` section of the `package.xml` file.

ros_buildfarm/sourcedeb_job.py
Blast545: Yes, I'm iterating on that locally. quilt complains if we generate a lockfile and ship it if it wasn't part of the original source, so I'm currently running cargo vendor and later delete the file. commit coming later

ros_buildfarm/templates/release/deb/binarypkg_job.xml.em
Blast545: Would you agree to disable it here for developing and then restore it? I wouldn't like us to hit a specific corner case of `cargo` ignoring its own `--offline` flag. (One case, apparently, build scripts and proc macros ignore this flag, as per LLM discussion)

Blast545: Just curious, any reason to keep existing precedent? I thought it was a bug when I found it, and imagined it was a good opportunity to do it differently here

Blast545: Reverted those changes for now

ros_buildfarm/sourcedeb_job.py
Blast545: I'm covering both possibilities in current implementation, not quite sure if the locked path will break, but can be easily removed after testing.

ros_buildfarm/sourcedeb_job.py
Blast545: Moved into cargo.py so that it can be re-used. Also modified the function so that it won't assume debian/vendor, and now instead receives a path, since it might be different for RPM.

PR #1146, opened 2026-07-30
Support multiple targets on import upstream

claraberendsen:
### Description 

The import_upstream jobs don't support the scenario for multiple os targets/repositories under the same package format.  That had the side effect that subsequent runs with different OS compited and only the last one succedded. This reworks that job to aggregate all targets regardless of the specific OS being configured.

Though the import_upstream_rpm configuration is updated multiple times on a single deployment, it's the simplest way to keep the repository configuration as part of the job config.


claraberendsen:
@cottsay not sure if we want to preserve the repository configuration as part of the job xml or if we want to make it more dynamic and copy some of the other jobs by having a python script that can list the targets dinamically. 

Open to suggestions here.

PR #1149, opened 2026-08-05
Remove obsolete setTriggered() call in disk-space check Groovy script

Crola1702:
## Summary

Removes the call to `DiskSpaceMonitorDescriptor.DiskSpace.setTriggered(Class, boolean)` from the "check free disk space" System Groovy builder snippet. This method has been removed from Jenkins core, and any controller running a version built from https://github.com/jenkinsci/jenkins/pull/8593 or later will fail this build step with:

```
groovy.lang.MissingMethodException: No signature of method: hudson.node_monitors.DiskSpaceMonitorDescriptor$DiskSpace.setTriggered()
```

## Root cause: Jenkins 2.555.3 migration

Jenkins core PR #8593 ([JENKINS-72009, JENKINS-72200, JENKINS-24947] various fixes and improvements around disk space monitoring, merged 2023-11-24) removed the triggered boolean field and its public setTriggered(Class, boolean) mutator from DiskSpaceMonitorDescriptor.DiskSpace, in commit [690f76d](https://github.com/jenkinsci/jenkins/pull/8593/changes/690f76dece58646407d1bb8090c334eff206968a) ("remove triggered field").

Current master instead exposes a `protected void setTrigger(Class<? extends AbstractDiskSpaceMonitor> trigger)`, a different name, a different signature (single arg, no boolean), and no longer public, so the old call is simply gone from the public API surface.

Because `setTriggered()` was an internal implementation detail rather than a documented Extension Point, its removal was not called out in the user-facing Jenkins changelog; it only shows up in this PR's diff.

### DIFF
--- a/ros_buildfarm/templates/snippet/builder_check-docker.xml.em
+++ b/ros_buildfarm/templates/snippet/builder_check-docker.xml.em
@@ -17,12 +17,11 @@ base_image = '%s:%s' % (
     'builder_system-groovy',
     command=
 """// DISABLE SLAVE IF OUTPUT INDICATES DOCKER RUN PROBLEMS
-import hudson.model.Cause.UserIdCause
+import hudson.model.Cause.UpstreamCause
 import hudson.model.ParametersAction
 import hudson.slaves.OfflineCause.UserCause
 import java.io.BufferedReader
 import java.util.regex.Pattern
-import org.jvnet.hudson.plugins.groovypostbuild.GroovyPostbuildAction
 
 pattern = Pattern.compile("'docker run' returned [^0].*")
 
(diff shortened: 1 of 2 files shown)


cottsay:
This is probably how the original script was taking the agent offline.

Either we should find a different way to do that, or we should update the message printed to the log saying we're doing so.

cottsay:
Alright, I really dug into this and tested it on build.test.ros2.org. There were a few things wrong with the script. I tested these changes and they're working as expected now.

1. Drop the badges - even when they're working (they're broken in this code) they seem to hang the jenkins UI
2. Use UpstreamCause so that we preserve the causal tree when we re-trigger
3. Fix the call to mark the node offline and populate the optional fields of the DiskSpace cause

cottsay (approved):

Crola1702:
Thank you Scott!

cottsay:
I've approved the two new Groovy script templates in Jenkins and started deploying jobs for the two ROS distributions planning syncs, Lyrical and Humble.

The other distros will catch the nightly reconfigures.

PR #1151, opened 2026-08-23
Auto-trigger sync-to-main after successful sync-to-testing

Yadunund:
## Background

We'd like to sync Rolling packages automatically to `main` based on some criteria that can be adjusted easily. See [Discourse Post](https://discourse.openrobotics.org/t/rethinking-rolling-syncs/56842) for more details. This PR adds opt-in support to automatically trigger it as a downstream job whenever a `sync-packages-to-testing` job succeeds. 

### Changes

**New config option — `sync.auto_trigger_main`**

A new boolean field is added to the `sync` section of the release build file
(default: `false`). When set to `true`, the `sync-to-testing` job will trigger
the `sync-to-main` job on success:

```yaml
sync:
  auto_trigger_main: true
  package_count: 499
  packages: [desktop, desktop-full]
```

The implementation is pure plumbing across four files:

- `ros_buildfarm/config/release_build_file.py` — parses `sync.auto_trigger_main` into `sync_auto_trigger_main` (`bool`, default `False`)
- `ros_buildfarm/release_job.py` — passes sync_to_main_job_name and sync_auto_trigger_main into the template data for `_get_sync_packages_to_testing_job_config`. Note: `get_sync_packages_to_main_job_name()` already existed and is reused as-is.
- `templates/release/deb/sync_packages_to_testing_job.xml.em` — adds a conditional `publisher_build-trigger` snippet in the publishers section
- `templates/release/rpm/sync_packages_to_testing_job.xml.em` — same

Each arch-specific testing job independently triggers the single `sync-to-main` job on success. When `auto_trigger_main` is absent or false, no `BuildTrigger` is emitted and behavior is identical to before.

### DIFF
--- a/doc/configuration_options.rst
+++ b/doc/configuration_options.rst
@@ -238,6 +238,8 @@ The following options are valid in version ``2`` (beside the generic options):
   * ``package_percent``: The minimum percentage of enabled, *binary* packages
     which must be available.  If both ``package_count`` and ``package_percent``
     are specified, both numbers have to be met for a sync to happen.
+  * ``auto_trigger_main``: if ``true``, automatically trigger the
+    sync-to-main job after a successful sync to testing (default: ``false``).
 
 * ``target_queue``: the path where incoming changes to the target repository
   are being queued (default: ``/var/repos/ubuntu/building/queue``).
--- a/ros_buildfarm/config/release_build_file.py
+++ b/ros_buildfarm/config/release_build_file.py
@@ -94,6 +94,7 @@ class ReleaseBuildFile(BuildFile):
         self.sync_package_count = None
         self.sync_package_percent = None
         self.sync_packages = []
+        self.sync_auto_trigger_main = False
         if 'sync' in data:
             if 'package_percent' in data['sync']:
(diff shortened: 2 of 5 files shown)


cottsay:
Alright, this was deployed to the rolling builds on https://build.test.ros2.org/.

Here's the config branch I'm using: https://github.com/ros-staging/ros_buildfarm_config/tree/Crola1702/build-test-ros2

After a brief look, I think the only thing we're missing is that we need to abort sync-to-main if any of the upstream sync-to-testing jobs are in-progress or not passing. There's an existing snippet for this, but it checks jobs recursively, meaning we'd have a single groovy script checking every single job status across all platforms of an entire distro, and I don't think that will scale. It should be sufficient to check just the sync-to-testing jobs themselves.

I think the right move is to make a new groovy script for checking just the upstream jobs. We can either make that a new snippet or just embed it in the job XML for now since it's the only use.

Yadunund:
@cottsay that's great to hear! 

Let's go with embedding it in the job for now? we can refactor it out later if needed.

